PR #672, opened 2019-08-01
performance tests

rien:
This PR adds some performance tests. It is currently unpolished and it depends on PR's of other gems because somehow nobody is performance testing rails 5.2 applications...
 
Note:
 - A benchmark can be run by running `bin/benchmark` and profiling can be done using `bin/profile`, both these commands run the perfomance tests in the `benchmark/` directory. The reason why they are not included in the normal tests is because they take a while to run and are very noisy.
 - To be able to create somewhat realistic results, the development seeds are loaded into the test database (encapsulated in a transaction which is rolled back after the test).
 - The reason why these tests are so noisy is because somehow, ruby warnings are turned on and I have not found a way to turn them off again. The warnings originate from within gems, so we cannot just fix them.

_[Original pull request](https://github.ugent.be/dodona/dodona/pull/672) by @rien on Thu Sep 20 2018 at 17:24._


bmesuere:
@charvp we can probably close this because of the new slow page detection and merge conflicts?

charvp:
Sure, the new slow-page detection should be sufficient for detecting slow pages for our use-case. Unless @rien has a different opinion?

I would keep the branch around, in case we want to continue this work at some point.

rien:
I agree with @charvp. If one should want to test the performance of Dodona in the future this is a good starting point to begin with, but this PR is currently not stable enough.

PR #1148, opened 2019-08-08
improve submission table links

### LINKED ISSUE #1089: links in submission table are confusing
This is the submissions listing for a course:
image

The name links to all submissions for that user, the exercise name links to the exercise page.

I think that linking the name to the course-scoped user profile page aligns more with the expected behaviour.

In addition, we could add a filter icon next to the user name and exercise name. Clicking that icon would then filter the submissions for that user or exercise. This could be implemented by linking to the existing filtered pages, or maybe even better by adding the respective name to the search box and filtering that way (if possible). This last option will also educate the user on that feature.

_[Original issue](https://github.ugent.be/dodona/dodona/issues/1089) by @bmesuere on Thu May 30 2019 at 17:11._

winniederidder:
Fixes #1089


app/assets/stylesheets/material_icons.css.less
```diff
@@ -125,6 +125,22 @@ a.btn-fab .mdi::before{
   left: -3px
 }
 
+.filter-icon {
+  color: @text-secondary;
+  letter-spacing: -2px; // make sure the tooltip is displayed correctly
+}
+
+td i {
```
charvp: We don't want this for every icon in the table (definitely the `margin-right`, the others I would assume as well). Make this specific to the filter icon.

app/assets/stylesheets/material_icons.css.less
```diff
@@ -125,6 +125,22 @@ a.btn-fab .mdi::before{
   left: -3px
 }
 
+.filter-icon {
+  color: @text-secondary;
+  letter-spacing: -2px; // make sure the tooltip is displayed correctly
```
charvp: This file offers little context to what "the tooltip" is. Explain this in more detail.

app/assets/stylesheets/material_icons.css.less
```diff
@@ -125,6 +125,22 @@ a.btn-fab .mdi::before{
...
+}
+
+td i { 
+  position: relative;
+  top: 3px; // move icons downwards in table row
+  margin-right: 3px; // add spacing between text and prevent tooltip from overflowing off screen
+}
+
+.filter-icon:hover {
```
charvp: Add this as a `&:hover { ... }` inside the `.filter-icon` definition.

app/assets/stylesheets/material_icons.css.less
```diff
@@ -125,6 +125,22 @@ a.btn-fab .mdi::before{
...
+  cursor: pointer;
```
charvp: The cursor definition doesn't need to be specific to the `:hover` pseudo-class.

charvp:
>     * On the submission listing at the bottom of an exercise page, the filter icon is visible. Since there is no search bar, this doesn't make sense.

Should the user column even be enabled? That table only shows submissions of the current user, so the name is redundant.

bmesuere:
Indeed you're right; we can remove that column. This is probably a leftover from a previous version of Dodona before we had a dedicated submissions page for an exercise.

winniederidder:
personal_table
Personal table when submitting exercises (column removed)

index_table
index table when requesting an overview

app/views/exercises/show.html.erb
```diff
@@ -76,7 +76,7 @@
               <% if !user_signed_in? %>
                 <div class="alert alert-info"><%= t ".not_logged_in_html", :url => sign_in_path %></div>
               <% elsif @submissions.length > 0 %>
-                <%= render partial: 'submissions/submissions_table', locals: {submissions: @submissions, exercise: @exercise, course: @course, user: current_user, remote_links: true} %>
+                <%= render partial: 'submissions/submissions_table', locals: {submissions: @submissions, exercise: @exercise, course: @course, user: current_user, remote_links: false, table_type: 'personal'} %>
```
charvp: You don't need this extra variable. If the user variable is set, the user column should not be shown. The `submissions/index.html.erb` and `submissions/index.js.erb` files will need to be changed to pass `@user` as the `user` local.

charvp: Also the `remote_links` variable should not be changed.

winniederidder: Remote links was accidental and is reverted. I don't think the index pages should pass @user, as now I can use local_assigns.haskey?. They weren't originally passed along so I think this better fits the state before the feature.

charvp: They should be passed along, because the current behaviour of the submissions index is wrong. `/users/<x>/submissions` shows the user column, which it shouldn't because that page only shows the submissions of one user. Also, the submissions table on the exercise page uses `submissions/index.js.erb` to switch pages, which means the user column will magically show up again when switching pages.

It is fine to do this in this PR, because the change isn't that big and otherwise this PR will make the problem more visible.

PR #1151, opened 2019-08-08
index page for submissions per judge

### LINKED ISSUE #1140: add index page with submissions per judge
When working on an existing judge, it's nice to have a collection of submissions to check for mistakes. Currently I go to a course I teach using the relevant judge, which works well enough if it's the only course. When other people start using our judges, it'd be nice to test their (style of) exercises as well.

I'd propose the addition of a `/judges/x/submissions` page which lists all submissions judged by `x`. @charvp suggested using the query `Submission.where(exercises: Exercises.where(judge: x))`.

winniederidder:
Fixes #1140

### DIFF
--- a/app/views/judges/show.html.erb
+++ b/app/views/judges/show.html.erb
@@ -32,7 +32,7 @@
             <strong><%= Judge.human_attribute_name("runner") %>:</strong> <%= @judge.runner %>
           </p>
           <p>
-            <strong><%= t ".all-submissions" %>:</strong> <%= link_to @judge.name, judge_submissions_path(@judge), title: @judge.name, class: "ellipsis-overflow"  %>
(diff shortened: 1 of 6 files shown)


app/views/judges/show.html.erb
```diff
@@ -31,6 +31,9 @@
           <p>
             <strong><%= Judge.human_attribute_name("runner") %>:</strong> <%= @judge.runner %>
           </p>
+          <p>
+            <strong><%= t ".all-submissions" %>:</strong> <%= link_to @judge.name, judge_submissions_path(@judge), title: @judge.name, class: "ellipsis-overflow"  %>
```
charvp: The judge name after the `:` doesn't really make much sense here. Either move this to the navbar or put the submission count of that judge there.

PR #1157, opened 2019-08-09
add javascript linting

### LINKED ISSUE #1126: create CircleCI config
I think we want the actual tests and linting checks. See dodona-edu/dolos for a similar config. We must do the following:
- add a plain circle ci config and merge in develop (so circleci starts building pull request)
- create the actual config in a new branch
- update the protected branch settings

bmesuere:
After playing with both eslint and prettier, I landed on just using eslint.

closes #1126

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -75,6 +75,12 @@ jobs:
       - run:
           name: Run RuboCop
           command: rubocop
+      - run:
+          name: Install eslint
+          command: yarn install
+      - run:
+          name: Run eslint
+          command: yarn lint
 
 workflows:
   version: 2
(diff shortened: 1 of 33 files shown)


.circleci/config.yml
```diff
@@ -75,6 +75,17 @@ jobs:
       - run:
           name: Run RuboCop
           command: rubocop
+      - run:
+          name: Install eslint
+          command: yarn install
+      - run:
+          name: Run eslint
+          command: yarn lint --quiet --format junit --output-file ~/reports/eslint.xml
```
charvp:
```suggestion
          command: yarn lint --quiet --format junit --output-file ~/reports/eslint/result.xml
```
If you do this rename, circleci will attribute the output to the correct tool.

PR #1161, opened 2019-08-11
remove old animation frame polyfill

bmesuere:
There is no need for the old animation frame polyfill anymore: https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame#Browser_compatibility.

Unfortunately, the same is not true for the full-screen API: https://developer.mozilla.org/en-US/docs/Web/API/Element/requestFullScreen#Browser_compatibility.

### DIFF
--- a/app/assets/javascripts/notifications.js
+++ b/app/assets/javascripts/notifications.js
@@ -23,7 +23,7 @@ function showNotification(content, properties = {}) {
         $notification.append("<div class='spinner'></div>");
     }
 
-    requestAnimationFrame(function () {
+    window.requestAnimationFrame(function () {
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/notifications.js
```diff
@@ -24,7 +23,7 @@ function showNotification(content, properties = {}) {
         $notification.append("<div class='spinner'></div>");
     }
 
-    requestAnimFrame(function () {
+    requestAnimationFrame(function () {
```
charvp: As mentioned in #1164, isn't it cleaner to explicitly call this on window?

PR #1164, opened 2019-08-11
refactor notifications to typescript

bmesuere:
Refactors the notifications to:
- use a class-based syntax
- use typescript
- drop jQuery

Depends on #1161

### DIFF
--- a/app/assets/javascripts/notification.ts
+++ b/app/assets/javascripts/notification.ts
@@ -1,16 +1,19 @@
+interface NotificationProperties {
+    autoHide: boolean;
+    loading: boolean;
+}
+
 /**
  * Shows a notification in the bottom left corner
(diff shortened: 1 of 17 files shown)


app/assets/javascripts/notification.ts
```diff
@@ -0,0 +1,46 @@
...
+        if (this.autoHide) {
+            setTimeout(() => {
+                this.hide();
+            }, 3000);
+        }
+    }
+
+    private show() {
+        document.querySelector(".notifications").prepend(this.notification);
+        requestAnimationFrame(() => {
```
charvp: Isn't it cleaner to explicitly call this on window? IMO, that makes it clearer that this isn't some imported function but a window member.

PR #1165, opened 2019-08-11
Seed API tokens

thepieterdc:
This PR adds API tokens to the database seeds. For every user (except for the 500 students), an API token is created, equal to their username. 

**Example:**
Zeus Kronosson will have an API token `zeus`.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -46,6 +46,15 @@ if Rails.env.development?
                 institution: ugent
   end
 
+  puts 'Creating API tokens'
+
+  [zeus, staff, jelix, mart, student].each do |user|
+    token = user.username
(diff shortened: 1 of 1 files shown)


db/seeds.rb
```diff
@@ -46,6 +46,15 @@
                 institution: ugent
   end
 
+  puts 'Creating API tokens'
+
+  [zeus, staff, jelix, mart, student].each do |user|
+    token = user.username
```
charvp:
```suggestion
    token = user.token || user.username
```
This way, the first three accounts will have API tokens zeus, staff and student respectively (same as their sign-in tokens).

thepieterdc: Looks good to me :)

PR #1166, opened 2019-08-12
visibility cue of class progress

### LINKED ISSUE #1076: visual hint about class progress visibility
Teachers don't have a visual clue whether or not the class progress is visible for students. Maybe we could add one that also serves as a toggle button:

image

_[Original issue](https://github.ugent.be/dodona/dodona/issues/1076) by @pdawyndt on Fri May 17 2019 at 17:41._

winniederidder:
Fixes #1076

### DIFF
--- a/app/views/exercises/_exercises_table.html.erb
+++ b/app/views/exercises/_exercises_table.html.erb
@@ -18,7 +18,7 @@
       <% if @course.blank? || local_assigns[:series]&.progress_enabled || current_user&.course_admin?(@course) %>
         <th class='count hidden-xs'>
           <%= t "exercises.index.class_progress" if @course %>
-          <% unless series.progress_enabled? %>
+          <% if defined? series && !series.progress_enabled %>
             <i class="mdi mdi-eye-off mdi-12 progress_visibility"
                data-toggle="tooltip"
                data-placement="top"
--- a/app/views/exercises/_exercises_table.html.erb
+++ b/app/views/exercises/_exercises_table.html.erb
@@ -19,7 +19,12 @@
         <th class='count hidden-xs'>
           <%= t "exercises.index.class_progress" if @course %>
           <% unless series.progress_enabled? %>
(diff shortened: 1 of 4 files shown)


app/views/exercises/_exercises_table.html.erb
```diff
@@ -16,7 +16,17 @@
       <th class='status-icon'></th>
       <th><%= t "exercises.index.exercise" %></th>
       <% if @course.blank? || local_assigns[:series]&.progress_enabled || current_user&.course_admin?(@course) %>
-        <th class='count hidden-xs'><%= t "exercises.index.class_progress" if @course %></th>
+        <th class='count hidden-xs'>
+          <%= t "exercises.index.class_progress" if @course %>
```
charvp: I think the conditional logic would be clearer if you moved the `if @course` to contain both the text and the icon. AFAIK there will always be a series if there is a course, so the `if defined? series && !series.progress_enabled` can be a simple `unless series.progress_enabled`.

config/locales/views/exercises/en.yml
```diff
@@ -30,6 +30,7 @@ en:
       repository: "Repository"
       path: "Path"
       class_progress: Class progress
+      class_progress_visibility_disabled: "The class progress is not visible for the students."
```
charvp:
```suggestion
      class_progress_visibility_disabled: "The class progress is not visible for students."
```

config/locales/views/exercises/nl.yml
```diff
@@ -30,6 +30,7 @@ nl:
       repository: "Repository"
       path: "Pad"
       class_progress: Voortgang groep
+      class_progress_visibility_disabled: "De voortgang van de groep is niet zichtbaar voor de studenten."
```
charvp:
```suggestion
      class_progress_visibility_disabled: "De voortgang van de groep is niet zichtbaar voor studenten."
```

PR #1169, opened 2019-08-13
reduce amount of queries when loading course members

### LINKED ISSUE #1153: 2*n queries are executed on course member page to load the users' institution and course labels
A well-placed `includes` in the course members controller should be able to avoid this.

winniederidder:
Fixes #1153


charvp:
There should also be an includes for the course labels.

PR #1194, opened 2019-08-19
Webpacker production environment

rien:
We can now compile assets for production without the javascript `devDependencies`.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -4,7 +4,7 @@
     "emitDecoratorMetadata": true,
     "esModuleInterop": false,
     "experimentalDecorators": true,
-    "lib": ["es6", "dom"],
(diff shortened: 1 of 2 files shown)


tsconfig.json
```diff
@@ -4,7 +4,7 @@
     "emitDecoratorMetadata": true,
     "esModuleInterop": false,
     "experimentalDecorators": true,
-    "lib": ["es6", "dom"],
+    "lib": ["es7", "dom"],
     "module": "es6",
```
charvp: Shouldn't this change to ES7 as well if we change it for `lib`?

rien: There is no `es7` value for this compiler option. The `module` option here is how the module code generation has to be done, the `lib` option specifies which polyfills can be used.

PR #1211, opened 2019-08-22
Fix and test user policy

rien:


app/policies/user_policy.rb
```diff
@@ -78,8 +78,10 @@ def heatmap?
   end
 
   def permitted_attributes
-    if user&.admin?
+    if user&.zeus?
       %i[username first_name last_name email permission time_zone]
+    elsif user&.admin?
```
charvp: Staff shouldn't be able to change any of this.

test/controllers/users_controller_test.rb
```diff
@@ -12,6 +12,17 @@ class UsersControllerTest < ActionDispatch::IntegrationTest
 
   test_crud_actions
 
+  test 'staff should not be able to make theirself zeus' do
```
charvp:
```suggestion
  test 'staff should not be able to make themself zeus' do
```

PR #1230, opened 2019-08-30
update yarn.lock

rien:
I updated our dependencies by removing `yarn.lock` and reinstalling. `yarn test` still succeeds and when going trough the site manually all the Javascript still seems to work.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -48,6 +48,7 @@ yarn-debug.log*
 
 # Ignore test reports
 test/reports/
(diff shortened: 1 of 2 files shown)


bmesuere:
We shouldn't do this anymore in the future. Dependabot does a good job of keeping everything up to date and updates both the lock and package file. Dependabot is set to update all direct dependencies and subdependencies in the case of security patches.

PR #1231, opened 2019-08-30
Rename exercise token

rien:
This copies the `exercises.token` column to `exercises.repository_token` (and changes all its usages to the new column) in order to complete the renaming with the follow-up PR.


db/migrate/20190829143205_add_repository_token_to_exercise.rb
```diff
@@ -0,0 +1,9 @@
+class AddRepositoryTokenToExercise < ActiveRecord::Migration[6.0]
+  def change
+    add_column :exercises, :repository_token, :string, limit: 64
+    Exercise.all.each do |exercise|
```
charvp:
```suggestion
    Exercise.all.find_each do |exercise|
```
Just to make sure we don't use too much memory.

PR #1234, opened 2019-08-30
description iframes

### LINKED ISSUE #1139: XSS risks from untrusted exercise sources
Now that more people can add exercises, we can't trust that content anymore.
There are two XSS security risks: the exercise descriptions and the html messages in the output table.

We must take action to reduce these risks.

## Exercise descriptions
For the exercise descriptions, I propose to host the actual descriptions on a separate (sub)domain and include them in the main website using an iframe and potentially setting the [sandbox attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#attr-sandbox). That way, the subdomain would be a (harmless) sandbox because the browser doesn't share (session)cookies between them.

Two issues come to mind if we would implement this: a UI problem regarding iframe height and exercise permissions.

### iframe height
Iframes have a fixed height and natively can't use the height of their content. This issue can be solved by
- using a fixed height for the exercise description and add a scroll bar for longer content
- some javascript that adapts the height of the iframe after loading it
- a combination of both

### Exercise permissions
The point of the sandbox is to be stateless and not contain any user information. Therefore, we can't use any additional information to check if a user has access to a certain exercise. We can, however, check these things at the "parent" page and only include the iframe if a user has the necessary permissions. To prevent guessing of the exercise URL, they should contain an additional random token that is regenerated every time the exercise permissions are changed.

## Feedback table output 
Judges can output html messages that can potentially contain harmful JavaScript. We could use the iframe strategy here, but the permission checking would be a bit more complicated. Using a separate iframe for each html message will solve this, but introduce several iframe height problems.

A more pragmatic solution might be to sanitise the html messages as much as possible. This sanitisation can easily be plugging in into the feedback table renderer. The biggest drawback of this is that JavaScript content in the judge output is no longer possible.

rien:
This is a basic implementation of sandboxing exercise descriptions within iframes. It also adds sanitation to course and series descriptions. 

This branch includes #1231 

Still TODO:
- [x] Check security settings
- [x] Check whether all current series and course descriptions are sanitized without too much impact.
- [x] ~~Sandbox or sanitize submission feedback tables~~ postponed to other PR


rien:
While checking the effects of sanitizing course and series descriptions, I noticed the following:
- 'Collapsible' panels are frequently used, and would stop working if we sanitize descriptions (because custom Javascript is used. Do we want to keep providing this functionality?
- A lot of links are present with `target="_blank"`, but these are filtered out. Should we allow the `target` attribute?
- Tables and `style` tags are heavily used in some courses for custom styling. Do we allow this?

rien:
Another consideration that should be made: we could add the [sandbox](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#attr-sandbox) attribute which would restrict the actions that can be performed within the iframe. However the `allow-top-navigation(-by-user-activation)` relaxation doesn't work on Firefox. so links won't work unless they have `target='_blank'`.

My opinion is that the `sandbox` is too strict for our use-case. The standard protections we achieve with `iframe` should be a huge improvement already.

bmesuere:
- If I'm not mistaken, the collapse doesn't need javascript because the functionality comes from bootstrap. Javascript is only used to toggle the icon which is not essential.
- target blank links should be allowed, but maybe add `rel="noopener noreferrer"` to prevent vulnerabilities xss issues
- not fond of users adding styles, but I guess we're stuck with it 😞 
- iframe without sandbox to start with is ok for me

rien:
@bmesuere :
> target blank links should be allowed, but maybe add `rel="noopener noreferrer"` to prevent vulnerabilities xss issues

Do we also want this in our iframes? They get the sandbox domain as referrer.

> not fond of users adding styles, but I guess we're stuck with it disappointed

Teachers who use the `style` attribute:
```
=> {"Wim Janin"=>3, "Nils Croes"=>3, "Dominiek Vandewalle"=>14, "Peter Dawyndt, Jan Fostier"=>1}`
```
Teachers who use tables:
```
=> {"Dominiek Vandewalle"=>14, "Wim Janin"=>1, "Nils Croes"=>1, "Helga Naessens"=>1}
```

I think it may be justified to contact those teachers and look if there's a possible solution?

bmesuere:
Viewing an exercise outside of a course seems to be broken. It seems that the same URL is used for viewing an exercise and for the iframe description which might cause this collision. (viewing an exercise outside of a course tries to load the description only but fails because the token is missing.

Maybe use a separate route for the iframe contents?

rien:
@bmesuere exercises out of a course are fixed now. 

Requests using the sandbox host which do not have a sandboxed 'route' are now redirected to the default host to prevent confusion.

app/helpers/application_helper.rb
```diff
@@ -104,9 +104,36 @@ def clipboard_button_for(selector)
...
+                           hard_wrap: false,
+                           syntax_highlighter: 'rouge',
+                           math_engine_opts: { preview: true })
+                      .to_html
+                      .html_safe
+  end
+
+  def sanitize(html)
+    tags = Rails::Html::SafeListSanitizer.allowed_tags.to_a
+    tags += %w[table thead tbody tr td]
```
charvp: Not sure if this is an extensive enough list. @toonijn uses `style` tags in [this course](https://dodona.ugent.be/nl/courses/172/).

app/helpers/exercise_helper.rb
```diff
@@ -14,9 +14,31 @@ def exercise_description_footnotes_and_first_image(exercise)
     [renderer.description_html, renderer.footnote_urls, renderer.first_image]
   end
 
+  def description_iframe(exercise)
+    id = "exercise-description-#{exercise.id}"
+    url = description_exercise_url(exercise,
+                                   token: exercise.access_token,
+                                   dark: session[:dark])
+    content_for :preload do
+      tag.link rel: 'preload', href: url, as: 'document'
```
bmesuere: I think this also needs the `crossorigin` attr (and it will not work on Firefox)

test/test_helper.rb
```diff
@@ -27,9 +27,6 @@ class ActiveSupport::TestCase
   include DelayedJobHelper
   include RemoteHelper
   include SeriesZipHelper
-
-  # Run tests in parallel with specified workers
-  parallelize(workers: :number_of_processors)
```
bmesuere: why is this removed?

rien: By mistake.

I have enabled parallel testing again.

app/helpers/exercise_helper.rb
rien: This currently also doesn't work in chrome: https://bugs.chromium.org/p/chromium/issues/detail?id=593267

Should we just skip preloading for now, since there's no browser who supports preloading iframes?

bmesuere: Ok, no preload for now

beardhatcode: ik ben niet volledig mee met deze PR, maar mischien is dit nuttig:

Prefetch (en prerender https://caniuse.com/#search=prefetch)

https://developer.mozilla.org/en-US/docs/Web/HTTP/Link_prefetching_FAQ

https://www.w3.org/TR/resource-hints/#dfn-prefetch

rien: We're only interested in `preload` here (declare early on that we want to load another document).

PR #1241, opened 2019-09-03
better ordering of exercises in series

### LINKED ISSUE #1080: wrong order of exercises in series
When a new series is created with more than 16 exercises, the order of the exercises in that series will differ from the order the exercises were added.

_[Original issue](https://github.ugent.be/dodona/dodona/issues/1080) by @dmouriss on Tue May 21 2019 at 16:43._

winniederidder:
Fixes #1080

### DIFF
--- a/app/models/series_membership.rb
+++ b/app/models/series_membership.rb
@@ -16,7 +16,7 @@ class SeriesMembership < ApplicationRecord
 
   delegate :course, to: :series
 
-  default_scope { order(order: :asc) }
+  default_scope { order(order: :asc, id: :desc) }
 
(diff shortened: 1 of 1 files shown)


app/models/series_membership.rb
```diff
@@ -16,7 +16,7 @@ class SeriesMembership < ApplicationRecord
 
   delegate :course, to: :series
 
-  default_scope { order(order: :asc) }
+  default_scope { order(order: :asc, id: :desc) }
```
charvp:
```suggestion
  default_scope { order(order: :asc, id: :asc) }
```

PR #1246, opened 2019-09-04
rework permission system for judges

### LINKED ISSUE #1085: Rework permission system for tabs in full output, add it to partial output
_[Original issue](https://github.ugent.be/dodona/dodona/issues/1085) by @charvp on Tue May 28 2019 at 17:41._

winniederidder:
Fixes #1085

### DIFF
--- a/app/runners/result_constructor.rb
+++ b/app/runners/result_constructor.rb
@@ -70,12 +70,12 @@ class ResultConstructor
                                       locale: @locale)
   end
 
-  def start_tab(title: nil, hidden: nil, permission: 'student')
+  def start_tab(title: nil, hidden: nil, permission: nil)
     check_level(:judgement, 'tab started')
     @tab = {}
     @tab[:description] = title
     @tab[:badgeCount] = 0
-    @tab[:permission] = permission
+    @tab[:permission] = permission unless permission.nil?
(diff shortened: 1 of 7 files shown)


app/runners/result_constructor.rb
```diff
@@ -70,11 +70,12 @@ def start_judgement
                                       locale: @locale)
   end
 
-  def start_tab(title: nil, hidden: nil)
+  def start_tab(title: nil, hidden: nil, permission: 'student')
```
ninewise: I'd use `permission: nil` here and `@tab[:permission] = permission unless permission.nil?` below. This avoids hardcoding the default (here), slightly decreases stored json size and is more consistent with the other methods in this class.

public/schemas/partial_output.json
```diff
@@ -33,7 +33,8 @@
         "properties": {
             "command": { "enum": ["start-tab"] },
             "title":   { "type": "string" },
-            "hidden":  { "type": "boolean" }
+            "hidden":  { "type": "boolean" },
+            "permission":  { "$ref": "#/definitions/permission" }
```
ninewise: Nitpick: all other properties are aligned.

PR #1252, opened 2019-09-06
link institution in user profile to institution page

### LINKED ISSUE #1115: link institution in user profile to institution page
User profile shows the institute of the user below the name of the user. Would be handy to link this institute to the page describing more details on the institute.

_[Original issue](https://github.ugent.be/dodona/dodona/issues/1115) by @pdawyndt on Thu Jun 27 2019 at 14:52._

LucianosLionakis:
Fixes #1115

### DIFF
--- a/app/views/course_members/show.html.erb
+++ b/app/views/course_members/show.html.erb
@@ -23,7 +23,9 @@
             </span>
           </span>
           <span class="profile-byline text-muted">
-            <%= link_to(@user.institution&.name, institution_path(@user.institution), class: "text-muted") %>
+            <% if @user.institution&.name.present? %>
+              <%= link_to(@user.institution&.name, institution_path(@user.institution), class: "text-muted") %>
+            <% end %>
             <% if @user.institution&.name.present? and @user.email.present? %>
               &middot;
             <% end %>
--- a/app/views/course_members/show.html.erb
+++ b/app/views/course_members/show.html.erb
@@ -26,7 +26,7 @@
             <% if @user.institution.present? and policy(@user.institution).show?%>
               <%= link_to(@user.institution&.name, institution_path(@user.institution), class: "text-muted") %>
             <% else %>
-              <%= @user.institution&.name %>ni
+              <%= @user.institution&.name %>
             <% end %>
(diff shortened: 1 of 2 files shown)


app/views/course_members/show.html.erb
```diff
@@ -23,7 +23,7 @@
             </span>
           </span>
           <span class="profile-byline text-muted">
-            <%= @user.institution&.name %>
```
charvp: User's don't always have an institution. This shouldn't be a link in that case.

charvp: You should also check that the user viewing the page can access the institution page via the policy (only zeus can).

app/views/course_members/show.html.erb
```diff
@@ -23,7 +23,9 @@
...
+            <% if @user.institution&.name.present? %>
```
charvp: You should also check permissions here.

app/views/course_members/show.html.erb
```diff
@@ -23,7 +23,9 @@
             </span>
           </span>
           <span class="profile-byline text-muted">
-            <%= @user.institution&.name %>
+            <% if policy(@user.institution).show? and @user.institution&.name.present? %>
```
charvp: The institution name should still be shown if the user can not access the institution.

charvp: The policy check fails if the institution is `nil` (for both pages).

charvp: When this works, you should check that the link isn't colored. We don't want this to stand out on the page.

app/views/course_members/show.html.erb
```diff
@@ -23,7 +23,11 @@
             </span>
           </span>
           <span class="profile-byline text-muted">
-            <%= @user.institution&.name %>
+            <% if @user.institution.present? and policy(@user.institution).show?%>
+              <%= link_to(@user.institution&.name, institution_path(@user.institution), class: "text-muted") %>
+            <% else %>
+              <%= @user.institution&.name %>ni
```
charvp:
```suggestion
              <%= @user.institution&.name %>
```

PR #1253, opened 2019-09-06
code selection in feedback update

### LINKED ISSUE #1052: allow selection in code from feedback table
Code as shown in feedback table does not allow to select a portion of the code. The entire code is automatically selected as soon as the code is clicked. The latter seems to be disturbing more, than it is useful.

_[Original issue](https://github.ugent.be/dodona/dodona/issues/1052) by @pdawyndt on Thu May 02 2019 at 09:33._

winniederidder:
Fixes #1052

### DIFF
--- a/app/views/submissions/_submissions_table.html.erb
+++ b/app/views/submissions/_submissions_table.html.erb
@@ -30,6 +30,8 @@
 <center><%= page_navigation_links submissions, true, 'submissions', params %></center>
 <script type="text/javascript">
     $(function () {
+      if(dodona.initFilterButtons){
         dodona.initFilterButtons();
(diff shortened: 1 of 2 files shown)


app/views/submissions/_submissions_table.html.erb
```diff
@@ -30,6 +30,8 @@
 <center><%= page_navigation_links submissions, true, 'submissions', params %></center>
 <script type="text/javascript">
     $(function () {
+      if(dodona.initFilterButtons){
```
charvp: Code style isn't checked in `erb` files, but we still like a space between the `if` and the condition.

PR #1268, opened 2019-09-10
delete series leads to course page

### LINKED ISSUE #1259: delete series leads to course page
In copying courses, you can't indicate which series have to be copied. It's either all or nothing, but I guess this issue has been resolved in #1090. For now, I had to delete some series after copying a course with all it's series. However, deleting one series from the `manage series` page redirects to the course page. Not very user-friendly when you want to delete multiple series from this page. I guess this is a leftover from the time we only allowed to delete series from the course page itself (#1090).

image

LucianosLionakis:
Fixes #1259

### DIFF
--- a/app/controllers/series_controller.rb
+++ b/app/controllers/series_controller.rb
@@ -92,7 +92,7 @@ class SeriesController < ApplicationController
   def destroy
     @series.destroy
     respond_to do |format|
-      format.html { redirect_to request.referer, notice: I18n.t('controllers.destroyed', model: Series.model_name.human) }
+      format.html { redirect_to request.referer || course_path(@series.course), notice: I18n.t('controllers.destroyed', model: Series.model_name.human) }
       format.json { head :no_content }
     end
   end
--- a/app/controllers/series_controller.rb
+++ b/app/controllers/series_controller.rb
@@ -95,7 +95,7 @@ class SeriesController < ApplicationController
     respond_to do |format|
       format.html { redirect_to course_url(course), notice: I18n.t('controllers.destroyed', model: Series.model_name.human) }
       format.json { head :no_content }
-      format.js 
+      format.js
     end
   end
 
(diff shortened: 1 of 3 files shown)


app/controllers/series_controller.rb
```diff
@@ -95,6 +95,7 @@ def destroy
     respond_to do |format|
       format.html { redirect_to course_url(course), notice: I18n.t('controllers.destroyed', model: Series.model_name.human) }
```
charvp: This line should definitely be changed as well. Note that there is some complexity, since a series can be deleted both from the course page and the manage series page.

app/controllers/series_controller.rb
```diff
@@ -95,6 +95,7 @@ def destroy
...
       format.json { head :no_content }
+      format.js
```
charvp: I'm not sure we want the dynamic updating. If we want to support it on both the course page and the manage series page I would think this would lead to some very complex code, which would not be worth the hassle (especially since deleting a series is not something we want to encourage).

app/controllers/series_controller.rb
LucianosLionakis: I think this line is fine, because this line is called when a series is deleted from the course page. Which refreshes the page and thus shows that the series is deleted.

app/controllers/series_controller.rb
LucianosLionakis: What if we don't support it on both pages but only on the manage series page where managing the series should be easy?

app/controllers/series_controller.rb
charvp: This could also be called from the manage series page if the user opens the link in a new tab or otherwise does something weird when opening the link. (Or when the dynamic updating is removed, which I am still a proponent of.)

app/controllers/series_controller.rb
charvp: It's still not really necessary. The page is fast enough, and like I said, deleting series is not something we want to encourage. It won't happen often and makes the code base more complex.

PR #1296, opened 2019-09-19
exercise model improve test coverage

LucianosLionakis:


test/models/exercise_test.rb
```diff
@@ -372,6 +372,36 @@ def config
...
+
+  test 'safe_delete should not destroy exercise if status is not removed' do
+    @exercise.safe_destroy
+    assert_not_nil @exercise
+  end
+
+  test 'safe_delete should destroy exercise if status is removed' do
+    @exercise.status = 2 # set status to removed
+    @exercise.safe_destroy
+    assert_not_nil @exercise
```
LucianosLionakis: Exercise should be destroy in this case. But it is stil not nil anyone know why?

charvp: A destroy only has effect on the database. The variable is not set to `nil`. (Which is impossible anyway, since the variable is not reassigned.)

test/models/exercise_test.rb
```diff
@@ -47,6 +47,55 @@ class ExerciseTest < ActiveSupport::TestCase
     end
   end
 
+  test 'accessible? should return false if user is not a member of the course' do
```
charvp: `accessible?` has a lot more cases. While they are apparently covered in other tests those other cases should probably be tested explicitly here since they are probably covered implicitly in the other tests.

PR #1297, opened 2019-09-19
Make sure we move the code and result files when a submission's attributes are manually updated

charvp:
This caused problems recently when some submissions had to be moved into a course. If [this method](https://github.com/dodona-edu/dodona/blob/develop/app/models/exercise.rb#L363) is called (also a manual action) it would cause similar problems. Same when users would need to be merged (e.g. if a user has multiple accounts through multiple institutions).

### DIFF
--- a/test/models/submission_test.rb
+++ b/test/models/submission_test.rb
@@ -115,4 +115,31 @@ class SubmissionTest < ActiveSupport::TestCase
     assert_equal 3, result[:groups][0][:groups][0][:groups][0][:messages].count
     assert_equal 3, result[:groups][0][:groups][0][:groups][0][:tests][0][:messages].count
   end
+
+  test 'transferring to another course should move the underlying result and code' do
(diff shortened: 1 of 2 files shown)


test/models/submission_test.rb
```diff
@@ -115,4 +115,31 @@ class SubmissionTest < ActiveSupport::TestCase
...
     assert_equal 3, result[:groups][0][:groups][0][:groups][0][:tests][0][:messages].count
   end
+
+  test 'transferring to another course should move the underlying result and code' do
+    submission = create :submission, result: 'result', code: 'code'
+    new_course = create :course
+    submission.update(course: new_course)
+    assert_equal 'result', submission.result
+    assert_equal 'code', submission.code
+    assert File.exist?(submission.fs_path)
```
rien: Maybe test if the old file is removed? Not that it is so essential though.

PR #1304, opened 2019-09-20
user model improve test coverage

LucianosLionakis:


test/models/user_test.rb
```diff
@@ -212,4 +212,148 @@ def assert_user_exercises(user, attempted, correct, unfinished)
...
+  test 'recent_exercises should return the 3 most recent exercises submissions have been submitted' do
+    user = create :user
+    exercise1 = create :exercise
+    exercise2 = create :exercise
+    create :series, exercises: [exercise1, exercise2]
+    create :submission, user: user, exercise: exercise1
+    assert_equal [exercise1], user.recent_exercises
+  end
+
+  test 'pending_series should return the 3 most recent exercises submissions have been submitted' do
```
charvp: I think you forgot to update the test names here?

test/models/user_test.rb
```diff
@@ -212,4 +212,148 @@ def assert_user_exercises(user, attempted, correct, unfinished)
...
+    user = create :user
+    course = create :course, users: [user]
+    create :series, course: course, exercise_count: 2, deadline: Time.current - 2.minutes # Not pending series
+    pending_series = create :series, course: course, exercise_count: 2, deadline: Time.current + 2.minutes
+    assert_equal [pending_series], user.pending_series
+  end
+
+  test 'homepage_series should return the 3 most recent exercises submissions have been submitted' do
```
charvp: I think you forgot to update the test names here?

test/models/user_test.rb
```diff
@@ -212,4 +212,148 @@ def assert_user_exercises(user, attempted, correct, unfinished)
...
+
+  test 'pending_courses should return the courses in which the user is a student' do
+    assert_equal [@pending_course], @user.pending_courses
+  end
+
+  test 'unsubscribed_courses should return the courses in which the user is a student' do
+    assert_equal [@unsubscribed_course], @user.unsubscribed_courses
+  end
+
+  test 'full_view? should return true because user has a course he/she has favorite' do
```
charvp:
```suggestion
  test 'full_view? should return true because user has a course they have favorited' do
```

test/models/user_test.rb
```diff
@@ -212,4 +212,148 @@ def assert_user_exercises(user, attempted, correct, unfinished)
...
+    assert_equal true, @user.full_view?
+  end
+
+  test 'full_view? should return false because user is not subscribed to any courses' do
+    user = create :user
+    assert_equal false, user.full_view?
+  end
+
+  test 'full_view? should return true because user is subscribed to more tha four courses' do
```
charvp:
```suggestion
  test 'full_view? should return true because user is subscribed to more than four courses' do
```

PR #1318, opened 2019-09-27
Change Naos to staging environment

rien:


config/deploy/staging.rb
```diff
@@ -1,55 +1,29 @@
...
-# group is considered to be the first unless any  hosts have the primary
-# property set. Specify the username and a domain or IP for the server.
-# Don't use `:all`, it's a meta role.
+set :delayed_job_workers, 3
 
-# role :app, %w{deploy@example.com}, my_property: :my_value
-# role :web, %w{user1@primary.com user2@additional.com}, other_property: :other_value
-# role :db,  %w{deploy@example.com}
+# Development mode doesn't use the `secret_key_base` stored in the credentials, but we would still like a stable key.
+set :linked_files, fetch(:linked_files, []).push('tmp/development_secret.txt')
```
charvp: I think this is no longer necessary? `/config/credentials/staging.key` might have to be added in its place?

config/environments/staging.rb
```diff
@@ -0,0 +1,108 @@
...
+
+  # Raises error for missing translations
+  # config.action_view.raise_on_missing_translations = true
+
+  # Use an evented file watcher to asynchronously detect changes in source code,
+  # routes, locales, etc. This feature depends on the listen gem.
+  # config.file_watcher = ActiveSupport::EventedFileUpdateChecker
+
+  config.middleware.use ExceptionNotification::Rack,
+                        ignore_if: ->(env, _exception) {env['HTTP_HOST'] == 'localhost:3000'},
```
charvp: Is this the config we want for staging? The `action_mailer.delivery_method` definitely seems wrong to me.

config/credentials/staging.yml.enc
```diff
@@ -0,0 +1 @@
+FbXR8Pry96aLKOyp/YYyKEnWjZf32bJUbVqqOw1bcqCCREA7AGmUVUUsJaylI28SHwbDO5qYw/N/GusC8Zh6WdHgyFv8ARalvgIWE4QMCo/xmq00WdyxJUTxMwtFPvb+hMqxKVyTZ8AiIeS1+cySebShJCU5+in4RticztxaiceCjfFgGkzJeA3ph7a2u+Y8zBR+AuJM1Q==--CjVchd12b8cbF22G--UtTqVaNRbp5q+Ql0jLbngA==
```
charvp: Is this used before the main `credentials.yml.enc`?

This file also seems a little short to me. Naos also has the Google/Smartschool/Office365 secrets.

config/environments/staging.rb
rien: Naos isn't sending error messages (At least for two years: https://github.com/dodona-edu/dodona/blame/d42b2df14f3739e8e0e73d857aa6e920908f87de/config/environments/development.rb#L89).

Do we want to change this to `:sendmail`?

config/credentials/staging.yml.enc
rien: This overwrites the main credentials with the secret key currently used on naos. So all the other secrets will stay the same.

config/environments/staging.rb
charvp: I know Naos didn't send anything until now, but I didn't really look at it before now. Not sure if we want it to send emails. @bmesuere ?

config/credentials/staging.yml.enc
charvp: :+1:

config/environments/staging.rb
bmesuere: I don't think we want mails from naos

PR #1319, opened 2019-09-27
Symlink staging.key instead of whole credentials directory

rien:


config/deploy.rb
```diff
@@ -23,10 +23,10 @@
 # set :pty, true
 
 # Default value for :linked_files is []
-set :linked_files, fetch(:linked_files, []).push('config/master.key')
+set :linked_files, fetch(:linked_files, []).push('config/master.key', 'config/credentials/staging.key')
```
charvp: Didn't see this in the previous PR, but this seems like something we only want to do in the staging environment.

rien: :+1:

PR #1327, opened 2019-09-30
Only use judged submission to determine exercise and series status

### LINKED ISSUE #1322: exercises that are being evaluated, are displayed as incorrect
The exercise below contains no previous submissions. While the exercise was being evaluated, the status at the course overview was "Incorrect" (see image below) instead of "Evaluating..."

image

charvp:
Also fixes the `users_tried` and `attempted_exercises` to only take judged submissions into account.

- [x] Tests were added
- No relevant documentation.

Closes #1322.

### DIFF
--- a/test/models/transient/exercise_summary.rb
+++ b/test/models/transient/exercise_summary.rb
@@ -282,6 +282,28 @@ module ExerciseSummaryTests
     end
   end
 
+  class NotYetJudged < ExerciseSummaryTest
+    setup do
+      submit status: :running
+    end
+
+    test 'should not have been submitted' do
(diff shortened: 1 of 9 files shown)


test/models/transient/exercise_summary.rb
```diff
@@ -282,6 +282,28 @@ def summary
     end
   end
 
+  class NotJudgedYet < ExerciseSummaryTest
```
bmesuere: NotYetJudged?

charvp:
Cache invalidation happens on every submission update, so will still work and the frequency won't change.

PR #1330, opened 2019-09-30
sanitize submission result

### LINKED ISSUE #1139: XSS risks from untrusted exercise sources

rien:
This pull request sanitizes submission results (i.e. we prevent additional `<script>` in judge output from being executed).

TODO:
- [x] Rebase with develop once #1234 is merged
- [x] Test if all judge output is properly escaped
- [x] Documentation at https://github.com/dodona-edu/dodona-edu.github.io/pull/10

Fixes #1139.

### DIFF
--- a/app/helpers/renderers/feedback_table_renderer.rb
+++ b/app/helpers/renderers/feedback_table_renderer.rb
@@ -22,7 +22,7 @@ class FeedbackTableRenderer
     @code = submission.code
     @exercise = submission.exercise
     @exercise_id = submission.exercise_id
-    @programming_language = sanitize @exercise.programming_language&.editor_name
+    @programming_language = @exercise.programming_language&.editor_name
   end
 
   def parse
@@ -70,7 +70,7 @@ class FeedbackTableRenderer
             @builder.a(href: "##{(t[:description] || 'test').parameterize}-#{i}", 'data-toggle': 'tab') do
               @builder.text!((t[:description] || 'Test').upcase_first + ' ')
               @builder.span(class: 'badge') do
-                @builder << tab_count(t)
+                @builder.text! tab_count(t)
               end
             end
           end
@@ -176,8 +176,6 @@ class FeedbackTableRenderer
     end
   end
 
-  def testcase_icons(tc); end
-
   def testcase(tc)
     @builder.div(class: "testcase #{tc[:accepted] ? 'correct' : 'wrong'}") do
       testcase_content(tc)
@@ -187,7 +185,6 @@ class FeedbackTableRenderer
   def testcase_content(tc)
     @builder.div(class: 'col-xs-12 description') do
       @builder.div(class: 'indicator') do
-        testcase_icons(tc)
         tc[:accepted] ? icon_correct : icon_wrong
       end
       message(tc[:description]) if tc[:description]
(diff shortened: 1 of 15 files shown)


app/helpers/renderers/feedback_table_renderer.rb
```diff
@@ -108,7 +108,6 @@ def tabs(submission)
   def tab_count(t)
     return '' if t[:badgeCount].nil?
     return '' if t[:badgeCount].zero?
-    return '' if t[:badgeCount] == '0'
```
bmesuere: Any reason why this was removed? We don't want to show a badge when there are 0 remarks/wrong tests.

charvp: Because the badge count can never be a string: the `.zero?` call above will throw an error if it is. (I noticed this because at first this was sanitized and I added a test for it.)

charvp:
Note: we get deprecation warnings because of [this](https://github.com/rails/rails-html-sanitizer/pull/100) but it should be fixed in the next version of `rails-html-sanitizer`.

charvp:
I've added tests for and fixed the stripping the content of tags in 03effe1. It's quite hard to test this in the tests of the renderers though, since we actually want to escape most of the HTML in the JSON (e.g. in the diff table), which means the content will still be there (unchanged, but escaped).

test/files/output.json
```diff
@@ -182,10 +194,19 @@
...
-                    "description": "test whether the user can XSS",
-                    "expected": "<script>alert('Your mother was a hamster and your father smelt of elderberries.')</script>",
-                    "generated": "<script>alert('Your mother was a hamster and your father smelt of elderberries.')</script>"
+                  "accepted": false,
+                  "description": "test whether the user can XSS",
+                  "expected": "<script>alert('Your mother was a hamster and your father smelt of elderberries.')</script><script>alert('test')</script><script>alert('test')</script><script>alert('test')</script><script>alert('test')</script><script>alert('test')</script><script>alert('test')</script><script>alert('test')</script><script>alert('test')</script><script>alert('test')</script><script>alert('test')</script><script>alert('test')</script><script>alert('test')</script><script>alert('test')</script><script>alert('test')</script><script>alert('test')</script><script>alert('test')</script><script>alert('test')</script><script>alert('test')</script><script>alert('test')</script><script>alert('test')</script><script>alert('test')</script>",
```
rien: :+1:

PR #1337, opened 2019-10-01
fix users being unable to subscribe to hidden courses

charvp:
The condition for adding the secret token to the URL was wrong, resulting in the secret not being added on courses that were open and hidden. This fixes the condition for the secret. The condition for the checking of the secret was also wrong, so I updated that one as well.

- [x] Tests were added
- No relevant documentation changes.

### DIFF
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -23,7 +23,7 @@ class CoursesController < ApplicationController
   # GET /courses/1
   # GET /courses/1.json
   def show
-    if @course.hidden? || (@course.visible_for_institution? && current_user&.institution != @course.institution)
+    if @course.secret_required?(current_user)
       redirect_unless_secret_correct
       return if performed?
     end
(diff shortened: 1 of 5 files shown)


app/controllers/courses_controller.rb
```diff
@@ -179,7 +179,7 @@ def unsubscribe
   end
 
   def subscribe
-    redirect_unless_secret_correct if @course.hidden?
+    redirect_unless_secret_correct if @course.hidden? || (@course.visible_for_institution? && current_user&.institution != @course.institution)
```
bmesuere: Not too fond of these complex logical tests, especially not in combination with modifiers. Would it make sense to move this check to the model?

test/controllers/courses_controller_test.rb
```diff
@@ -469,4 +469,13 @@ def with_users_signed_in(users)
     post courses_url, params: { course: { name: new_course.name, description: new_course.description, visibility: new_course.visibility, registration: new_course.registration, teacher: new_course.teacher }, copy_options: { base_id: course.id }, format: :json }
     assert_equal 0, Course.find(JSON.parse(response.body)['id']).series.count
   end
+
+  test 'hidden course page shown to unsubscribed student should include registration url with secret' do
```
bmesuere: Technically, the institution case isn't tested

PR #1345, opened 2019-10-02
upgrade to Ruby 2.6.5

charvp:
I don't think any of the [CVE's](https://www.ruby-lang.org/en/news/2019/10/01/ruby-2-6-5-released/) apply to us, but no reason to stay behind.

- No relevant tests to add.
- No relevant documentation to change.


bmesuere:
If we deploy this, won't it break because that rails version isn't installed on the servers yet?

charvp:
Yes, a similar PR will have to be made to the ansible repo.

PR #1350, opened 2019-10-03
notify us if there are requests that take over 2 seconds

charvp:
The StatisticsController is excluded since we know that there are long-running requests there. (This is something that should probably be fixed.)

- No tests were added.
- No relevant documentation changes.


ninewise:
Does this include the webhooks? They might take longer than 2 seconds, too.

charvp:
> Does this include the webhooks? They might take longer than 2 seconds, too.

Only the pull is done in the foreground for the webhooks, so they shouldn't take longer than 2 seconds.

PR #1357, opened 2019-10-04
disallow "command" in full output json schema

### LINKED ISSUE #1329: failed to parse partial output
The following series of json cannot be parsed by the resultcontstructor:

```rb
rc = ResultConstructor.new('nl')
rc.feed('{ "command": "start-judgement" }')
rc.feed('{ "title": "Test", "command": "start-tab" }')
rc.feed('{ "description": { "format": "code", "description": "..." }, "command": "start-context" }')
rc.feed('{ "description": { "format": "plain", "description": "" }, "command": "start-testcase" }')
rc.feed('{ "expected": "70", "command": "start-test" }')
rc.feed('{ "generated": "45", "status": "wrong", "accepted": false, "command": "close-test" }')
rc.feed('{ "accepted": false, "command": "close-testcase" }')
```

I think this is valid partial-output and I should check why it crashes.

ninewise:
So we avoid confusion with the partial output json schema, which
requires a "command" enum. This assumes there are no existing full
output json results with a "command" key in the json root.

- [x] Tests were added
- [x] No documentation update is required as it's mentioned nowhere you can use extra keys in the judge output.

Closes #1329.

### DIFF
--- a/public/schemas/judge_output.json
+++ b/public/schemas/judge_output.json
@@ -12,6 +12,11 @@
         "messages":    { "type": "array", "items": { "$ref": "#/definitions/message" } },
         "groups":      { "type": "array", "items": { "$ref": "#/definitions/tab" } },
         "annotations": { "type": "array", "items": { "$ref": "#/definitions/annotation" } }
+      },
+      "not": {
+        "properties": {
+          "command": { "type": "string" }
+        }
       }
     },
     "tab": {
--- a/public/schemas/judge_output.json
+++ b/public/schemas/judge_output.json
@@ -14,6 +14,7 @@
         "annotations": { "type": "array", "items": { "$ref": "#/definitions/annotation" } }
       },
       "not": {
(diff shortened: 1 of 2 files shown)


charvp:
Did you check some pythia submission results that they don't contain `command` as a key in the root?

ninewise:
Not yet; I was planning to ask before I ran a script to check all JSON's on Dodona.

charvp:
I have started a script to check 20% of submissions on whether they have a `command` field in their root that is a string.
```
Submission.where("MOD(id,5) = 0").find_each do |s|
    if s.result["command"].is_a? String
        ExceptionNotifier.notify_exception Exception.new('Submission result with String `command` found')
    end
    sleep (0.1).seconds
end
```
It should be finished in about a day.

charvp:
The script didn't find any JSON's with command in the root (as a string).

PR #1361, opened 2019-10-05
Don't show progress header on xs screens

### LINKED ISSUE #1360: Group progress column not shown on mobile
The progress header is shown, but the column itself is not.
Screenshot_20191005-083659

charvp:
image

- No relevant tests to add
- No relevant documentation changes

Closes #1360.


pdawyndt:
Shouldn't the labels be left out as well?

PR #1395, opened 2019-10-11
clean up submission statistics

charvp:
This also fixes the summertime/wintertime bug.

- [x] Tests were added
- [x] No relevant documentation changes

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -84,7 +84,7 @@ Rails.application.configure do
   end
 
   # Use a different cache store in production.
-  config.cache_store = :mem_cache_store, 'elysium.ugent.be', {namespace: :"1"}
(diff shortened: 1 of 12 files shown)


config/environments/production.rb
```diff
@@ -84,7 +84,7 @@
   end
 
   # Use a different cache store in production.
-  config.cache_store = :mem_cache_store, 'elysium.ugent.be', {namespace: :"1"}
+  config.cache_store = :mem_cache_store, 'elysium.ugent.be', {namespace: :"2"}
```
bmesuere: This will invalidate all caches and not only the ones changed?

charvp: Yes, but except for moving this namespacing to the cache string templates there's little we can do about that. (And I personally don't really like that solution.)

PR #1399, opened 2019-10-14
don't escape the feedback table

charvp:


app/assets/javascripts/exercise.js
```diff
@@ -123,6 +123,7 @@ function initExerciseShow(exerciseId, programmingLanguage, loggedIn, editorShown
         }
         swapActionButtons();
         initDeadlineTimeout();
+        fixSubmissionTableLinks();
```
bmesuere: maybe "enable" is a better name than "fix"?

charvp:
Loading indicator: https://www.cvpetegem.be/public/recording.mp4

app/assets/javascripts/exercise.js
```diff
@@ -213,8 +213,33 @@ function initExerciseShow(exerciseId, programmingLanguage, loggedIn, editorShown
...
+        $("#submission-wrapper").html("<center><i class=\"mdi mdi-loading mdi-spin\"></i></center>");
+        feedbackLoaded(submissionId);
+        fetch(updateURLParameter(url, "format", "js"), {
+            headers: {
+                "accept": "text/javascript",
+                "x-csrf-token": $("meta[name=\"csrf-token\"]").attr("content"),
+                "x-requested-with": "XMLHttpRequest",
+            },
+            credentials: "same-origin",
+        }).then(resp => resp.text()).then(data => {
```
bmesuere: Maybe check for a resp.ok here and show an error if it fails? Otherwise this will simply crash and keep showing the spinner

PR #1414, opened 2019-10-17
render student code manually in the feedback table instead of using the ACE editor

### LINKED ISSUE #1388: stop using the ACE editor to render the submission in the feedback table
The ACE editor is a limitation to render richer feedback messages on lines of a submission.

We can add syntax highlighting to the code by using Rouge.

Pro:
- more flexible
- lighter to load
- will work on ipad (although that issue might be fixed with the release of ipad OS)

Con:
- no more code collapsing

MrKickkiller:
The pull request fixes the issue laid out in #1388 

TODO:
- [x] Styling
- [x] Adding the annotations via Javascript/Typescript

Current state. Will update
image

References:
image
image

- [x] Tests were added
- [x] No relevant documentation is requried

Closes #1388 .


pdawyndt:
@MrKickkiller could you add some screenshot?

app/assets/javascripts/feedback_code_table.ts
```diff
@@ -0,0 +1,65 @@
+export class FeedbackCodeTable {
+    table: Element;
+    messages;
```
bmesuere: If you want to store messages as a property in the object, you should initialize it as an empty array in the constructor

app/assets/javascripts/feedback_code_table.ts
```diff
@@ -0,0 +1,65 @@
...
+    annotationCounter: number = 0;
```
bmesuere: Do you need this? Can't you use the length of messages?

app/assets/javascripts/feedback_code_table.ts
```diff
@@ -0,0 +1,65 @@
...
+
+    constructor(feedbackTableSelector = ".feedback-code-table") {
+        this.table = document.querySelector(feedbackTableSelector);
+
+        if (this.table === null) {
+            console.error("The feedback table could not be found");
+        }
+    }
+
+    initAnnotations(messages: object[]): void {
```
bmesuere: I would rename it to addAnnotations and define a message type in typescript

app/assets/javascripts/feedback_code_table.ts
```diff
@@ -0,0 +1,65 @@
...
+        this.messages = messages;
```
bmesuere: add to this instead of replace, so we can call it multiple times

app/assets/javascripts/feedback_code_table.ts
```diff
@@ -0,0 +1,65 @@
...
+        for (const message of this.messages) {
+            // Linter counts from 0, rouge counts from 1
+            const correspondingLine = this.table.querySelector(`#line-${message.row + 1}`);
+            const annotationRow = this.createAnnotation(message);
+
+            correspondingLine.parentElement.insertBefore(annotationRow, correspondingLine.nextSibling);
+        }
+    }
+
+    private createAnnotationRow(lineNumber: number): object {
```
bmesuere: return type object is a bit generic

app/assets/javascripts/feedback_code_table.ts
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    initAnnotations(messages: object[]): void {
+        this.messages = messages;
+        for (const message of this.messages) {
+            // Linter counts from 0, rouge counts from 1
+            const correspondingLine = this.table.querySelector(`#line-${message.row + 1}`);
+            const annotationRow = this.createAnnotation(message);
+
+            correspondingLine.parentElement.insertBefore(annotationRow, correspondingLine.nextSibling);
```
bmesuere: did you check if you could get rid of this parent and sibling stuff?

app/assets/javascripts/submission.js
```diff
@@ -57,12 +57,11 @@ function initSubmissionShow() {
...
             }
             if (line !== undefined) {
-                const Range = ace.require("ace/range").Range;
-                const editor = ace.edit("editor-result");
-                if (typeof currentMarkerId !== "undefined") {
-                    editor.getSession().removeMarker(currentMarkerId);
-                }
-                currentMarkerId = editor.getSession().addMarker(new Range(line - 1, 0, line, 0), "ace_active-line tab-link-marker", "line", false);
+                // Reset
+                $(".feedback-code-table tr.lineno.marked").removeClass("marked");
```
bmesuere: Marking and clearing highlights should be functions in the FeedbackCodeTable class. Adding and remove classes is an implementation detail and shouldn't be exposed outside that class.

app/assets/stylesheets/components/feedback_code_table.less
```diff
@@ -0,0 +1,70 @@
...
+      }
+
+      .annotation.info {
+        border-left-color: @info-500;
+      }
+    }
+  }
+
+  .lineno.marked {
+    background-color: @accent-300;;
```
bmesuere: I think it's better to use warning-300 or something like that.

app/assets/stylesheets/components/feedback_code_table.less
```diff
@@ -0,0 +1,70 @@
...
+  }
+}
+
+// Mixin such that this could be reused.
+// TODO: Move to utility css file?
+.user-select(@option) {
```
bmesuere: No need to define a mixin for this. We have an autoprefixer that compiles the assets. Simply using user-select: foobar should to the trick

app/assets/javascripts/feedback_code_table.ts
```diff
@@ -0,0 +1,90 @@
+class Message {
```
bmesuere: use interface and remove constructor

app/assets/javascripts/feedback_code_table.ts
```diff
@@ -0,0 +1,90 @@
...
+    constructor(feedbackTableSelector = "table.feedback-code-table") {
+        this.table = document.querySelector(feedbackTableSelector) as HTMLTableElement;
+        this.messages = [];
+
+        if (this.table === null) {
+            console.error("The feedback table could not be found");
+        }
+    }
+
+    addAnnotations(messages: object[]): void {
```
bmesuere: use `Message[]` here

app/assets/javascripts/feedback_code_table.ts
```diff
@@ -0,0 +1,90 @@
...
+        const newMessages: Message[] = [];
```
bmesuere: no need to create a separate variable

app/assets/javascripts/feedback_code_table.ts
```diff
@@ -0,0 +1,90 @@
...
+        let idOffset: number = this.messages.length;
+        for (const message of messages) {
+            newMessages.push(new Message(message, idOffset + 1));
+            idOffset += 1;
+        }
+        this.messages.push(...newMessages);
```
bmesuere: I don't think you need this

app/assets/javascripts/feedback_code_table.ts
```diff
@@ -0,0 +1,90 @@
...
+
+        for (const message of newMessages) {
+            // Linter counts from 0, rouge counts from 1
+            const correspondingLine: HTMLTableRowElement = this.table.querySelector(`#line-${message.row + 1}`);
+            this.createAnnotation(message, correspondingLine.rowIndex + 1, message.row + 1);
```
bmesuere: add message to this.messages after succesful creation

app/assets/javascripts/feedback_code_table.ts
```diff
@@ -0,0 +1,90 @@
...
+
+        const textNode: Text = document.createTextNode(message.text.replace(/-*$/, ""));
+        annotationCell.classList.add(message.type);
+        annotationCell.appendChild(textNode);
+
+        annotationTD.appendChild(annotationCell);
+        return annotationRow;
+    }
+
+    unmarkAllAnnotations(): void {
```
bmesuere: This can become confusing since there are two separate things that are called annotations. Maybe rename this to `clearHighlights` and the next funtion to `addHighlight` or `highlightLine`

app/assets/stylesheets/components/feedback_code_table.less
bmesuere: We don't want this to be dependant on the Dodona theme, but always something yellow-ish

app/helpers/renderers/feedback_code_renderer.rb
```diff
@@ -0,0 +1,30 @@
+class FeedbackCodeRenderer
+  require 'json'
+
+  def initialize(code, programming_language, messages, builder)
+    @programming_language = programming_language
+    @code = code
+    @builder = builder
+    @messages = messages
```
bmesuere: use same order as parameters

MrKickkiller:
@bmesuere 
image

Requested changes 4 - 6 give this result

MrKickkiller:
@bmesuere 
image

Source: https://github.com/dracula/pygments

app/assets/javascripts/exercise.js
```diff
@@ -138,8 +138,10 @@ function initExerciseShow(exerciseId, programmingLanguage, loggedIn, editorShown
         });
 
         $("#submission-copy-btn").click(function () {
-            const submissionSource = ace.edit("editor-result").getValue();
-            editor.setValue(submissionSource, 1);
+            const submissionCode = [];
+            document.querySelectorAll(".lineno .rouge-code")
+                .forEach( codeLine => submissionCode.push(codeLine.textContent));
```
bmesuere: I would move these lines as a function to the new code listing class since it depends on implementation details of that listing.

app/assets/javascripts/feedback_code_table.ts
```diff
@@ -77,4 +77,11 @@ export class FeedbackCodeTable {
         const toMarkAnnotationRow = this.table.querySelector(`tr.lineno#line-${lineNr}`);
         toMarkAnnotationRow.classList.add(this.markingClass);
     }
+
+    setSubmissionEditorCode(editor): void {
```
bmesuere: rename to `getCode` and return a string

app/assets/javascripts/feedback_code_table.ts
```diff
@@ -77,4 +77,11 @@ export class FeedbackCodeTable {
...
+        const submissionCode = [];
+        document.querySelectorAll(".lineno .rouge-code")
+            .forEach( codeLine => submissionCode.push(codeLine.textContent));
+        editor.setValue(submissionCode.join(""), 1);
```
bmesuere: `editor.setvalue...` should stay in exercise.js

app/assets/javascripts/feedback_code_table.ts
```diff
@@ -0,0 +1,87 @@
+interface Message {
+    id?: number;
+
+    type: string;
+    text: string;
+    row: number;
+}
+
+export class FeedbackCodeTable {
```
bmesuere: maybe CodeListing is a better name?

app/assets/javascripts/feedback_code_table.ts
```diff
@@ -0,0 +1,87 @@
...
+    table: HTMLTableElement;
+    messages: Message[];
+
+    private markingClass: string = "marked";
```
bmesuere:
```suggestion
    private readonly markingClass: string = "marked";
```

app/assets/javascripts/feedback_code_table.ts
```diff
@@ -0,0 +1,87 @@
...
+    id?: number;
+
+    type: string;
+    text: string;
+    row: number;
+}
+
+export class FeedbackCodeTable {
+    table: HTMLTableElement;
+    messages: Message[];
```
bmesuere: these should also be private I think?

app/assets/javascripts/feedback_code_table.ts
```diff
@@ -0,0 +1,87 @@
...
+
+    clearHighlights(): void {
+        const markedAnnotations = this.table.querySelectorAll(`.tr.lineno.${this.markingClass}`);
+        markedAnnotations.forEach(markedAnnotation => {
+            markedAnnotation.classList.remove(this.markingClass);
+        });
+    }
+
+    highlightLine(lineNr: number): void {
+        this.clearHighlights();
```
bmesuere: As discussed, this should be here (we want to be able to highlight multiple lines). This line should be in the link handler before highlightLine gets called.

MrKickkiller: I knew I was forgetting a thing.

app/assets/stylesheets/components.css.less
```diff
@@ -3,6 +3,7 @@
 @import "components/card.css.less";
 @import "components/dropdown.css.less";
 @import "components/favorite.css.less";
+@import "components/code_listing";
```
charvp: I would import `code_listing.css.less` here, just for consistency. Unless there's a reason not to do that in which case it should be explained with a comment.

app/helpers/application_helper.rb
```diff
@@ -119,7 +119,7 @@ def sanitize(html)
     tags = Rails::Html::SafeListSanitizer.allowed_tags.to_a
     tags += %w[table thead tbody tr td th colgroup col style svg circle line rect]
     attributes = Rails::Html::SafeListSanitizer.allowed_attributes.to_a
-    attributes += %w[style target data-toggle data-parent id x1 y1 x2 y2 stroke stroke-width fill cx cy r]
+    attributes += %w[style target data-toggle data-parent data-tab data-line data-element id x1 y1 x2 y2 stroke stroke-width fill cx cy r]
```
charvp: The cherry-pick of this change should probably be removed since it's already on develop.

app/helpers/renderers/feedback_code_renderer.rb
```diff
@@ -0,0 +1,30 @@
...
+    line_formatter = Rouge::Formatters::HTML.new
+    table_formatter = Rouge::Formatters::HTMLLineTable.new line_formatter, table_class: 'code-listing highlighter-rouge'
+
+    lexer = (Rouge::Lexer.find(@programming_language) || Rouge::Lexers::PlainText).new
+    lexed_c = lexer.lex(@code)
+
+    @builder << table_formatter.format(lexed_c)
+
+    @builder.script(type: 'application/javascript') do
+      @builder << 'var messages = '
```
charvp: You can use modern JS here IMO. We don't support IE.

app/javascript/packs/submission.js
```diff
@@ -1,4 +1,5 @@
-import { initSubmissionShow, loadResultEditor } from "submission.js";
+import { initSubmissionShow } from "submission.js";
+import { CodeListing } from "../../assets/javascripts/code_listing";
```
charvp: You don't need the `../../assets/javascripts/` part. So just make it like the line above.

PR #1416, opened 2019-10-18
fix synchronization issues around timeout detection

### LINKED ISSUE #1120: time vs. memory limit exceeded
The heuristic to differ between a time and memory limit exceeded still makes some mistakes (https://dodona.ugent.be/nl/submissions/2046707/, https://dodona.ugent.be/nl/submissions/3398007/). I think the cause might be following execution path:

- timer thread starts
- main thread starts and waits for the container
- timer threads stops container
- main thread continues, and notices the timer thread is still running (which makes it conclude there's no timeout)
- timer thread finishes

The results for above two exercises seem consistent, though. Needs more investigation.

Could be solved by running the timer and the container both in a thread, putting either true (timer hread) or false (container thread) in a timeout MVar (if ruby has something similar), with the main thread waiting for the MVar to be filled. The actual container could be killed afterwards (because container.stop isn't atomic).

_[Original issue](https://github.ugent.be/dodona/dodona/issues/1120) by @ninewise on Tue Jul 02 2019 at 14:04._

charvp:
It seems that in most (all?) cases, the main thread continues when `container.stop` is called from the timer thread. This results in the timer thread not having finished yet (and therefore not having set its value to `true`). This made us misreport the reason the container was stopped.

By synchronizing around a mutex we can be sure the timer thread gets to set the `timeout` value if it was the reason the container stopped.

- [x] Tests were added
- [x] No relevant documentation changes

Closes #1120.

### DIFF
--- a/test/runners/submission_runner_test.rb
+++ b/test/runners/submission_runner_test.rb
@@ -175,20 +175,6 @@ class SubmissionRunnerTest < ActiveSupport::TestCase
                       accepted: false
   end
 
-  test 'dockers killed by 15 before timeout should result in memory limit exceeded' do
-    Thread.stubs(:new).returns(mock(kill: nil, value: nil))
-    evaluate_with_stubbed_docker output: nil, status_code: 128 + 15
-    assert_submission status: 'memory limit exceeded',
-                      accepted: false
-  end
-
-  test 'dockers killed by 15 at timeout should result in time limit exceeded' do
-    Thread.stubs(:new).returns(mock(kill: nil, value: true))
-    evaluate_with_stubbed_docker output: nil, status_code: 128 + 15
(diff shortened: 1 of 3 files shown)


test/runners/submission_runner_test.rb
```diff
@@ -205,4 +205,59 @@ def assert_submission(status: nil, summary: nil, message_includes: nil, accepted
...
+
+    def start; end
+
+    def stop
+      @running_cond.signal
+    end
+
+    def attach(_)
+      @running.synchronize do
+        @running_cond.wait(@running)
```
beardhatcode: Maybe there should be a delay here to simulate the time it takes to kill the container?

PR #1417, opened 2019-10-18
implement safe exercise descriptions in API

thepieterdc:
This pull request adds exercise descriptions to the API, via the sandbox.

**/exercise/x.json**
```json
{
"description": "http://sandbox.url/exercises/x/description/?token=some_token"
}
```

- [x] Tests were added
- [x] No relevant documentation updates

### DIFF
--- a/app/views/exercises/_exercise.json.jbuilder
+++ b/app/views/exercises/_exercise.json.jbuilder
@@ -9,5 +9,6 @@ if current_user
   json.has_solution exercise.last_submission(current_user).present?
   json.has_correct_solution exercise.last_correct_submission(current_user).present?
 end
-json.description exercise.description_localized
+json.description 'Deprecated. Use description_url'
+json.description_url description_exercise_url(exercise, token: exercise.access_token)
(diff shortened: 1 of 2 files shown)


app/views/exercises/_exercise.json.jbuilder
```diff
@@ -9,5 +9,5 @@ if current_user
   json.has_solution exercise.last_submission(current_user).present?
   json.has_correct_solution exercise.last_correct_submission(current_user).present?
 end
-json.description exercise.description_localized
+json.description description_exercise_url(exercise, token: exercise.access_token).html_safe
```
charvp: I don't know if this `html_safe` call is necessary in this context.

charvp: Also, I know this would change the API, but I think ending with a `_url` postfix is more common (also in this codebase). Not necessarily saying it has to change to that, but it's a trade-off to think about.

thepieterdc: I have added an extra field named `description_url`, but I have also kept the original `description` field and added a deprecation notice, instead of removing the field directly.

PR #1425, opened 2019-10-21
dockerisation

thepieterdc:
As requested by @ninewise, this PR adds a Dockerfile to run Dodona in development mode. Ideally, this should be added to the GitHub CI to allow images to be generated automatically when commits are pushed to the develop branch.

**Instructions:**
```
docker build . -t dodona
docker run -it -p 3000:3000 dodona
```
Dodona is now accessible at http://localhost:3000

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,77 @@
+# Set the base image.
+FROM ruby:2.6.5-stretch
+
+################################################################################
+################################### PACKAGES ###################################
+################################################################################
+
+# Refresh the packages list.
+RUN apt-get update
+
+# Install MySQL server.
+RUN DEBIAN_FRONTEND=noninteractive apt-get install -y mysql-server
+
+# Install dockerize.
+RUN wget https://github.com/jwilder/dockerize/releases/download/v0.6.0/dockerize-alpine-linux-amd64-v0.6.0.tar.gz \
+    && tar -C /usr/local/bin -xzvf dockerize-alpine-linux-amd64-v0.6.0.tar.gz \
+    && rm dockerize-alpine-linux-amd64-v0.6.0.tar.gz
+
+# Install NodeJS.
+RUN curl -sL https://deb.nodesource.com/setup_12.x | bash - \
+    && apt install -y nodejs
+
+################################################################################
+#################################### FILES #####################################
+################################################################################
+
+# Copy the application files and enter that directory.
+COPY . /dodona
+WORKDIR dodona
+
+# Patch the seeds to clone over https instead of ssh.
+RUN sed -i 's+git@github.com:dodona-edu/+https://github.com/dodona-edu/+g' db/seeds.rb
+# Patch the seeds to use the java judge since that's open source.
+RUN sed -i 's/judge-pythia.git/judge-java12.git/' db/seeds.rb
+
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -0,0 +1,77 @@
...
+RUN DEBIAN_FRONTEND=noninteractive apt-get install -y mysql-server
+
+# Install dockerize.
+RUN wget https://github.com/jwilder/dockerize/releases/download/v0.6.0/dockerize-alpine-linux-amd64-v0.6.0.tar.gz \
+    && tar -C /usr/local/bin -xzvf dockerize-alpine-linux-amd64-v0.6.0.tar.gz \
+    && rm dockerize-alpine-linux-amd64-v0.6.0.tar.gz
+
+# Install NodeJS.
+RUN curl -sL https://deb.nodesource.com/setup_12.x | bash - \
+    && apt install -y nodejs
```
charvp: Alpine has a good version of node and yarn in its repositories :eyes:

Dockerfile
```diff
@@ -0,0 +1,77 @@
...
+################################################################################
+
+# Copy the application files and enter that directory.
+COPY . /dodona
+WORKDIR dodona
+
+# Patch the seeds to clone over https instead of ssh.
+RUN sed -i 's+git@github.com:dodona-edu/+https://github.com/dodona-edu/+g' db/seeds.rb
+# Patch the seeds to use the java judge since that's open source.
+RUN sed -i 's/judge-pythia.git/judge-java12.git/' db/seeds.rb
```
charvp: If this is meant for development this will break all seeded exercises that use the python judge (which is most of them).

Dockerfile
```diff
@@ -0,0 +1,77 @@
...
+
+################################################################################
+################################ START THE SERVER ##############################
+################################################################################
+
+# Expose the webserver to the host.
+EXPOSE 3000
+
+# Start the database server
+CMD /etc/init.d/mysql start \
```
charvp: If this is meant for development some persistence might be desired. I would split the mysql server into a separate docker container and write a docker-compose file that allows users to just perform a `docker-compose up -d` that either restores their previous work or creates a clean environment if there is no previous work.

thepieterdc:
> This seems to work great, I really appreciate this. If you want, I can have a look at @charvp's remarks and try to fix them.

Feel free to do so :)

Dockerfile
thepieterdc: I am aware of this, but the python judge is currently not publicly accessible, so this would require an SSH key to be embedded in the docker image.

Dockerfile
```diff
@@ -0,0 +1,77 @@
...
+# Install NodeJS.
+RUN curl -sL https://deb.nodesource.com/setup_12.x | bash - \
+    && apt install -y nodejs
+
+################################################################################
+#################################### FILES #####################################
+################################################################################
+
+# Copy the application files and enter that directory.
+COPY . /dodona
```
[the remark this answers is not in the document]
thepieterdc: Primarily meant for integration testing (I need this :eyes: ), but can be used for local development as well if the files are not served from mounted volumes.

bmesuere:
I suggest we close this one for now?

PR #1459, opened 2019-11-04
add CSP headers

### LINKED ISSUE #1418: add CSP headers to description iframes
Since the `X-Frame-Options: allow-from` directive [is deprecated](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Frame-Options), we should add a Content Security Policy [`frame-ancestors`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/frame-ancestors) directive to support modern browsers. Firefox nightly already errors on the `allow-from` directive.

rien:
This pull request adds `Content-Security-Policy-Headers` (and other secure headers) using the [secure_headers](github.com/twitter/secure_headers) gem.

The policy for our own webapp is strict. We can only include js, images, stylesheets, ... if they are explicitly whitelisted in the config file.

The policy for descriptions (iframes) is rather lax. Allowing the inclusion of any script or other resource loaded over https.

Closes #1418


config/initializers/secure_headers.rb
```diff
@@ -0,0 +1,49 @@
+SecureHeaders::Configuration.default do |config|
+  config.csp = {
+    preserve_schemes: true,
+
+    default_src: %w('self'),
+    frame_ancestors: %w('none'),
+    connect_src: %w('self' https://pandora.ugent.be),
```
charvp: Shouldn't firebase and GA be included here?

rien: Correct. Fixed with 9be558b.

rien:
CSP Violation reporting has been added. Each violation should now create an `Event` and send an email.
I have deployed this branch on Naos again.

config/environments/development.rb
```diff
@@ -89,7 +89,7 @@
   # config.file_watcher = ActiveSupport::EventedFileUpdateChecker
 
   config.middleware.use ExceptionNotification::Rack,
-                        ignore_if: ->(env, _exception) { env['HTTP_HOST'] == 'localhost:3000' || env['HTTP_HOST'] == 'dodona.localhost:3000' },
+                        #ignore_if: ->(env, _exception) { env['HTTP_HOST'] == 'localhost:3000' || env['HTTP_HOST'] == 'dodona.localhost:3000' },
```
charvp: This line should probably be enabled again.

config/initializers/content_security_policy.rb
```diff
@@ -4,19 +4,41 @@
...
+                       'http://localhost:3035', 'ws://localhost:3035'
+  else
+    policy.connect_src :self, 'https://pandora.ugent.be',
+                       'https://dodona-cea23.firebaseio.com',
+                       'https://www.google-analytics.com'
+  end
+
+  policy.font_src    :self, 'https://fonts.gstatic.com',
+                     'https://cdn.materialdesignicons.com',
+                     'https://cdnjs.cloudflare.com'
```
charvp: Where do we request fonts/images/scripts from cloudflare?

PR #1464, opened 2019-11-07
fix slow exercise description

### LINKED ISSUE #1427: exercise descriptions load very slow sometimes
We parse the html of the description, which makes it slow. Check if we can do this in JS maybe?

rien:
This pull request removes the regex replacing relative media URL's, which was unbearably slow when rendering a description containing a lot of attributes. This is now replaced by having a `<base>` HTML tag which makes all links in the description relative to an URL containing the  exercise access token.

Now we also don't have to maintain that huge regex anymore.

Before:
```
Completed 200 OK in 5864ms (Views: 5844.0ms | ActiveRecord: 2.2ms | Allocations: 355583)
```

After:
```
Completed 200 OK in 236ms (Views: 213.4ms | ActiveRecord: 3.5ms | Allocations: 298984)
````

The following changes were applied to the tests:
- The `MediaPathContextualizerTest` testing this regex was removed.
- The `ExerciseHelperTest` was changed because we do not translate relateive media URL's ourselves anymore.
- Tests were added to check if the right base URL is included.

Closes #1427.


rien:
@charvp I was unaware `feedback_table_renderer` used this functionality. I apparently hadn't removed the URL rewriting completely, hence the tests didn't break.

I have re-added the tests for the rewriting, because it is used at the moment. But this method is probably slowing the submission page down a lot. I will try to look if a nokogiri-based solution would be more efficient.

rien:
I've just looked at the flamegraph for a [very slow submission result](https://dodona.ugent.be/nl/submissions/3919493/) and it's 70% `contextualize_media_paths`, specifically [this rails method](https://github.com/rails/rails/blob/master/activesupport/lib/active_support/core_ext/string/output_safety.rb#L264) which uses metaprogramming and other ruby magic, which is of course very slow.

Nokogiri to the rescue!

PR #1468, opened 2019-11-10
Allow adding annotations on a submission

MrKickkiller:
This pull request allows users to add custom annotations when reviewing submissions

image

- [x] Tests were added


bmesuere:
Old review comments that were missing, might not be applicable anymore

- remove markdown from database
- on hover, when displaying the "+" button, the line jumps up and down a bit
- make the + button more prominent (change bg color? especially in dark mode)
- editor should come before the annotations instead of below
- fix "verzenden" button styling (maybe align button to the right?)
- add a cancel button
- reduce margin around button
- fix styling of comment directly after submitting
- bring annotation more in line with linting comments (green left border?)
- add author and date to the comment
- remove the delete button from the "default" view, but add it to the "edit" view
- remove the edit button from the "edit" view
- use a simple pencil icon instead of a real button for the edit action

app/helpers/renderers/feedback_code_renderer.rb
```diff
@@ -1,15 +1,27 @@
...
   end
 
+  def add_submission(submission_id)
+    @submission_id = submission_id
+    @submission = Submission.find(@submission_id)
+    @code = @submission.code
+    @programming_language = @submission.exercise&.programming_language&.name
+    @annotations = @submission.annotations
+    self
+  end
```
charvp: You don't use a lot of these variables? Or they were already set? Anyway, just make the submission an optional parameter of the constructor.

app/helpers/renderers/feedback_table_renderer.rb
```diff
@@ -16,6 +16,8 @@ class << self
 
   def initialize(submission, user)
     result = submission.safe_result(user)
+    @submission_id = submission.id
+    @annotations = submission.annotations
```
charvp: Not used?

app/models/annotation.rb
```diff
@@ -0,0 +1,14 @@
...
+
+  belongs_to :submission
+  belongs_to :user
+
+  validates :user, presence: true
+  validates :annotation_text, presence: true, length: { minimum: 1, maximum: 2048 }
+  validates :line_nr, numericality: {
+    only_integer: true,
+    greater_than_or_equal_to: 0,
+    less_than_or_equal_to: ->(annotation) { annotation.submission.code.lines.count }
```
charvp: This reads the code file for this validation. Seems like this will have a performance impact we don't want.

app/policies/annotation_policy.rb
```diff
@@ -0,0 +1,31 @@
+class AnnotationPolicy < ApplicationPolicy
+  def create?
+    record.submission.user == record.user || course_admin?
```
charvp: We only want this to be zeus and maybe Peter for the initial version.

app/policies/annotation_policy.rb
```diff
@@ -0,0 +1,31 @@
...
+  end
+
+  def update?
+    record.user == user || course_admin?
```
charvp: Don't see why course admins should be able to edit annotations of other users? (Zeus should be able to, though.)

app/policies/annotation_policy.rb
```diff
@@ -0,0 +1,31 @@
...
+  end
+
+  def destroy?
+    record.user == user || course_admin?
```
charvp: Same as update

app/policies/annotation_policy.rb
```diff
@@ -0,0 +1,31 @@
...
+  def permitted_attributes
+    attribs = [:annotation_text]
+    attribs += [:line_nr] if record == Annotation || record.new_record?
+    attribs
+  end
+
+  # Record is a submission in this case
+  # TODO: Add owner of submission to this.
+  def show_comment_button?
+    user&.course_admin?(record.course)
```
charvp: Is this not (or should it not be) the same as create?

test/controllers/annotations_controller_test.rb
```diff
@@ -0,0 +1,77 @@
...
+
+class AnnotationControllerTest < ActionDispatch::IntegrationTest
+  def setup
+    @submission = create :correct_submission
+    @zeus = create(:zeus)
+    sign_in @zeus
+  end
+
+  test 'can create annotation' do
+    post "/submissions/#{@submission.id}/annotations", params: {
```
charvp: Use url helpers (also for the other tests).

app/assets/javascripts/code_listing/code_listing.ts
```diff
@@ -161,4 +191,191 @@ export class CodeListing {
...
+            return;
+        }
+
+        const annotationSubmissionDiv: HTMLFormElement = this.createAnnotationSubmissionDiv(lineId);
+        const annotationCell: HTMLTableDataCellElement = tr.querySelector("td.annotation-cell");
+        annotationCell.append(annotationSubmissionDiv);
+    }
+
+    createAnnotationSubmissionDiv(lineId: string, annotation?: UserAnnotation): HTMLFormElement {
+        const annotationSubmissionDiv: HTMLFormElement = document.createElement("form");
```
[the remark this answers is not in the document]
bmesuere: unescaped user content should be added with setText

PR #1487, opened 2019-11-13
allow words to be broken in feedbacktable if there is no other option

### LINKED ISSUE #1486: doctests with long input arguments should trigger an automatic line break
When the input argument of a (Python) function is a string with many characters, this string is displayed outside of the CSS container. I think line breaks should automatically be triggered in this case. Example:
Screenshot 2019-11-13 at 10 19 53

charvp:
According to [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap) words will only be broken if there are no other break targets, which seems like a fine compromise to me.

image

Closes #1486.

### DIFF
--- a/app/assets/stylesheets/models/submissions.css.less
+++ b/app/assets/stylesheets/models/submissions.css.less
@@ -68,6 +68,7 @@
     font-family: @font-family-monospace;
     font-size: 13px;
     white-space: pre-wrap;
+    overflow-wrap: break-word;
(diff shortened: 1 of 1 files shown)


pdawyndt:
just a question; won't wrapping long lines be confused with truely multiline strings, e.g. when the description is not a statement, but. e.g. the content of stdin?

PR #1489, opened 2019-11-14
exercise information page

### LINKED ISSUE #1386: exercise info page
Each exercise should get an info page, visible for course admins. This page could contain the following things:
- who is responsible for the exercise (i.e. who to contact in case of questions or errors)
- in what language the exercise is available (also store this in the database)
- a 'description' sourced from the repo. This would allow exercise makers to add hints for other teachers
- an optional example solution sourced from the repo
- the config that will be used for judging (e.g. time and memory limits)
- where the exercise is used? (maybe this should only be visible to the exercise owner)

rien:
This pull request adds a page for each exercise where teachers can see exercise metadata.

- [x] Add tests
- [x] Add documentation (PR: https://github.com/dodona-edu/dodona-edu.github.io/pull/32)

Closes #1386.


rien:
A few suggestions for the navbar link to this page:

`information`
image
`information-outline`
image
`information-variant`
image
`help`
image

Preferences?

charvp:
First one looks best (and most consistent) to me. I would place it to the right of the edit button though.

rien:
I have tried to make a quick and dirty mockup of this page:
image
I'm not completely sure of the separate cards for different 'types' of information. Should I put it in in a single card together with the 'about'  information?

pdawyndt:
Show exercise labels on this page

rien:
image
 I have made a version with less cards. The idea would be t make the sample solution and courses cards collapsible.

pdawyndt:
Looks better. Labels directly below description (in analogy to a list of keywords that comes after an abstract).

rien:
> Looks better. Labels directly below description (in analogy to a list of keywords that comes after an abstract).

Great idea, I was looking for a proper place for them.

> Take into account that list of course might become quite long. Maybe put that one below the sample solution?

I was thinking of making the courses and sample solutions (all the files in the `solutions/` folder) cards collapsible.

rien:
image
This is starting to look a lot cleaner now.

pdawyndt:
- what about using ACE editor with multiple tabs as a way to show multiple solutions?
- should we include all files from the solutions folder, or only files that match some file name pattern?

rien:
> * what about using ACE editor with multiple tabs as a way to show multiple solutions?

Yes, I will look into that.

> * should we include all files from the solutions folder, or only files that match some file name pattern?

I don't want to enforce a single name, because for Java the filename has to match the class declared within that file. We have the file extension, so I can use all files matching `solutions/*.#{extension}`.

rien:
Question: do we want do add 'about' information as a field in the `config.json` or as a file (or even multiple files with localization)?

Since it seem useful to support markdown, I'm inclined to go for the file approach.

pdawyndt:
Given the fact that the about might become quite lengthy, I would also consider to put it in a separate file.

rien:
I am experimenting with the layout of the author and creation information. Which layout do you prefer?

As subtitle:
image

As a sentence in muted text:
image

As an unordered list with icons:
image

rien:
I have picked option 3. The time information could be useful and the subtitle would become too crowded.

I now have the following layout:

- On a large screen
image
- On a medium screen
image
- On a small screen
image

Suggestions, improvements and formic fornication is welcome

bmesuere:
It would be cool if we could show what the source of a certain setting was (e.g. tooltip showing "set by _filename_")
Not sure about the use of a badge to display the count of something that is visible. Maybe change the title to "Usage" and add an info-style subheader "3 courses are using this exercise".

rien:
Do we want the 'new' design for "courses using this exercise"?
image
 Or should I move the old one from the exercise information page:
image

rien:
@charvp:
>* The unselected tabs in the examples aren't very noticeable, especially when there is only one file.

I'm not sure how I can fix this, maybe a small hint?
Like this: 
image

>* Do we have an icon for every programming language currently supported by Dodona? What happens if the icon doesn't exist?
There are icons for C, C++, **C#**, Fortran, Go, **Haskell**, **Javascript**, Lua, PHP, **Python**, **R**, **Ruby**, Swift, and TypeScript.

So we are missing an icon for bash, sh, prolog and text. I'll make a helper for this icon, which defaults to the name of the language. We could also contribute the missing logo's to the Material Design Icons ourselves.

charvp:
> @charvp:
> 
> > * The unselected tabs in the examples aren't very noticeable, especially when there is only one file.
> 
> I'm not sure how I can fix this, maybe a small hint?
> Like this:
> image

The hint could definitely help.

> 
> > * Do we have an icon for every programming language currently supported by Dodona? What happens if the icon doesn't exist?
> 
> There are icons for C, C++, **C#**, Fortran, Go, **Haskell**, **Javascript**, Lua, PHP, **Python**, **R**, Ruby, Swift, and TypeScript.
> 
> So we are missing an icon for bash, sh, prolog and text. I'll make a helper for this icon, which defaults to the name of the language. We could also contribute the missing logo's to the Material Design Icons ourselves.

I don't know if a helper is the right idea. We have fields in the programming language model for the different names in the different usages. Maybe this could be a new field?

rien:
Should there be a documentation entry on dodona-edu.github.io about this?

charvp:
I would mention it on the getting started page where users are shown how to add exercises to a series. When the documentation for creating exercises is added, the about document should be mentioned there.

rien:
A PR for the documentation is ready: https://github.com/dodona-edu/dodona-edu.github.io/pull/32.

rien:
TODO: add a limit to the included file size

app/views/exercises/_navbar_links.html.erb
```diff
@@ -9,12 +9,19 @@
...
                   if: can_view_submissions %>
+
   <%= navbar_link url: edit_exercise_scoped_path(exercise: @exercise, course: @course, series: @series),
                   title: t('.edit'),
                   icon: 'pencil',
                   if: policy(@exercise).edit? %>
+
+  <%= navbar_link url: info_exercise_scoped_path(exercise: @exercise, course: @course, series: @series),
+                  title: t('.info'),
+                  icon: 'information',
```
bmesuere:
```suggestion
                  icon: 'information-outline',
```

I would use the outline icon here because it's shown "inverted" on a colored background

app/views/exercises/info.html.erb
```diff
@@ -0,0 +1,222 @@
...
+        <% end %>
+        <div class="row">
+          <div class="col-md-4 col-md-push-8 col-sm-6  col-sm-push-6 col-xs-12">
+            <div class="row">
+              <div class="col-xs-12 repo-details">
+                <div class="row">
+                  <div class="col-xs-12">
+                    <p title="<%= t '.contact' %>">
+                    <i class="mdi mdi-account-box mdi-24"></i>
+                    <%= mail_to(@config['contact'] || @repository.first_admin&.pretty_email) %>
```
bmesuere: probably because of my dev-setup, but this was shown as a blank string. Also, some of the icons are clipped (on Chrome), so maybe some extra line height solves this.
image

rien: There contact info will be a blank string if there no contact defined for the exercise and there is no repository admin. In production this would normally not be the case.

PR #1510, opened 2019-11-27
Fix csp tutor, re-add X-Frame-Options

rien:
This pull request

- [ ] Tests were added

Closes # .

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -54,6 +54,10 @@ class ApplicationController < ActionController::Base
 
   protected
 
+  def allow_iframe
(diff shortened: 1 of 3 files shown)


charvp:
Don't we want to disable report-only again? Also, can you squash these commits, that makes it easier to create the hotfix.

charvp:
Also, shouldn't the `default_url` be added as frame_src for `submissions#show`?

rien:
> Also, shouldn't the `default_url` be added as frame_src for `submissions#show`?

`frame_src` should fallback to the `default_src` directive, which is `'self'` in our case.
That's why it is broken on the exercise page, because we replace `frame_src` with medusa, instead of adding medusa to the `frame_src` list.

PR #1519, opened 2019-11-28
Various code tab style fixes

### LINKED ISSUE #1516: preserve newlines in code annotations
When there are newlines in the code annotations:

```ruby
{
    "text"=>"Use lambda-case. Why not: \\case\n    (Message s) -> putStrLn s >>= const loop\n    (Stop sig) -> putStrLn \"logger: stop\" >>= const (putMVar sig ())",
    "type"=>"info",
    "row"=>20,
    "rows"=>22,
    "column"=>30,
    "columns"=>84
}
```

this results in correct newlines in the HTML code:

```html
<div class="annotation info" id="annotation-id-2">Use lambda-case. Why not: \case
    (Message s) -&gt; putStrLn s &gt;&gt;= const loop
    (Stop sig) -&gt; putStrLn "logger: stop" &gt;&gt;= const (putMVar sig ())</div>
```

but since this is a `div`, these are of course collapsed, resulting in:

no-newlines

MrKickkiller:
This pull request fixes a couple issues laid out in #1516 

image

Closes #1516  .

### DIFF
--- a/app/assets/javascripts/code_listing.ts
+++ b/app/assets/javascripts/code_listing.ts
@@ -60,7 +60,7 @@ export class CodeListing {
         annotationCell.setAttribute("class", "annotation");
         annotationCell.setAttribute("id", `annotation-id-${message.id}`);
 
-        const textNode: Text = document.createTextNode(message.text.split("\n").filter(s => s.match("^--*$")).join("\n"));
+        const textNode: Text = document.createTextNode(message.text.split("\n").filter(s => ! s.match("^--*$")).join("\n"));
         annotationCell.classList.add(message.type);
         annotationCell.appendChild(textNode);
 
--- a/app/assets/javascripts/code_listing.ts
+++ b/app/assets/javascripts/code_listing.ts
@@ -60,7 +60,7 @@ export class CodeListing {
         annotationCell.setAttribute("class", "annotation");
         annotationCell.setAttribute("id", `annotation-id-${message.id}`);
 
-        const textNode: Text = document.createTextNode(message.text.replace(/\n?--*\n?$/, ""));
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/code_listing.ts
```diff
@@ -45,6 +45,10 @@ export class CodeListing {
     }
 
     private createAnnotation(message: Message, tableIndex: number, rougeRow: number): HTMLTableRowElement {
+        if (message.text.match("^--*$")) {
```
charvp: Are the messages with dashes always single-line? If there is other content we don't just want to ignore it.

MrKickkiller: I understood @ninewise 's comment as if the rstripping of dashes might break some other judges output. 

As such, I think that simply ignoring a message that is only dashes is currently the best approach to take. 

image

As we discussed a couple of weeks back: The dashes should probably be handled in the judges that produce them.

charvp: Your second example should be shown as `text met dashes op het einde`. If the javascript regex engine does not handle that correctly you can split on newlines to make sure you get the correct behaviour.

MrKickkiller: image

for following inputs

image

charvp: The one without a newline should not be edited. Only if a line contains nothing but newlines should it be stripped.

MrKickkiller: So `\n--*\n?$`, as I assume you mean a line full of dashes instead of a line of newlines?

charvp: Indeed, I meant dashes. The simplest way to do this with code that is also readable is to split text on newlines. Then delete all elements that consist of nothing but newlines. Then join the array again.

PR #1520, opened 2019-11-28
Fix selection in Microsoft Edge including the tooltips

### LINKED ISSUE #1517: Inline code tips are included when copy/pasting in Edge
With the recent upgrade of Dodona on dodona.ugent.be the code tips are now inlined with the code.

However, Microsoft Edge has a bug that it also copies and pastes these tips when copy/pasting the code.

On Firefox, extra newlines are added on the location of the code tips.

Can you implement the necessary workarounds to prevent this behavior? (cfr. https://stackoverflow.com/questions/47250505/text-of-element-with-css-user-select-none-gets-copied-if-it-is-located-between)

MrKickkiller:
**Warning: currently untested** on edge as I use Linux as my main OS

This pull request should disallow selection in Microsoft Edge according to the given solution link

Closes #1517 .

### DIFF
--- a/app/assets/javascripts/code_listing.ts
+++ b/app/assets/javascripts/code_listing.ts
@@ -61,6 +61,11 @@ export class CodeListing {
         annotationCell.appendChild(textNode);
 
         annotationTD.appendChild(annotationCell);
+
+        const edgeCopyBlocker = document.createElement("div");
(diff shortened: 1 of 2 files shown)


MrKickkiller:
Looks to be good on Edge now

PR #1541, opened 2019-12-02
hide warnings and info messages by default when there are errors in code tab

### LINKED ISSUE #1540: reduce information overload in Code tab
I'm observing students during a test, and the following code review screen caught my eye:

image

This student has a total overload in messages to process.

Can you hide the tips (blue left bar), and maybe the warnings (orange  left bar), as long as one or more syntax errors (red left bar) are present? Imho, they have no value towards the student on such a moment.

MrKickkiller:
This pull request compresses the messages down on page load

Initial page load
image

After clicking show all
image

After clicking hide all
image

After clicking "Show only errors"
image

No messages
image

Only a simple warning/info message
image

Closes #1540 .


MrKickkiller:
@bmesuere 
image

Is the locale not being set on page load?

rien:
> @bmesuere
> image
> 
> Is the locale not being set on page load?

The locale is set here:
https://github.com/dodona-edu/dodona/blob/81242168921df9b55700105490f6dd6092d9903f/app/views/layouts/application.html.erb#L136

pdawyndt:
Seems logical to sort messages on a line by severity: first errors, then warnings, then info.

MrKickkiller:
@rien Found why the string was always in English. The locale is being set at the end of the document while the `checkForErrorAndCompress` function is being run mid document, when the table from rouge has just finished loading.

app/assets/javascripts/code_listing.ts
```diff
@@ -52,8 +52,10 @@ export class CodeListing {
 
                 const showMoreLink: HTMLSpanElement = document.createElement("span");
                 showMoreLink.setAttribute("class", "displayMore");
-                const showMoreLinkText: Text = document.createTextNode(I18n.t("js.compressed_warning_info"));
-                showMoreLink.append(showMoreLinkText);
+                window.addEventListener("DOMContentLoaded", () => {
```
rien: Can you confirm this event is always fired when you need it to be? For example: on the exercise page we dynamically load the feedback table by requesting a `.js` file and I don't think that this event will be triggered.

rien: Alternatively we could move the `I18n.locale = "<%= I18n.locale %>";` line to the top of the document. It's just one line of JavaScript, so its impact on the load time would be negligible.

MrKickkiller: Moving the I18N locale set to the top would indeed be better as we then eliminate any further event-juggling

charvp: We usually wrap all JS in the generated HTML in jquery's `$()`. This makes sure the JS is only executed once the DOM is fully loaded. Is there a reason we can't do this here. Seems like a cleaner (and more consistent) solution to me.

rien: I think in the long term we want to get rid of JQuery. Also, does `$()` trigger with remote requests?

charvp: Yes, it does. And while we haven't gotten rid of jQuery I don't think this is a problem, since we use it everywhere in the `erb` files, so switching it out will be a mass action anyway once we get rid of it.

app/assets/javascripts/code_listing.ts
```diff
@@ -21,8 +21,15 @@ export class CodeListing {
         }
     }
 
+    static readonly ordering = ["error", "warning", "info"];
```
charvp: Don't we want to UPPER_SNAKE_CASE class constants?

app/assets/javascripts/code_listing.ts
```diff
@@ -19,18 +19,71 @@ export class CodeListing {
         if (this.table === null) {
             console.error("The code listing could not be found");
         }
+        const hideAllButton = document.querySelector("#hide_all_annotations");
```
charvp: The naming of these variables seems off to me? AFAICT the buttons aren't use to hide *all* annotations.

app/assets/javascripts/code_listing.ts
```diff
@@ -19,18 +19,71 @@ export class CodeListing {
...
+            });
+            showAllButton.addEventListener("click", () => {
+                this.decompressWarningsAndInfo();
+                hideAllButton.classList.remove("active");
+                showAllButton.classList.add("active");
+            });
+        }
+    }
+
+    static readonly ORDERING = ["error", "warning", "info"];
```
charvp: I would put this at the top of the class.

app/assets/javascripts/code_listing.ts
```diff
@@ -19,18 +19,71 @@ export class CodeListing {
...
+        for (const message of this.messages) {
             message.id = idOffset + 1;
             idOffset += 1;
             // Linter counts from 0, rouge counts from 1
             const correspondingLine: HTMLTableRowElement = this.table.querySelector(`#line-${message.row + 1}`);
             this.createAnnotation(message, correspondingLine.rowIndex + 1, message.row + 1);
-            this.messages.push(message);
         }
+
+        this.checkForErrorAndCompress();
```
charvp: If at some point in the future we allow annotations to be added at a later time, this would be wrong I think. The behaviour should depend on the selection of the user in the buttons.

config/locales/js/en.yml
```diff
@@ -81,3 +81,4 @@ en:
         fri: Fr
         sat: Sa
         sun: Su
+    compressed_warning_info: "One or more warnings and information messages are not shown. Click to show more"
```
charvp: Are these messages still used?

MrKickkiller: No, they should indeed be removed. Coming right up

MrKickkiller:
Page load => Click on 'Code'
image

Page load => Click on 'Code' => Click on "Toon suggesties"
image

Only warning and info
image

pdawyndt:
Why no show/hide when there are no errors? Seems like a functionality that we want as soon as there is at least one annotaton (irrespective of its type).

pdawyndt:
The design of the show/hide buttons is different in the different tabs

- correctness tabs use buttons with icons for showing/hiding correct test cases
- code tab uses buttons with text for showing/hiding annotations

pdawyndt:
Do we want to display a "show annotations"-button when all annotations are shown? Do we want to display a "hide annotations"-button when all annotations are hidden?

MrKickkiller:
@pdawyndt 

> Do we want to display a "show annotations"-button when all annotations are shown? Do we want to display a "hide annotations"-button when all annotations are hidden?

This behaviour seems most inline with the "below or side by side" buttons from the test cases. Which is also a toggle between 2 modes.

> The design of the show/hide buttons is different in the different 
correctness tabs use buttons with icons for showing/hiding correct test cases
code tab uses buttons with text for showing/hiding annotations

Replacing text with an icon would be simple to do. 
mdi-arrow-collapse-all for hiding and mdi-arrow-expand-all for showing perhaps?

> Why no show/hide when there are no errors? Seems like a functionality that we want as soon as there is at least one annotaton (irrespective of its type).

The original issue mentioned the following:
> Can you hide the tips (blue left bar), and maybe the warnings (orange left bar), as long as one or more syntax errors (red left bar) are present? Imho, they have no value towards the student on such a moment.

As such the code currently looks if there is any error message and then hides the 2 other message types. The Show & Hide button therefor is useless if there is no error present on the page. The user would be able to click on a button that does nothing.

pdawyndt:
What I was refering to is that the buttons should be neutral with respect to the types of the annotations. Now it seems that there are no buttons when there are no error-type annotations, but there are warning-type annotations or info-type annotations (see your last screenshot). In that case I would add the buttons in order to be able to show/hide the warning- and info-type annotations.

Of course, if there are not annotations at all, there should also be no buttons.

MrKickkiller:
@bmesuere 
image

> we also want to display a message at the top of the table mentioning this with a link to show all annotations

Would you rather have it at location 1 or 2?

beardhatcode:
My intuition would be to put it at spot 2, just like the horizontal vs vertical diff selectors

MrKickkiller:
image

beardhatcode:
It might be useful to show a little red, orange and/or blue dot in the sidebar (next to the line number) if there was a hidden message on that spot. Upon clicking the dot, the warnings should be shown again.

You could also choose to colorize the icon of the icon in the switcher if there are hidden messages of that kind. (Or also put like a supper tiny dot in there)

app/assets/javascripts/code_listing/code_listing.ts
```diff
@@ -0,0 +1,134 @@
+import { Message, MessageData } from "../../../../app/assets/javascripts/code_listing/message";
```
bmesuere: this is a strange import path...

app/assets/javascripts/code_listing/code_listing.ts
```diff
@@ -0,0 +1,134 @@
...
+
+        this.table.addEventListener("copy", function (e) {
+            e.clipboardData.setData("text/plain", window.dodona.codeListing.getSelectededCode());
+            e.preventDefault();
+        });
+
+        this.initButtonsForView();
+    }
+
+    private initButtonsForView(): void {
```
bmesuere: rename to `initAnnotationToggleButtons`

app/assets/javascripts/code_listing/code_listing.ts
```diff
@@ -0,0 +1,134 @@
...
+        const hideAllButton: HTMLButtonElement = document.querySelector("#hide_all_annotations");
+        const showOnlyErrorButton: HTMLButtonElement = document.querySelector("#show_only_errors");
+        const showAllButton: HTMLButtonElement = document.querySelector("#show_all_annotations");
+        const messagesWereHidden: HTMLSpanElement = document.querySelector("#messages-were-hidden");
+
+        const showAllListener = (): void => {
+            this.showAllAnnotations();
+            if (messagesWereHidden) {
+                messagesWereHidden.remove();
+            }
```
bmesuere: I think our typescript version is recent enough to replace this by `messagesWereHidden?.remove()`

app/assets/javascripts/code_listing/code_listing.ts
```diff
@@ -0,0 +1,134 @@
...
+            hideAllButton.addEventListener("click", this.hideAllAnnotations.bind(this));
+        }
+
+        if (showOnlyErrorButton && hideAllButton && showAllButton) {
+            showOnlyErrorButton.addEventListener("click", this.compressMessages.bind(this));
+        }
+
+        if (messagesWereHidden && showAllButton) {
+            messagesWereHidden.addEventListener("click", () => showAllButton.click());
+        }
```
bmesuere: what is the reason for wrapping these in if statements?

app/assets/javascripts/code_listing/code_listing.ts
```diff
@@ -0,0 +1,134 @@
...
+
+    addAnnotation(message: MessageData): void {
+        this.messages.push(new Message(this.messages.length, message, this.table, this));
+    }
+
+    compressMessages(): void {
+        this.showAllAnnotations();
+
+        const errors = this.messages.filter(m => m.type === "error");
+        if (errors.length != 0) {
```
bmesuere:
```suggestion
        if (errors.length !== 0) {
```

app/assets/javascripts/code_listing/code_listing.ts
```diff
@@ -0,0 +1,134 @@
...
+
+    highlightLine(lineNr: number, scrollToLine = false): void {
+        const toMarkAnnotationRow = this.table.querySelector(`tr.lineno#line-${lineNr}`);
+        toMarkAnnotationRow.classList.add(this.markingClass);
+        if (scrollToLine) {
+            toMarkAnnotationRow.scrollIntoView({ block: "center" });
+        }
+    }
+
+    getCode(): string {
```
bmesuere: Is there a reason why we still do this as opposed to simply passing the code upon the construction of this object and returning that?

app/assets/javascripts/code_listing/message.ts
```diff
@@ -0,0 +1,119 @@
...
+        this.dot = document.createElement("span");
+        this.dot.setAttribute("class", `dot dot-${this.type}`);
+        codeGutter.prepend(this.dot);
+    }
+
+    removeDot(): void {
+        if (this.dot != undefined) {
+            this.dot.remove();
+            this.dot = null;
+        }
```
bmesuere: use `this.dot?.remove()` instead of wrapping with if

app/assets/javascripts/code_listing/message.ts
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    order(): number {
+        return ORDERING.findIndex(p => p == this.type);
```
bmesuere:
```suggestion
        return ORDERING.findIndex(p => p === this.type);
```

app/assets/javascripts/code_listing/message.ts
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    removeDot(): void {
+        if (this.dot != undefined) {
+            this.dot.remove();
+            this.dot = null;
+        }
+    }
+
+    order(): number {
```
bmesuere: shouldn't this better be called `getPosition`?

app/assets/javascripts/code_listing/message.ts
```diff
@@ -0,0 +1,119 @@
...
+        return annotationRow;
+    }
+
+    addDot(): void {
+        if (this.dot) {
+            return;
+        }
+
+        const ownOrder = this.order();
+        const others: Message[] = this.codeListing.messages.filter(p => this.line == p.line && p.dot != undefined);
```
bmesuere: add a utility function that returns all annotations for a certain line to codeListing

app/assets/javascripts/code_listing/message.ts
```diff
@@ -0,0 +1,119 @@
...
+        others.forEach(dotDisplayed => {
```
bmesuere: this logic seems to be flawed (see general comment)

app/assets/stylesheets/components/code_listing.css.less
```diff
@@ -1,91 +1,149 @@
...
-    }
+  .feedback-table-options {
+    margin-bottom: 0.5em !important;
   }
 
-  .lineno:last-child {
-    .rouge-code, .rouge-gutter {
-      padding-bottom: 5px;
-    }
+  #messages-were-hidden {
```
bmesuere: Do we really need this specific styling for a simple message?

app/assets/stylesheets/components/code_listing.css.less
```diff
@@ -1,91 +1,149 @@
...
-    white-space: pre-wrap;
-  }
+.code-table {
 
-  .lineno:first-child {
-    .rouge-code, .rouge-gutter {
-      padding-top: 5px;
-    }
+  .feedback-table-options {
+    margin-bottom: 0.5em !important;
```
bmesuere: !important should be a last resort measure. Have you tried other options without it?

app/assets/stylesheets/components/code_listing.css.less
```diff
@@ -1,91 +1,149 @@
...
+      word-break: break-word;
+    }
+
+    .rouge-gutter {
+      min-width: 30px;
+      text-align: right;
+      vertical-align: top;
+      padding-right: 5px;
+
+      display: flex;
```
bmesuere: is there a reason to use a flex layout here?

app/assets/javascripts/code_listing/code_listing.ts
MrKickkiller: https://github.com/dodona-edu/dodona/pull/1541/files#diff-9d3a545f5d6451a196966d617750cef8R23

If there are no messages to be displayed, the buttons are not inserted into the DOM

app/assets/javascripts/code_listing/message.ts
MrKickkiller: In my opinion, the name `getPosition` would indicate the messages position in a list of messages. While `order` simply gives me a numerical value that tells us what elements are of greater/lower value.

app/assets/javascripts/code_listing/code_listing.ts
bmesuere: In that case I would always add the buttons to the html (maybe hidden?) and make them visible and/or initialize them when the first message gets added?

app/assets/javascripts/code_listing/code_listing.ts
MrKickkiller: The Jest runtime does not properly find the other files when they are defined like `code_listing/message`. It looks relative to itself, not relative to the file it has just fetched

app/assets/stylesheets/components/code_listing.css.less
MrKickkiller: The dots are inserted at the same level as the line number. As they need to be `display: inline-block` or `display:block` to display, they would push the line number to a new line (appearing above the line number.

app/assets/javascripts/code_listing/code_listing.ts
```diff
@@ -0,0 +1,168 @@
...
+
+    set code(codeString: string) {
+        this._code = codeString;
+    }
+
+    get code(): string {
+        return this._code;
+    }
+
+    private getSelectededCode(): string {
```
charvp:
```suggestion
    private getSelectedCode(): string {
```

config/locales/js/nl.yml
```diff
@@ -81,4 +81,9 @@ nl:
         fri: Vr
         sat: Za
         sun: Zo
+    message:
+      hidden: Een bericht werd verborgen
+      were_hidden:
+        single: Enkel errors worden getoond, toon {0} ander bericht
```
bmesuere: split into 2 parts and make only the last part a link as suggested in the previous review

app/assets/stylesheets/components/code_listing.css.less
MrKickkiller:
```html
<td class="rouge-gutter gl">
    <span class="dot" title="Een annotatie werd verborgen"></span>
    <pre>1</pre>
</td>

```

The dot class needs to have `display: block` or `display: inline-block` to even be displayed. Without making the rouge-gutter a row-based flex layout, the dots would appear above the line-number.

image

PR #1572, opened 2019-12-09
Make the export action asynchronous and add notifications

### LINKED ISSUE #1578: Series export with `last_submission = true` still downloads all submissions
Reported by @thepieterdc.

### LINKED ISSUE #1633: Exercise order in CSV file created during export is not stable
The API used by Indianio no longer preserves the original order of the exercises in a series.

charvp:
Closes #1578.
Closes #1633.


charvp:
Export overview:
image
View when there is an unfinished export:
image
Extra link in user menu (only shown if user has exports):
image
Bell with icon if user has unread notifications:
image
Notification listing:
image
Bell when user has notifications, but no unread + menu
image

Clicking the notification takes the user to the exports index (and sets the notification to read).

app/assets/javascripts/notification.ts
```diff
@@ -1,51 +1,56 @@
...
-                this.hide();
-            }, Notification.hideDelay);
-        }
-    }
-
-    private show(): void {
-        document.querySelector(Notification.notificationsContainer).prepend(this.notification);
-        window.requestAnimationFrame(() => {
-            this.notification.classList.remove("notification-show");
+        this.element.querySelector(".notification-link").addEventListener("click", event => {
```
bmesuere: split these two actions into separate functions (eg. markAsRead)

app/helpers/export_helper.rb
```diff
@@ -0,0 +1,270 @@
...
+      }
+      @names = {
+        courses: {},
+        exercises: {},
+        series: {},
+        users: {}
+      }
+    end
+
+    # Exporting a zip from Indianio: make sure to return specific output for this request
```
bmesuere:
```suggestion
    # Exporting a zip for Indianio: make sure to return specific output for this request
```

app/policies/notification_policy.rb
```diff
@@ -0,0 +1,23 @@
+class NotificationPolicy < ApplicationPolicy
+  class Scope < Scope
+    def resolve
+      scope.where(user: user)
```
bmesuere: is there a reason to do this using a scope instead of simply using the relation in the model?

app/views/courses/show.html.erb
```diff
@@ -78,15 +78,15 @@
               <% end %>
               <% if current_user&.course_admin?(@course) %>
                 <li>
-                  <%= link_to t(".download_submissions"), export_course_path(@course) %>
+                  <%= link_to t(".download_submissions"), for_course_exports_path(@course) %>
```
bmesuere: the `for` in these names seem a bit strange

app/views/notifications/_notifications_table.html.erb
```diff
@@ -0,0 +1,35 @@
+<table class="table table-index table-resource">
+  <tbody>
+  <% notifications.each do |n| %>
+    <tr class="notification" data-id="<%= n.id %>">
+      <td>
+        <a class="notification-link" href="<%= notifiable_url n %>"><%= t n.message %></a>
+        <div class="text-muted">
+          <small>
+            <span><%= "#{time_ago_in_words n.created_at} #{t 'series.deadline.ago'}" %></span>
```
bmesuere: make separate or renamed translation key for this

app/views/notifications/destroy.js.erb
```diff
@@ -0,0 +1 @@
+$(".notification[data-id=\"<%= @notification.id %>\"]").remove();
```
bmesuere: when is this called?

charvp:
> minor remark, by renaming everything to export, it becomes less clear what is what when reading the code. eg, the name scoresheet directly implies an overview for course admins.

I'm not sure what you mean by this. As far as I can tell, the word download is used for things related to the scoresheet functionality.

app/assets/javascripts/notification.ts
```diff
@@ -1,51 +1,56 @@
...
  */
 export class Notification {
-    static readonly hideDelay = 3000;
-    static readonly removeDelay = 1000;
-    static readonly notificationsContainer = ".notifications";
+    readonly element: Element;
+    read: boolean;
 
-    readonly notification: Element;
+    constructor(readonly id: number, readonly url: string, readonly _read: boolean) {
```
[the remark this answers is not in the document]
charvp: Not sure what you mean. AFAICT the arguments of the constructor are not automatically made fields of the class, which is what I think you mean?

app/policies/notification_policy.rb
charvp: Not really. I guess we will never have notifications the user is not allowed to see, so I guess a policy scope is not the right abstraction. Will change.

app/assets/javascripts/notification.ts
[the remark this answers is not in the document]
charvp: Oh wow, this behaviour is confusing. If I add a modifier it is automatically added as a field to the class, without modifiers it isn't. I can see why you've started to dislike the shorthand.

app/views/notifications/destroy.js.erb
charvp: When a notification is deleted by clicking the x in the notification view.

app/policies/series_policy.rb
```diff
@@ -77,10 +77,6 @@ def scoresheet?
     course_admin?
   end
 
-  def scoresheet_download?
```
charvp: If this is what you meant with changing everything to export, this isn't what it looks like. This was some dead code that I noticed was still around, so I removed it.

charvp:
> maybe "data export" is a better name than just "export"?

Tried this, but it looked clunky to me in most places.

> would it make sense to first sort notifications on read state and only then on time?

Also tried this, but the notifications jumping around between page loads made me confused, and I was the one who programmed it. Maybe this can be revisited when the notifications are more distinct?

config/locales/views/notifications/nl.yml
```diff
@@ -0,0 +1,9 @@
+nl:
+  notifications:
+    index:
+      created: "%{when} geleden"
+      view_all: "Alle notificaties bekijken"
+      title: Notificaties
+      message: Bericht
+      created_header: "Aangemaakt"
+      toggle_read_status: "Toggle gelezen"
```
rien: This sounds weird in dutch. Maybe split this in `Markeer als gelezen` and `Markeer als ongelezen`?

app/models/export.rb
```diff
@@ -0,0 +1,43 @@
...
+    ).bundle
+
+    archive.attach(
+      io: StringIO.new(bundle[:data]),
+      filename: bundle[:filename],
+      content_type: 'application/zip',
+      identify: false
+    )
+
+    delay(queue: 'cleaning', run_at: AUTOMATICALLY_DELETE_AFTER.from_now).destroy
```
rien: Will this survive for so long? An alternative would be to periodically (once a day) remove all exports older that `AUTOMATICALLY_DELETE_AFTER`.

app/assets/javascripts/toast.ts
```diff
@@ -0,0 +1,51 @@
...
+    hide(): void {
+        this.toast.classList.add("toast-hide");
+        setTimeout(() => {
+            this.toast.remove();
+        }, Toast.removeDelay);
+    }
+
+    private generateToastHTML(content: string, loading: boolean): Element {
+        const element = this.htmlToElement(
+            `<div class='toast toast-show'>${content}</div>`
```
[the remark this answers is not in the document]
charvp: This is not actually new code in this PR, so this should be fixed in a different one.

app/models/export.rb
charvp: These are just rows in the database, so it would be weird if they disappeared.

charvp:
> I haven't found a link to the notifications page to delete notifications, but I don't think it is needed. I would maybe add an option to 'delete all read notifications', because that's what a user would want to do most of the time I think.

The link to the notifications page appears when the user has more than five notifications.

> Just a suggestion: we could make the text of exports that have been downloaded (or 'seen') muted? So new exports would be extra visible.

We don't really have hooks on when an export is downloaded, and using the notification read status would require some significant changes that aren't worth it IMO. (I also have a gut feeling that it would make the code a lot uglier.)

rien:
> The link to the notifications page appears when the user has more than five notifications.

Then there is no way to hide the notification bell again if the user has less than 5 notifications.

Also, removing each notification separately seams a bit tedious if you just want to remove all of them. I would personally prefer a  'clear all notifications' button on the notifications page and/or in the small notification panel.

config/locales/views/exports/en.yml
```diff
@@ -1,14 +1,24 @@
...
+      title: "Exports"
+      status: "Status"
+      filename: "Filename"
+      started: "Started"
+      ready_for_download: "Your export is ready for download"
+      created: "%{when} ago"
+      export_started: "We are preparing your export and will notify you when it's ready"
+      automatic_deletion: Created exports will be available for 30 days
+      explanation_html: "Here you can find all exports you created. Exports are an archive of the submissions you made in a course or series, or even <a href=\"%{user_link}\">all your submissions</a>. Teachers can also download the submissions that their students made in the courses they administrate. You can create an export of a series or a course by using its menu."
+      no_exports_yet_html: "It seems like you haven't created any exports yet, or they were deleted automatically after 30 days. Exports are an archive of the submissions you made in a course or series, or even <a href=\"%{user_link}\">all your submissions</a>. Teachers can also download the submissions that their students made in the courses they administrate. You can create an export of a series or a course by using its menu."
```
bmesuere: I would still display the explanation_html message and a shorter message instead of the table.
```suggestion
      no_exports_yet_html: "It seems like you haven't created any recent data archives. Get started and<a href=\"%{user_link}\">export your submissions</a>."
```

app/views/notifications/index.html.erb
```diff
@@ -0,0 +1,17 @@
+<div class="row">
+  <div class="col-xs-12">
+    <div class="card">
+      <div class="card-title card-title-colored">
+        <h2 class="card-title-text"><%= t('.title') %></h2>
+        <div class="card-title-fab">
+          <%= render 'application/fab_link', url: destroy_all_notifications_path, icon: 'delete', method: :delete, title: t('.destroy_all') %>
```
rien: The 'delete_all' button doesn't seem to be properly aligned:
fab
This is in both firefox and chromium the case.

app/assets/javascripts/notification.ts
```diff
@@ -1,51 +1,85 @@
...
+            this.toggleRead();
+            event.stopPropagation();
+        });
 
-        if (this.autoHide) {
-            setTimeout(() => {
-                this.hide();
-            }, Notification.hideDelay);
+        // Delete button isn't shown on small view in navbar
+        if (this.element.querySelector(".delete-button") != null) {
```
bmesuere: Any reason not to use `!==`?

app/assets/stylesheets/components/notification.css.less
```diff
@@ -1,37 +1,11 @@
...
-  opacity: 0;
-}
-
-.notification .spinner {
-  width: 20px;
-  height: 20px;
-  float: right;
-  margin-left: 10px;
-  background-color: @notification-spinner-bg;
+.notification {
```
bmesuere: I'm probably missing something, but why is this here and the rest in the navbar file?

app/views/layouts/application.html.erb
```diff
@@ -17,8 +17,8 @@
   <meta name="msapplication-TileColor" content="#2196F3"/>
   <meta name="msapplication-TileImage" content="/icon.png"/>
   <link rel="apple-touch-icon-precomposed" href="/icon.png"/>
-  <link rel="shortcut icon" sizes="192x192" href="/icon.png">
-  <link rel="shortcut icon" href="/favicon.ico"/>
+  <link rel="shortcut icon" sizes="192x192" href="<%= current_user&.notifications&.where(read: false)&.any? ? "/icon-not.png" : "/icon.png" %>">
+  <link rel="shortcut icon" href="<%= current_user&.notifications&.where(read: false)&.any? ? "/favicon-not.ico" : "/favicon.ico" %>">
```
bmesuere: In a future version, it will probably make sense to only set this using JavaScript (and add background polling for new notifications)

app/assets/stylesheets/components/notification.css.less
charvp: This CSS applies to the notification index page as well. The CSS in the navbar file only applies to the notification drop-down in the navbar.

app/assets/javascripts/notification.ts
charvp: Nope, probably a typo. I will change it before merging.

app/views/layouts/application.html.erb
charvp: Agreed.

PR #1574, opened 2019-12-09
add line numbers to external text file

### LINKED ISSUE #1556: display of extra text-test-files does not contain trailing newline
When the content of a data file is viewed, trailing newlines are omitted (see screenshot). This causes confusion.

See also: https://dodona.ugent.be/nl/submissions/4456002/

Selection_102

toonijn:
This pull request adds line numbers to external files in a pythia-submission.

Selection_104
Selection_105
Selection_106
Selection_107

Closes #1556  .

### DIFF
--- a/app/assets/javascripts/pythia_submission.js
+++ b/app/assets/javascripts/pythia_submission.js
@@ -80,11 +80,9 @@ function initPythiaSubmissionShow(submissionCode) {
         $.get(path, function (data) {
             let lines = data.split("\n");
             const maxLines = 99;
-            let omitted = false;
             if (lines.length > maxLines) {
                 lines = lines.slice(0, maxLines);
                 lines.push("...");
-                omitted = true;
             }
 
             const table = document.createElement("table");
@@ -94,7 +92,7 @@ function initPythiaSubmissionShow(submissionCode) {
 
                 const number = document.createElement("td");
                 number.className = "line-nr";
-                number.textContent = (i === maxLines && omitted) ? "": i + 1;
+                number.textContent = (i === maxLines) ? "": i + 1;
                 tr.appendChild(number);
 
                 const line = document.createElement("td");
--- a/app/assets/javascripts/pythia_submission.js
+++ b/app/assets/javascripts/pythia_submission.js
@@ -87,11 +87,26 @@ function initPythiaSubmissionShow(submissionCode) {
                 omitted = true;
             }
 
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/pythia_submission.js
```diff
@@ -78,13 +78,20 @@ function initPythiaSubmissionShow(submissionCode) {
...
-                data = lines.slice(0, maxLines).join("\n") + "\n...";
+                lines = lines.slice(0, maxLines);
+                lines.push("...");
+                omitted = true;
             }
-            $("#file-" + random).html(data);
+
+            $("#file-" + random).html(`<table class='external-file'>${
+                lines.map((l, i) => `<tr><td class="line-nr">${
+                    i == maxLines && omitted ? "": i+1
```
bmesuere: - use `===` instead of `==`
- Do we need the `omitted` variable? Isn't the line equalling the value of `maxLines` always the one with `...`?

app/assets/javascripts/pythia_submission.js
```diff
@@ -78,13 +78,20 @@ function initPythiaSubmissionShow(submissionCode) {
...
+                }</td><td class="line">${l}</td></tr>`).join("")
```
bmesuere: I think `l` needs to be escaped, @rien?

app/assets/stylesheets/components/tutor.css.less
```diff
@@ -21,6 +21,59 @@
...
+    overflow: auto;
+    font-size: 13px;
+    margin: 0;
+    padding: 0;
+    width: 100%;
+    table-layout: auto;
+    user-select: all;
+    -moz-user-select: all;
+    -webkit-user-select: all;
+    -ms-user-select: all;
```
bmesuere: our auto prefixer takes care of this.

app/assets/javascripts/pythia_submission.js
rien: I think so.
I'll try to break this...

rien: image

rien: Should be fixed by d2b6256.

app/assets/javascripts/pythia_submission.js
toonijn: `omitted` is not needed.

toonijn:
Reusing css is not that easy because the code-listing is using Rouge.

PR #1576, opened 2019-12-11
Set charset to utf-8 if sending plaintext media

rien:
This pull request fixes plaintext files containing UTF-8 not being correctly displayed by the browser.

- [X] Tests were added


rien:
My first fix, always sending `text/plain; charset=utf-8`, didn't break images in my browser somehow. So I think browsers do their own content-type sniffing anyway.

Possible problems could be fixed simply by changing the file extension, since that is used to determine the content-type.

PR #1597, opened 2019-12-17
Update gemfile to avoid yanked version of httparty

MrKickkiller:
This pull request skips version 0.17.2 of Httparty, since that has been removed from rubygems.org


charvp:
Already fixed by #1595, so I would rather just merge that.

PR #1598, opened 2019-12-18
Fix inaccessible private hidden exercise media

### LINKED ISSUE #1588: Media folder of private exercises not accessible
At the start of today's evaluation, we had to make the exercises public, otherwise the files from the media folder were not accessible (both images, text files or any other files). Private exercises were contained in a hidden exercise series, so they should have been accessible, but they didn't.

note: @rien observed that the media files were pointing to dodona instead of medusa

rien:
The `@course` was not set when checking whether a student could access a private exercise within a course. This PR fixes this and solves a few other problems:
- There were two routes defined for `/submisssions/:id/media/`, I removed the route that was overridden by the 'actual' route (the one going to `submissions#media`).
- Rails' ActiveRecord `include?` sometimes returns `false` even when a record is part of the relation. This is because two objects representing the same record are sometimes not equal to each other. I have fixed this by checking if their id's match.

- [X] Tests were added

Fixes #1588.

### DIFF
--- a/app/policies/course_policy.rb
+++ b/app/policies/course_policy.rb
@@ -123,6 +123,10 @@ class CoursePolicy < ApplicationPolicy
     statistics?
   end
 
+  def media?
+    show?
+  end
+
   def permitted_attributes
     # record is the Course class on create
     if course_admin? || (record == Course && user&.admin?)
--- a/app/assets/javascripts/exercise.js
+++ b/app/assets/javascripts/exercise.js
@@ -105,11 +105,10 @@ function initMathJax() {
     }
 }
 
-function initExerciseDescription(parentId, exercisePath, token) {
+function initExerciseDescription() {
     initLightboxes();
     centerImagesAndTables();
     initMathJax();
-    contextualizeMediaPaths(parentId, exercisePath, token);
 }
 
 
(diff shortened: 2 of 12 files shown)


rien:
This doesn't work because description URL's on medusa are not scoped.

rien:
Thoughts about media access control:
- If we add the token to the description path (`/exercises/:exercise_id/description/:token/`) then each media item will be accessible and authorized because there is a token present in the URL. Each media item is then served from Medusa without any problems.
- The exercise media paths `/exercise/:exercise_id/media/` served from Dodona will only be used to see feedback tables, meaning that a user should have access to at least one submission for that exercise. This is the case when:
  - The user has a submission for that exercise.
  - The exercise has submissions where the `course_id` is equal to one of the courses the user is administrating.
 - The user is an admin.

Is this correct? @charvp?

charvp:
Seems correct to me. Keep in mind that submissions can be viewed on their own page or on the exercise page. If they are viewed on the exercise page, it is known that the submitter is viewing them.

rien:
The exercise media paths are currently delegated to the exercise they belong to, so they have a slightly more permissible access control.

So in this PR the only thing we have to do is add a token to the description path and serve media items in this description directly from medusa relative to the description.

charvp:
> The exercise media paths are currently delegated to the exercise they belong to, so they have a slightly more permissible access control.

I'm not entirely sure what you mean by this? Or do you mean that **submission** media paths are delegated to their exercise?

rien:
> > The exercise media paths are currently delegated to the exercise they belong to, so they have a slightly more permissible access control.
> 
> I'm not entirely sure what you mean by this? Or do you mean that **submission** media paths are delegated to their exercise?

We currently check if a user can access a media item by checking if they can access the exercise. This is more permissive than it could be, but I think we could leave it this way to not add extra complexity.

rien:
A few extra things were changed:
- As discussed, the access token is now again in the exercise description path. Relative media files will always have this token, solving a lot of the issues they have.
- This removes the need for URL rewriting to happen in exercise descriptions. It still needs to be done for submissions, so this function is moved from `exercise.js` to `submission.js`.
- The token is now always added for media files fetched by the python-tutor.

app/policies/course_policy.rb
```diff
@@ -123,6 +123,10 @@ def heatmap?
     statistics?
   end
 
+  def media?
+    show?
+  end
```
charvp: Where is this used?

rien: This is used by the `set_course` callback in `exercise#media`.

rien:
> Why was the decision made to pass the token to the python tutor for public exercises as well?

I noticed that the Python tutor wasn't working for an exercise on Naos. But now I think about it, it could be that this is caused by Naos and Dodona not having the same exercise id in their database...

charvp:
I would revert the change that sends the token for public exercises. Like you said, it doesn't work because Naos and Dodona have different exercise ids.

pdawyndt:
If this issue is fixed, we might consider a hotfix. Since this issue was introduced in release 3.2, we have to make all exercises public just before an evaluation/exam, and make them private again afterwards.

charvp:
I was indeed planning on hotfixing once this is merged (hence the 3.2 milestone).

charvp:
Spoke too soon, I can't seem to access media files via the feedback table. Or was that last fix not deployed?

charvp:
It seems the exercise token was changed while I was testing, resulting in me not being able to access the media files. Works now.

PR #1625, opened 2020-01-06
Docker: don't limit device IO in development env

rien:
This pull request fixes a bug where container creation would hang on a device where there is no `/dev/sda` present. The current fix is to only limit IO to the `/dev/sda` block device in production and staging environments.

### DIFF
--- a/app/runners/submission_runner.rb
+++ b/app/runners/submission_runner.rb
@@ -87,11 +87,7 @@ class SubmissionRunner
     # fetch execution memory limit from submission configuration
     memory_limit = @config['memory_limit']
 
-    # process submission in docker container
-    # TODO: set user with the --user option
-    # TODO: set the workdir with the -w option
-    begin
-      container = Docker::Container.create(
+    docker_options = {
         # TODO: move entry point to docker container definition
         Cmd: ['/main.sh',
               # judge entry point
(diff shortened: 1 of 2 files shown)


app/runners/submission_runner.rb
```diff
@@ -104,12 +100,19 @@ def execute
         HostConfig: {
           Memory: memory_limit,
           MemorySwap: memory_limit, # memory including swap
-          BlkioDeviceWriteBps: [{ Path: '/dev/sda', Rate: 1024 * 1024 }],
+          # WARNING: this will cause the container to hang if /dev/sda does not exist
+          BlkioDeviceWriteBps: [{ Path: '/dev/sda', Rate: 1024 * 1024 }].filter{|| Rails.env.production? || Rails.env.staging? },
```
charvp: I would just do `!Rails.env.development?`. That makes it clearer IMO.

rien: This would also enable this limiting in tests, although we currently stub docker handling in the test environment (so this wouldn't give problems). That's why I explicitly specified production and staging environments in order to avoid confusion if we would actually start using docker in tests.

charvp: Hmm, true.

PR #1630, opened 2020-01-06
Fix double newline by overriding the copy behaviour of browsers

### LINKED ISSUE #1581: Code copied from code tab contains extra newlines
When selecting all the code in the code tab and then copying it, each line will have an extra newline. This is annoying.

MrKickkiller:
This pull request attempts to fix the double newline that is introduced by some browser for some odd reason.

Tested browsers:
 * Firefox Developer Edition 72.0b11 (64-bit) on Arch Linux
 * Chromium 79.0.3945.88 on Arch Linux

- [ ] Tests were added
No testing available with Jest since the Selection & Range API are required

Closes #1581 .

### DIFF
--- a/app/assets/javascripts/code_listing.ts
+++ b/app/assets/javascripts/code_listing.ts
@@ -20,6 +20,7 @@ export class CodeListing {
             console.error("The code listing could not be found");
         }
 
+        // Override the default copy behaviour of the browser to have a standard format for code copying.
         this.table.addEventListener("copy", function (e) {
             e.clipboardData.setData("text/plain", window.dodona.codeListing.getSelectededCode());
             e.preventDefault();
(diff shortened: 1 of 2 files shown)


charvp:
Works in Edge :+1:

app/assets/javascripts/code_listing.ts
```diff
@@ -89,7 +94,29 @@ export class CodeListing {
...
+                strings.push(v.textContent.trimRight());
+            });
+        }
+
+        // TODO: Fix this other way
+        // Somewhat hacky way to circumvent the bad filtering in Firefox
+        const documentFragment1 = selection.getRangeAt(selection.rangeCount - 1).cloneContents();
+        if (documentFragment1.firstElementChild.tagName == "PRE") {
+            strings.push(documentFragment1.textContent);
+        }
```
charvp: Can you explain in a comment what this code actually does? I can't tell from reading it.

MrKickkiller: Example source document:
```html
<tr><div class='c'>print( |2 + 3 )</div></tr>
<tr><div class='c'>print( 4 + 6 )</div></tr>
<tr><div class='c'>print( 5| + 4 )</div></tr>
```
Selection border marked by |

Chrome:
1 Selection with 1 Range. The clonedContents of the 0'th range contains the following:

```html
<div class='c'>2 + 3 )</div>
<div class='c'>print( 4 + 6 )</div>
<div class='c'>print( 5</div>
```

Firefox:
1 Selection with multiple Ranges

Range 0:
```html
<div class='c'>2+3)</div>
```
Range 1:
```html
<div class='c'>print( 4 + 6 )</div>
```

However Range 2 gets handled differently than Chrome:
```html
<pre>print( 5</pre>
```

Simply expanding the `querySelectorAll` to `.rouge-code, pre` does not work since that would result in duplicate content in the copy buffer. As the rouge-code elements have pre's inside them. Just selecting the pre's would also be infeasible since the line numbers are also wrapped in pre's

MrKickkiller: So, to not lose the final selected line, i'm hackishly adding it back if the last range of the `clonedContent` is simply a `<pre>`

charvp: After doing some experimenting myself, I don't see why a `pre` selection wouldn't work. The line numbers are not in the document fragment that `cloneContents()` returns.

app/assets/javascripts/code_listing.ts
```diff
@@ -89,7 +94,29 @@ export class CodeListing {
...
+
+    private getSelectededCode(): string {
+        const selection = window.getSelection();
+        const strings = [];
+
+        for (let rangeIndex = 0; rangeIndex < selection.rangeCount; rangeIndex++) {
+            const documentFragment = selection.getRangeAt(rangeIndex).cloneContents();
+            const fullNodes = documentFragment.querySelectorAll("td.rouge-code");
+            fullNodes.forEach((v, _n, _l) => {
+                strings.push(v.textContent.trimRight());
```
charvp: A `trimRight()` is not correct. If the code contains trailing whitespace we want to copy it as well.

app/assets/javascripts/code_listing.ts
charvp: And even if they are, you could delete them first using a simple `querySelectorAll(".rouge-gutter").foreach(node => node.remove())`

charvp: When you change the code, please document it extensively showing why things were implemented the way they were implemented.

PR #1635, opened 2020-01-09
Generalize token generation

### LINKED ISSUE #1233: generalize token generation
We regularly generate random tokens, but each time a slightly different generation method is used. We should make the token generation and token length uniform.

Possibly with some meta-programming? E.g. a `generates_token :access_token` class method.

### LINKED ISSUE #1634: Exercise access token changes with every update
The exercise access token changes with every update, revoking access for users who have an exercise description with an old token.

In the scenario of an exercise which is updated during an exam, this would require all students to refresh their pages, which is not ideal. Correcting a simple typo is enough to trigger this.

I suggest to change this behavior to only change access token when the permissions change. This fix could be combined with #1233.

rien:
This pull request centralized token generation to a `Tokenable` concern. Tokens are base58-ish (base64 without the characters 'oO0l1i') and we make sure a token is unique (unless explicitly mentioned not to).

Token generation changes:
 - `Course.secret` length stays 5 characters because students need to be able to type it over.
 - `Series.access_token` length changed to 5 characters because students need to be able to type it over.
 - `Exercise.access_token` will now be 16 characters instead of 12 in order to be consistent.
 - `Exercise.access_token` will now only be regenerated when:
    - `Exercise.access` has been changed,
    - `Series.visibility` has been changed (of a series containing the exercise),
    - The exercise is removed from an exercise (an associated `SeriesMembership` has been deleted).

- [x] Tests were added

Closes #1233 and fixes #1634.

### DIFF
--- a/app/controllers/series_controller.rb
+++ b/app/controllers/series_controller.rb
@@ -102,15 +102,22 @@ class SeriesController < ApplicationController
 
   def reset_token
     type = params[:type].to_sym
-    @series.generate_token(type)
-    @series.save
     value =
       case type
       when :indianio_token
+        @series.generate_indianio_token
         @series.indianio_token
       when :access_token
+        @series.generate_access_token
         series_url(@series, token: @series.access_token)
+      else
+        # unknown token type
+        head :unacceptable
       end
+
+    return if performed?
+
+    @series.save
     render partial: 'application/token_field', locals: {
       container_name: :access_token_field,
       name: type,
--- a/app/models/concerns/tokenable.rb
+++ b/app/models/concerns/tokenable.rb
@@ -5,7 +5,7 @@ module Tokenable
     def token_generator(name, length: 16, unique: true)
       token_name = name.to_sym
       raise "Model does not have attribute '#{name}'" \
-        unless self.column_names.include? token_name.to_s
+        unless column_names.include? token_name.to_s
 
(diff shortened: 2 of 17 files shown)


rien:
> Looks good, but any idea why tests are failing?

It looks like it has something to do with simplecov. Locally, the tests succeed.

> Also in exercise the changing of the access token explicitly avoided validations/callbacks. I'm pretty sure this was for a reason (though I don't remember it :upside_down_face:). Did you test that the behaviour around writing the config file is still correct? (I think it was related to that...)

I did not test it explicitly. I'll look into what could be the reason for the avoided validations.

rien:
Tests are broken apparently, it seems I requested a review a bit too early.

rien:
About the skipping of validations/callbacks. I couldn't pin down the exact reason, but I guess it got something to do with the tests. Because we stub exercises (so they are not backed by files), validations and callbacks would cause an exercise to be marked as `not_valid`, breaking al lot of tests.

The alternative token generation and changed invalidations caused extra callbacks, breaking these tests. So I added an `stub_all_exercises!` helper to stub all `Exercise` instances so they are always valid and `ok`. This helper is only added in tests which were broken to prevent stubbing too much functionality.

rien:
(This PR is now actually ready for review)

PR #1644, opened 2020-01-17
Update yarn.lock

rien:
This pull request updates the `yarn.lock` file so a few JavaScript dependencies got updated. None of our own dependencies' versions changed.

This fixes a security warning (but this warning was likely not an issue with us).


charvp:
I would prefer this happens via `yarn upgrade` (also fixes the vulnerability if I do it locally). Seems a bit cleaner to me.

PR #1687, opened 2020-02-06
Limit memory usage of statistics calculations

charvp:
Instead of loading all (new) submissions, the `in_batches` method is used to process 1000 new submissions at a time.


charvp:
I was going to test this on Naos, but https://github.com/dodona-edu/dodona-ansible/issues/52 prevented me from doing so.

charvp:
Works on naos.

PR #1710, opened 2020-02-18
Don't try to read result file if submission is queued or running

charvp:
This pull request should fix the occasional Gzip errors.


test/models/submission_test.rb
```diff
@@ -69,31 +69,31 @@ class SubmissionTest < ActiveSupport::TestCase
 
   test 'safe_result should remove staff tabs for students' do
     json = FILE_LOCATION.read
-    submission = create :submission, result: json
+    submission = create :submission, result: json, status: :correct
```
rien: We could also make the submission factory set the default status to `correct`.

PR #1734, opened 2020-02-28
Correct counting problem with hidden annotations

### LINKED ISSUE #1733: Show X other annotations is incorrrect
On https://dodona.ugent.be/nl/submissions/4945860/#code-tab I see "Only errors are shown, show 1 other annotation".

Screenshot with line: Only errors are shown, show 1 other annotation

When I click that nothing happens because there is only one annotation:
`jq -r '.result' 4945860.json  | jq '.annotations'`
```json
[
  {
    "type": "error",
    "text": "Error: submission/source:5:10: Syntax error: Unexpected end of file",
    "col": 9,
    "row": 4
  }
]
```

Also:
- the cursor is not pointer on that link
- the tooltip on `button#show_all_annotations` is now "show annotations", I think it should become "show *all* annotations"

MrKickkiller:
This pull request fixes the issues laid out in #1733 

- Count was incorrect due to only checking if > 1 when choosing 1 or many.
- Added specific css to make the link portion a cursor: pointer
- Locale was modified to be "Show all annotations" & "Toon alle annotaties"

- [x] Tests were added

Closes #1733  .

### DIFF
--- a/app/assets/javascripts/code_listing/code_listing.ts
+++ b/app/assets/javascripts/code_listing/code_listing.ts
@@ -151,6 +151,10 @@ export class CodeListing {
 
     public createHiddenMessage(count: number): HTMLSpanElement {
         const span = document.createElement("span");
+        if (count === 0) {
+            return null;
+        }
+
         const spanText: Text = document.createTextNode(I18n.t("js.annotation.were_hidden.first") + " ");
         span.appendChild(spanText);
 
--- a/app/assets/javascripts/code_listing/code_listing.ts
+++ b/app/assets/javascripts/code_listing/code_listing.ts
@@ -152,7 +152,7 @@ export class CodeListing {
     public createHiddenMessage(count: number): HTMLSpanElement {
         const span = document.createElement("span");
         if (count === 0) {
-            return null;
+            return span;
         }
 
         const spanText: Text = document.createTextNode(I18n.t("js.annotation.were_hidden.first") + " ");
--- a/app/assets/javascripts/code_listing/code_listing.ts
+++ b/app/assets/javascripts/code_listing/code_listing.ts
@@ -67,9 +67,13 @@ export class CodeListing {
             this.annotationsWereHidden.classList.remove("hide");
         }
 
-        const nonErrorAnnotationCount = this.createHiddenMessage(this.annotations.filter(m => m.type !== "error").length);
-        this.annotationsWereHidden.innerHTML = "";
-        this.annotationsWereHidden.appendChild(nonErrorAnnotationCount);
+
+        const errorCount = this.annotations.filter(m => m.type !== "error").length;
+        if (errorCount > 0) {
+            const nonErrorAnnotationCount = this.createHiddenMessage(errorCount);
+            this.annotationsWereHidden.innerHTML = "";
+            this.annotationsWereHidden.appendChild(nonErrorAnnotationCount);
+        }
     }
 
     compressAnnotations(): void {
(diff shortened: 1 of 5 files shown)


app/assets/stylesheets/components/code_listing.css.less
```diff
@@ -2,6 +2,10 @@
 
 .code-table {
 
+  #annotations-were-hidden a {
+    cursor: pointer;
```
charvp: If you add a `href="#"` to the `a` element, you don't need any styling for it.

app/assets/javascripts/code_listing/code_listing.ts
```diff
@@ -151,6 +151,10 @@ export class CodeListing {
 
     public createHiddenMessage(count: number): HTMLSpanElement {
         const span = document.createElement("span");
+        if (count === 0) {
```
charvp: it feels like this method shouldn't be called if there is no message to show.

app/assets/stylesheets/components/code_listing.css.less
MrKickkiller: I believe I have tried this, but I would have had to insert some more javascript to stop the page from reloading since clicking a `<a href="#"></a>` will navigate to the same page.

charvp: Just `return false;` at the end of your event handler or call `event.preventDefault();`

app/assets/javascripts/code_listing/code_listing.ts
```diff
@@ -151,12 +155,17 @@ export class CodeListing {
...
+
         const spanText: Text = document.createTextNode(I18n.t("js.annotation.were_hidden.first") + " ");
         span.appendChild(spanText);
 
         const link: HTMLAnchorElement = document.createElement("a");
         const data: string = I18n.t(`js.annotation.were_hidden.second.${count > 1 ? "plural" : "single"}`).replace(/{count}/g, String(count));
         const linkText: Text = document.createTextNode(data);
+        link.href = "#";
+        link.addEventListener("click", function (ev) {
+            ev.preventDefault();
```
charvp: Where is the event handler for expanding all annotations? You should add this there.

MrKickkiller: It is already attached here, giving the user a bit of leniency about where to click.

https://github.com/dodona-edu/dodona/blob/ab51ea48598b067ce0332c6c0c183db69e5e3ff3/app/assets/javascripts/code_listing/code_listing.ts#L51

charvp: Do we want this? The whole point of a link is to make users know that it's clickable, so I don't think we need clickability outside of the link. @bmesuere Do you have an opinion on this?

app/assets/stylesheets/components/code_listing.css.less
bmesuere: +1 for using href without css pointer

app/assets/javascripts/code_listing/code_listing.ts
bmesuere: only make link clickable, it's what users expect

config/locales/views/submissions/en.yml
```diff
@@ -51,7 +51,7 @@ en:
         hidden: Hide correct tests
       annotations:
         title: Annotations
-        show_all: Show annotations
+        show_all: Show all annotations
         hide_all: Hide annotations
```
bmesuere:
```suggestion
        hide_all: Hide all annotations
```

config/locales/views/submissions/nl.yml
```diff
@@ -51,7 +51,7 @@ nl:
         hidden: Verberg correcte tests
       annotations:
         title: Annotaties
-        show_all: Toon annotaties
+        show_all: Toon alle annotaties
         hide_all: Verberg annotaties
```
bmesuere:
```suggestion
        hide_all: Verberg alle annotaties
```

PR #1755, opened 2020-03-03
Encode with universal newline

MrKickkiller:
Encodes each string to have a universal newline
Should solve the issue of high comment lines when code is inputted with Windows newlines


MrKickkiller:
@charvp 
image

Top left: Online tool where I pasted the content of the file you linked to me.
Top right: Writing the file you linked to my clipboard
Bottom: Submission made with the contents of my clipboard.

charvp:
I'm not convinced that is a sufficient test. Does the code file that Dodona created on your filesystem contain CRLF line endings? I'm not sure whether browsers normalize line endings in input fields.

rien:
You could write a test which submits solutions with all kinds of weird newlines.

PR #1777, opened 2020-03-23
Add reviews for course admins

MrKickkiller:
This pull request adds a system where course administrators can create a review session, and within that review session, many reviews can be made on the submissions of the students.

- [x] Tests were added


pdawyndt:
Now I’m really curious to see this at work.

charvp:
Answers to some questions up until where I got today:

> Series selection is fixed: would it make sense to make this a select?

The reason I didn't make this a select is because an evaluation is very much tied to a series. There is no way to create an evaluation from the course, and there is no way to list the evaluations in a course. It also didn't seem that necessary to me to add those functionalities immediately, since the usecase we have in mind is for tests/examinations, where starting the evaluation from the series makes sense (at least to me).

> Making a subselection of exercises is not included in the first version?

The reasoning here is similar to the previous point: we expect this functionality to be used for tests/examinations where not giving feedback on all exercises seems like an exception. This can of course be added later, but didn't seem necessary to me for this to be used in the upcoming exams.

pdawyndt:
Now I’m even more curious to see this at work. Looking forward to start using this!

app/views/series/_series.html.erb
```diff
@@ -71,6 +71,19 @@
                 <% end %>
               </li>
             <% end %>
+            <% if policy(series).create_review_session? %>
```
bmesuere, line 74: move this higher in the menu, just below delete

config/locales/views/review_sessions/nl.yml
```diff
@@ -0,0 +1,46 @@
+nl:
+  review_sessions:
+    show:
+      view_session: "Toon evaluatie"
```
bmesuere, line 4:
```suggestion
      view_session: "Evaluatie bekijken"
```

config/locales/views/review_sessions/nl.yml
```diff
@@ -0,0 +1,46 @@
...
+      release: Feedback vrijgeven
+      unrelease: Feedback verbergen
+      deadline_html: De oplossingen voor deze evaluatie werden bepaald met als deadline <b>%{deadline}</b>.
+      next_incomplete_review: Volgende onafgewerkte oplossing
+      progress:
+        finished: "%{review_count} oplossingen hebben afgewerkte feedback."
+        unfinished: "%{review_count} oplossingen hebben nog geen afgewerkte feedback."
+    new:
+      create_session: Evalueren
+      explanation_html: "Bij het evalueren van een reeks kies je eerst een deadline. Dan kies je de gebruikers die in de evaluatie opgenomen werden en geef je hen feedback. Wanneer je de feedback vrijgeeft krijgen de gebruikers in de evaluatie een notificatie."
```
bmesuere, line 16:
```suggestion
      explanation_html: "Het evalueren van een reeks is een gestructureerde manier om de ingediende oplossingen van een reeks te bekijken en er feedback op te geven. Dodona houdt bij welke oplossingen al bekeken zijn zodat je het evaluatiewerk eenvoudig kan verdelen over meerdere lesgevers. De feedback die je geeft is niet onmiddelijk zichtbaar voor de studenten; je kiest zelf wanneer je deze vrijgeeft."
```

config/locales/views/review_sessions/nl.yml
```diff
@@ -0,0 +1,46 @@
...
+      next: Volgende
+    edit:
+      title: Gebruikers aanpassen
+      delete_title: Evaluatie verwijderen
+      delete_consequences: Het verwijderen van een evaluatie zal ook alle feedback die binnen de evaluatie gegeven werd verwijderen.
+      add_enrolled: Studenten in deze evaluatie vervangen door alle ingeschreven studenten
+      add_submitted: Studenten in deze evaluatie vervangen door alle studenten die indienden
+    destroy: Verwijderen
+    form:
+      deadline-help_html: "De deadline wordt gebruikt om de default oplossingen te selecteren waarop je feedback geeft. De laatste oplossing voor deze deadline wordt gebruikt. <b>Let op:</b> je kan de deadline achteraf niet meer aanpassen."
```
bmesuere, line 26:
```suggestion
      deadline-help_html: "Voor elke student zal de laatst ingediende oplossing voor deze deadline opgenomen worden in de evaluatie. <b>Let op:</b> je kan deze deadline achteraf niet meer aanpassen, maar kan voor individuele studenten steeds een andere oplossing selecteren."
```

config/locales/views/review_sessions/en.yml
```diff
@@ -0,0 +1,46 @@
...
+      release: Release feedback
+      unrelease: Unrelease feedback
+      deadline_html: The submissions for this evaluation were chosen with <b>%{deadline}</b> as a deadline.
+      next_incomplete_review: Next incomplete submissions
+      progress:
+        finished: "%{review_count} submissions have completed feedback."
+        unfinished: "%{review_count} submissions don't have completed feedback."
+    new:
+      create_session: Evaluate
+      explanation_html: "When evaluating a series you first choose a deadline. Then you select the users included in the evaluation and give them feedback. When you release the feedback, the users in the evaluation get a notification."
```
bmesuere, line 16:
```suggestion
      explanation_html: "Evaluating a series is a structured way to give feedback on all submitted solutions of a series. Dodona keeps track of which submissions are already checked, so you can easily split work between multiple course administrators. The feedback you give isn't immediately visible for the students; you can decide when it is released."
```

charvp:
Making the user select a three-way toggle button isn't trivial: we would need to calculate which option was last selected when rendering the view (or store it in the db, but I'm not a fan of that). When a user is removed or added manually, there is also no sane option to select as default.

bmesuere:
UI rework of the creation form:
image

pdawyndt:
Maybe we should mention "multiple course admins" instead of "multiple lecturers" in the text at the right hand side of creation form.

pdawyndt:
Are you convinced about the term "evaluation/evaluatie"? I find it rather confusing with what I call an evaluation (outside of the Dodona context).

bmesuere:
> Are you convinced about the term "evaluation/evaluatie"? I find it rather confusing with what I call an evaluation (outside of the Dodona context).

I did a search for how this is called in other tools. The majority (including Brightspace) uses evaluation/to evaluate/feedback to talk about this (and it translates nicely). The only other common term that I have found in English is "grading", but since we don't give a grade (for now) this doesn't seem applicable. I know you also use the term evaluation for the examination itself, but most other courses seem to use the term "test" for that (other tools use "assignment", "task" and "test" for the thing that is to be evaluated).

pdawyndt:
@bmesuere OK, I could see myself adopting the term "test" for what I now call an evaluation, and then the use of "evaluation" makes more sense.

pdawyndt:
A one-to-one relation between a series and an evaluation, with a fixed deadline after creation, means you run into trouble when a wrong deadline was picked (all you can do is pick another solution for each student one-by-one). In later iterations we'll need some refinements.

bmesuere:
@pdawyndt I see no reason to stop using the term evaluation in real life because Dodona uses it for something slightly different. I don't think students ever come into contact with "evaluations" on Dodona (but I still have to review that part). The phrasing could be like "someone evaluated your code, here is your feedback".

bmesuere:
image

pdawyndt:
Also add option to remove all course admins grom selected users?

Also add option to exclude course managers when adding users?

charvp:
None of the presets add course admins to the selection. (Hence the student terminology instead of course member.)

bmesuere:
image

bmesuere:
image

bmesuere:
image

bmesuere:
Some ideas about the feedback page, will try to implement tomorrow:
- "oplossing veranderen" button should not be present if there are no or only the current submission. It probably better if this would be a link instead of a button
- maybe in a next version we can make this section more interesting by showing a (visual?) submission history and include the submission history there
- the navigation part is too complex: there are too many buttons:
  - since the number of exercises in a series will probably be limited, remove the buttons (and progressbar) and show the individual submissions/feedbacks instead by reusing the representation of the details table
  - remove previous, the browser back button is good enough
  - the feedback items are already randomized, so random and next are more or less the same: remove random and only keep the next button, but add a checkbox to skip already completed feedback. This should look and behave the same as the "auto mark as done" checkbox (save status in browser)
- there needs to be a bigger difference between complete and incomplete feedback. Right now, there is only a very slight change in wording on the button. Maybe add an icon?
- possible order of the right part of the page from top to bottom: check/uncheck button, exercise progress, next button, checkbox to skip, checkbox to autocomplete, hline, user stuff

rien:
- If a series has no deadline, the current moment is used as a deadline. Maybe set this to nil?
- If you remove a deadline and start the review, it crashes on a validation:
EDlJ

beardhatcode:
All comments from rien+

```diff
-1 students are registered for this course select these users
+1 student is registered for this course select this user
```

maybe "select" istead of select users
image

screenshot_2020-05-29-152438

Tooltip for progress bar is confusing if students did not hand in something. Maybe there should be a grey part and a tooltip like "4 students did not hand in anything, ..."


https://naos.ugent.be/en/feedbacks/349094951/

```
ActionController::UrlGenerationError in Feedbacks#show

Showing /home/dodona/rails/releases/20200529123720/app/views/feedbacks/_feedback_actions.html.erb where line #60 raised:

No route matches {:action=>"show", :controller=>"feedbacks", :id=>nil, :locale=>:en}, possible unmatched constraints: [:id]
````

File: app/models/series.rb [hunk 6]
```diff
@@ -36,6 +36,8 @@ class Series < ApplicationRecord
   has_many :activities, through: :series_memberships
   has_many :activity_statuses, dependent: :destroy
 
+  has_one :evaluation, dependent: :destroy
```
bmesuere, line 39: Potentially unwanted effect

app/views/feedbacks/_feedback_actions.html.erb
```diff
@@ -0,0 +1,64 @@
+<% content_for :javascripts do %>
+  <%= javascript_pack_tag 'evaluation' %>
+<% end %>
+
+<% feedback_pointers = @feedback.siblings %>
+<% state = !@feedback.completed ? "completed": "unfinished" %>
```
bmesuere, line 6: no longer used, I think?

config/locales/js/en.yml
```diff
@@ -106,3 +106,4 @@ en:
       fields:
         annotation_text: 'Annotation text'
       delete_confirm: 'Are you sure you want to delete this annotation?'
+      not_released: This annotation is not yet visible for the student
```
bmesuere, line 109:
```suggestion
      not_released: This annotation is not yet visible for students
```

charvp:
> If a series has no deadline, the current moment is used as a deadline. Maybe set this to nil?

The choice was made explicitly to set this to the current time. See https://github.com/dodona-edu/dodona/pull/1777#pullrequestreview-416123139

> Tooltip for progress bar is confusing if students did not hand in something. Maybe there should be a grey part and a tooltip like "4 students did not hand in anything, ..."

Not sure which progress bar you're talking about. Can you expand on this a bit more @beardhatcode?

charvp:
Deployed to naos.

config/locales/views/evaluations/en.yml
```diff
@@ -0,0 +1,80 @@
...
+      user: User
+    feedback_status:
+      feedback_finished: evaluation completed
+      feedback_unstarted: evaluation not yet completed
+      submission_not-started: "No submission, "
+      submission_correct: "Correct submission, "
+      submission_wrong: "Wrong submission, "
+    overview:
+      title: Evaluation overview
+      explanation: A course admin evaluated your submissions for the series "%{series}". Your last submission before %{deadline} was used, unless the course admin changed it. Note that this doesn't necessarily mean feedback was added to your submissions.
```
bmesuere, line 64:
```suggestion
      explanation: A course admin evaluated your submissions for the series "%{series}". For each exercise, your last submission before %{deadline} was automatically selected, but the course admin might have manually selected a different one. Note that this evaluation doesn't necessarily mean that feedback was added to all of your submissions.
```

config/locales/views/evaluations/nl.yml
```diff
@@ -0,0 +1,80 @@
...
+      user: Gebruiker
+    feedback_status:
+      feedback_finished: evaluatie afgewerkt
+      feedback_unstarted: evaluatie nog niet afgewerkt
+      submission_not-started: "Geen oplossing, "
+      submission_correct: "Juiste oplossing, "
+      submission_wrong: "Foute oplossing, "
+    overview:
+      title: Evaluatie-overzicht
+      explanation: Een cursusbeheerder evalueerde je oplossingen voor de reeks "%{series}". Je laatste oplossing voor %{deadline} werd gebruikt, behalve als de cursusbeheerder dit veranderde. Merk op dat dit niet noodzakelijk betekent dat er feedback toevoegd werd aan je oplossingen.
```
bmesuere, line 64:
```suggestion
      explanation: Een cursusbeheerder evalueerde je oplossingen voor de reeks "%{series}". Voor elke oefening werd je laatst ingediende oplossing voor %{deadline} automatisch geselecteerd, maar de cursusbeheerder selecteerde mogelijks handmatig een andere oplossing. Merk op dat deze evaluatie niet noodzakelijk betekent dat er feedback werd toevoegd aan elk van je oplossingen.
```

config/locales/views/evaluations/en.yml
```diff
@@ -0,0 +1,80 @@
...
+      released: Feedback was added to your code.
+      exercise: Exercise
+      no_annotations: "# annotations"
+      annotation_count:
+        other: "No annotations"
```
bmesuere, line 69:
```suggestion
        zero: "No feedback was given"
```

config/locales/views/evaluations/nl.yml
```diff
@@ -0,0 +1,80 @@
...
+      released: Er werd feedback gegeven op je code
+      exercise: Oefening
+      no_annotations: "# annotaties"
+      annotation_count:
+        zero: "Geen annotaties"
```
bmesuere, line 69:
```suggestion
        zero: "Er werd geen feedback gegeven"
```

charvp:
Deployed to naos.

app/assets/javascripts/evaluation.ts
```diff
@@ -48,6 +48,7 @@ function interceptFeedbackActionClicks(
         if (nextButton.getAttribute("disabled") === "1") {
             return;
         }
+        nextButton.setAttribute("disabled", "1");
```
bmesuere, line 51: This does not work as expected. The button is indeed disabled, but only till the first request (setting the evaluation to done) comes back and replaces the action button by a new one which isn't disabled.

PR #1787, opened 2020-03-25
Annotation tweaks

### LINKED ISSUE #1736: Better tooltip text for hidden annotations
Tooltip for hidden annotations says "An annotation was hidden" even if multiple annotations were hidden. May be better to say "Multiple hidden annotations" in this case.

image

### LINKED ISSUE #1742: Copy submitted solution to clipboard from feedback page
Add an option to copy the submitted solution to the clipboard from the feedback page.

### LINKED ISSUE #1758: Remove spacing between tab headers and code in feedback table
There may be extra spacing between the tab headers and the source code in the code tab of the feedback table, in case there are no feedback table options to be displayed.

image

### LINKED ISSUE #1767: Known issues with user annotations
- [x] typescript code should be refactored, eg the creation of the annotation div is in the code twice
- [x] still relatively hard to click the add annotation button. It might help to increase the row height of the code listing with a few pixels
- [x] add timestamp to the user annotation
- [x] the badge of the code tab doesn't include the number of user annotations
- [x] add small info text under the text area mentioning markdown and shift+enter
- [x] move focus to the text area when adding an annotation
- [x] fix copy to clipboard. Do we still need the workaround to exclude annotations now that they can be hidden? We could restore the default behaviour and add a copy button (#1742)
- [x] some kind of overview of annotations (both received and given?) (#1774)

bmesuere:
This pull request adds the remarks mentioned in #1767: 

- refactored code

- move focus to the text area when adding an annotation

- add help text

- display creation timestamp
image

- update badge count to include user annotations
badge

- no longer possible to create annotations at lines that don't exist using the api

- spurious whitespace removed if no annotations are present
image

- copy submitted solutions to the clipboard
image

- global (non-line bound) annotations
image

- reworked error handling
old:
image
new:
image

Fixes #1736, Fixes #1742, Fixes #1758, Fixes #1767.


thepieterdc:
@bmesuere Ready for review; I cannot assign you since you opened this PR :(

The idea of the refactor was to make the annotations somewhat agnostic of where they are being rendered, and the code listing agnostic to what kind of annotations it is dealing with. This should allow for flexibility, should a third annotation type ever need to be added (e.g. questions of students)

The clipboard-icon is not yet production-ready, I'm not sure how I would cope with this, any advice?

config/locales/js/nl.yml
```diff
@@ -91,6 +91,7 @@ nl:
           single: toon 1 andere annotatie
           plural: toon {count} andere annotaties
     user_annotation:
+      meta: "%{user} op %{time}"
```
bmesuere, line 99:
```suggestion
      meta: "Door %{user} op %{time}"
```

config/locales/js/en.yml
```diff
@@ -91,6 +91,7 @@ en:
           single: show 1 other annotation
           plural: show {count} other annotations
     user_annotation:
+      meta: "%{user} at %{time}"
```
bmesuere, line 99:
```suggestion
      meta: "By %{user} at %{time}"
```

app/assets/javascripts/code_listing/code_listing.ts
```diff
@@ -13,13 +13,15 @@ export class CodeListing {
...
 
+    private badge: HTMLSpanElement;
     private hideAllButton: HTMLButtonElement;
     private showOnlyErrorButton: HTMLButtonElement;
     private showAllButton: HTMLButtonElement;
     private annotationsWereHidden: HTMLSpanElement;
     private diffSwitchPrefix: HTMLSpanElement;
 
     constructor(code: string) {
+        this.badge = document.querySelector("span.badge[data-description=code]");
```
bmesuere: What's the benefit of doing this with a data attribute? A class name is a lot faster to query.

app/assets/javascripts/code_listing/code_listing.ts
```diff
@@ -65,6 +67,25 @@ export class CodeListing {
         this.showOnlyErrorButton.addEventListener("click", () => this.compressAnnotations());
     }
 
+    private initCopyToClipboard(): void {
+        const $clipboardBtn = $("#copy-to-clipboard");
```
bmesuere: Any reason jquery is used here? We try not to write new code using it.

config/locales/js/nl.yml
thepieterdc: I think it would make more sense to lowercase the first word, opinion?

config/locales/js/en.yml
thepieterdc: I think it would make more sense to lowercase the first word, opinion?

app/assets/javascripts/code_listing/code_listing.ts
thepieterdc: If I recall correctly, this data attribute was already being set by the feedback code renderer, so I decided to simply use that to find out which badge belongs to the code tab. I will look into this however.

app/assets/javascripts/code_listing/code_listing.ts
thepieterdc: This code was copied from https://github.com/dodona-edu/dodona/blob/022f995b1ff4c4a67a7c820897ad731e8f60ead8/app/assets/javascripts/util.js#L5 (which I cannot use directly), I'm not a fan of jQuery myself either, but I thought that since ClipboardJS is a jQuery library, this would be required. I will take a look at this.

config/locales/js/en.yml
bmesuere: How about dropping the "By" and "at", and using "name · timestamp" (that's a middle dot)

app/assets/javascripts/code_listing/code_listing.ts
```diff
@@ -1,308 +1,348 @@
...
+    // /////////////////////////////////////////////////////////////////////////
+    // Show and hide annotations ///////////////////////////////////////////////
+    // /////////////////////////////////////////////////////////////////////////
 
-    handleCommentButtonClick(lineNumber: number, codeLine: HTMLTableRowElement): void {
-        const tableRowId: number = codeLine.rowIndex;
+    public hideAnnotations(keepImportant = false): void {
+        this.annotations.forEach((annotations, line) => {
+            // Do not hide global annotations.
+            if (line == 0) {
```
bmesuere, line 284:
```suggestion
            if (line === 0) {
```

app/assets/javascripts/code_listing/code_listing.ts
```diff
@@ -1,308 +1,348 @@
...
+
+            // Find the colour of the dot for this line.
+            const dotColourIdx = annotations
+                .filter(annotation => !annotation.important || !keepImportant)
+                .map(annotation => annotation.type)
+                .map((type: AnnotationType) => ANNOTATION_ORDER.indexOf(type))
+                .sort()[0];
+
+            // Configure the dot.
+            if (dotColourIdx != null) {
```
bmesuere, line 299:
```suggestion
            if (dotColourIdx !== null) {
```

app/assets/javascripts/code_listing/code_listing.ts
```diff
@@ -311,81 +351,172 @@ export class CodeListing {
               ${I18n.t("js.user_annotation.cancel")}
             </button>
             <button class="btn btn-text btn-primary annotation-control-button annotation-submission-button" type="button">
-              ${edit ? I18n.t("js.user_annotation.update") : I18n.t("js.user_annotation.send")}
+              ${annotation != null ? I18n.t("js.user_annotation.update") : I18n.t("js.user_annotation.send")}
```
bmesuere, line 354:
```suggestion
              ${annotation !== null ? I18n.t("js.user_annotation.update") : I18n.t("js.user_annotation.send")}
```
Search for ` == ` and ` != ` and replace them.

app/assets/javascripts/code_listing/code_listing.ts
thepieterdc: I thought TypeScript had finally fixed this, apparently not

thepieterdc:
> Codelisting became even larger, but I don't really know how to split it up.

Me neither, splitting up wouldn't make much sense. Keep in mind that comments were added as well, those also contribute to the length of the file.

charvp:
>     * removing the last annotation doesn't remove the notification @charvp: is this possible?

Should definitely possible in the `after_destroy` of an annotation.

app/assets/javascripts/code_listing/code_listing.ts
```diff
@@ -1,278 +1,348 @@
...
-                this.annotationsWereHidden.appendChild(nonErrorAnnotationCount);
-            }
+        if (!this.annotations.has(line)) {
+            this.annotations.set(line, []);
+        }
+
+        // Add the annotation in the map.
+        const lineAnnotations = this.annotations.get(line);
+        lineAnnotations.push(annotation);
+        this.annotations.set(line, lineAnnotations);
```
charvp, line 94:
```suggestion
        this.annotations.get(line).push(annotation);
```
Isn't this equivalent?

app/assets/javascripts/code_listing/code_listing.ts
```diff
@@ -1,278 +1,348 @@
...
-        const formSubmissionRow: HTMLTableRowElement = this.table.insertRow(lineNumber);
-        formSubmissionRow.setAttribute("class", "annotation-set");
-        formSubmissionRow.setAttribute("id", `annotation-row-id-${rougeRow}`);
-        const gutterTD: HTMLTableDataCellElement = formSubmissionRow.insertCell();
-        gutterTD.setAttribute("class", "rouge-gutter gl");
-        const annotationTD: HTMLTableDataCellElement = formSubmissionRow.insertCell();
-        annotationTD.setAttribute("class", "annotation-cell");
-        return formSubmissionRow;
+    // noinspection JSUnusedGlobalSymbols used by FeedbackCodeRenderer as this crashes unit tests if
+    // called in constructor.
```
charvp, line 265: Do you know why it crashes in the constructor when called in the constructor? Is it mockable? I prefer not to do changes like this just to appease the tests.

app/assets/javascripts/code_listing/code_listing.ts
```diff
@@ -1,278 +1,348 @@
...
-        // Remove previous submission window
-        const tr: HTMLTableRowElement = this.findOrCreateTableRow(lineNumber, tableRowId + 1, lineNumber);
-        const existingAnnotationSubmissionDiv: HTMLFormElement = tr.querySelector("form.annotation-submission");
-        if (existingAnnotationSubmissionDiv) {
-            return;
-        }
+            // Find the dot for this line.
+            const dot = this.table.querySelector<HTMLSpanElement>(`#dot-${line}`);
+
+            // Find the colour of the dot for this line.
```
charvp, line 291: Finding the colour of the dot should not be necessary. The ordering of the CSS does this for us. Just adding all relevant classes to the dot should work.

app/assets/javascripts/code_listing/code_listing.ts
```diff
@@ -281,77 +351,172 @@ export class CodeListing {
...
+            try {
+                const updated = await annotation.update(annotationData);
+                this.updateAnnotation(annotation, updated);
+            } catch (err) {
+                inputField.classList.add("validation-error");
+            }
+        };
+
+        const formId = `annotation-submission-update-${annotation.id}`;
+        return this.createAnnotationForm(formId, annotation, onSubmit, () => callback());
```
charvp, line 445:
```suggestion
        return this.createAnnotationForm(formId, annotation, onSubmit, callback);
```
Isn't this the same?

app/models/submission.rb
```diff
@@ -95,6 +95,7 @@ def initialize(params)
     super(params)
     # We need to do this after the rest of the fields are initialized, because we depend on the course_id, user_id, ...
     self.code = code.to_s unless code.nil?
+    self.line_count = self.code.lines.length unless code.nil?
```
charvp, line 98: Can't we do this update in the `code=` method? It would make sense to keep those together I think. Right now we never update code, but still.

app/assets/javascripts/code_listing/code_listing.ts
thepieterdc: This is not the same as it changes the context of `this` in the callback.

app/assets/javascripts/code_listing/code_listing.ts
thepieterdc: image

thepieterdc:
I have implemented all remarks, and fixed #1736 as well.

PR #1790, opened 2020-03-29
Add time and memory limit status to partial format

niknetniko:
This pull request adds the status `memory limit exceeded` and `time limit exceeded` to the partial output format.

- [x] Tests were added
- [x] Documentation updates: not needed, documentation already says this is possible (see https://dodona-edu.github.io/en/guides/creating-a-judge/)

### DIFF
--- a/public/schemas/partial_output.json
+++ b/public/schemas/partial_output.json
@@ -197,6 +197,8 @@
       "enum": [
         "internal error",
         "compilation error",
+        "memory limit exceeded",
+        "time limit exceeded",
         "runtime error",
         "wrong", "wrong answer",
         "correct", "correct answer"
--- a/test/runners/result_constructor_test.rb
+++ b/test/runners/result_constructor_test.rb
@@ -269,6 +269,18 @@ class ResultConstructorTest < ActiveSupport::TestCase
       '{ "command": "close-judgement", "status": { "enum": "runtime error", "human": "Runtime" } }'
     ])
     assert_equal('runtime error', result[:status])
+    result = construct_result([
+      '{ "command": "start-judgement" }',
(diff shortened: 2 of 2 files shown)


ninewise:
Could you clarify why this is required? The partial output format
automatically assigns these statusses to the submission if it killed the
judge because it exceeded either limit. In fact, the whole point of this
output format was eliminating the need to check for these in the judge.

pdawyndt:
@ninewise Is there any reporting on finished (closed) test cases/contexts/tabs in case Dodona traps a timeout or memory overflow?

ninewise:
Quoting Peter Dawyndt (2020-03-30 09:19:33)
> Is there any reporting on finished (closed) test cases/contexts/tabs
> in case Dodona traps a timeout or memory overflow?

Yes. Every testcase emitted before the kill is reported on. The last
(and yet unclosed) emitted test/testcase/... is closed and marked as
incorrect. The status of the submission is set to either 'time lmit
exceeded' or 'memory limit exceeded'. Code can be found here [1].

[1]: https://github.com/dodona-edu/dodona/blob/develop/app/runners/result_constructor.rb#L44-L59

charvp:
I'm in favor of merging this. This doesn't increase the complexity of the result constructor at all (it's code isn't even touched), and allows judges using the partial output format to do more.

bmesuere:
@ninewise what's your final say on this?

ninewise:
I wouldn't really mind, as Charlotte says it barely increases
complexity. But I'd still like to have it documented why a judge should
be able to say it ran out of time/memory. If judges do not actually need
this, they should not be able to do it.

charvp:
One reason for memory at least is that depending on how the language handles out of memory errors the judge might crash and result in an internal error.

PR #1814, opened 2020-04-06
Add link to make series publicly visible. Fixes #1705.

### LINKED ISSUE #1705: Button to make hidden series visible
I think it would be nice if there was a button to make a series visible directly form the info banner (maybe with an additional confirmation box to prevent accidental clicks). Now this action takes 4 clicks.

Banner with the text "This series is only visible for students using the secret link!" and a link to make it visible

thepieterdc:
This pull request adds a link to the information message that allows the user to make the series visible to students. Layout-wise this is identical to the impersonation-feature.

Schermafdruk van 2020-04-06 10 58 52

Closes #1705.

### DIFF
--- a/app/views/series/_series.html.erb
+++ b/app/views/series/_series.html.erb
@@ -117,7 +117,10 @@
     </div>
   <% end %>
   <% if current_user&.course_admin?(series.course) && !series.exercises_visible %>
-    <div class="alert alert-info hidden-print"><%= t "series.show.exercises_not_visible" %></div>
+    <div class="alert alert-info hidden-print">
+      <%= t "series.show.exercises_not_visible" %>
+      <%= link_to(t('series.show.make_visible.link'), series_path(series, series: {exercises_visible: true}), method: :patch) + " " +  t('series.show.make_visible.for_students') %>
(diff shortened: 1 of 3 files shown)


thepieterdc:
The same message has also been added to make hidden exercises of series visible, as well as make closed series available to students.

app/views/series/_series.html.erb
```diff
@@ -105,13 +105,22 @@
     </div>
   </div>
   <% if current_user&.course_admin?(series.course) && series.hidden? %>
-    <div class="alert alert-info hidden-print"><%= t "series.show.series_not_visible" %></div>
+    <div class="alert alert-info hidden-print">
+      <%= t "series.show.series_not_visible" %>
+      <%= link_to(t('series.show.make_visible.link'), series_path(series, series: {visibility: :open}), method: :patch) + " " +  t('series.show.make_visible.for_students') %>
```
charvp, line 110: I would prefer this to be one translation and for the translation to take the link as a parameter. This works for the current languages, but the order of "Make visible" and "for students" might not be the same if we add another language in the future.

PR #1815, opened 2020-04-06
Fix focus issue with lightbox in exercise descriptions

### LINKED ISSUE #1759: Hide image popups with escape key
When images in exercise descriptions are clicked, they are displayed in a popup on top of the page. Before the popup could be hidden by clicking the escape key, but this no longer seems the case. At least not in Chrome on Windows.

thepieterdc:
This pull request fix an issue with focus after a lightbox is opened in an exercise description. I have added a `tabindex` attribute to the document body, to allow the parent document to be focused. This could be transferred to another attribute specific to the exercise-show page as well, but having a focusable body might prevent future headaches.

**Consequences:**
- Lightboxes can now be closed with the `ESC` key without explicitly having to click them.
- Users can use arrow keys to cycle between images in the lightbox.

Closes #1759  .

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -89,7 +89,7 @@
   <%= csrf_meta_tags %>
   <%= csp_meta_tag %>
 </head>
-<body>
+<body tabindex="-1">
 <div id="page-wrapper">
   <%= yield :modal %>
   <%= render 'layouts/navbar' %>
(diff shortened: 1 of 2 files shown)


app/views/layouts/application.html.erb
```diff
@@ -89,7 +89,7 @@
   <%= csrf_meta_tags %>
   <%= csp_meta_tag %>
 </head>
-<body>
+<body tabindex="-1">
```
charvp, line 92: Just curious, why is this required?

PR #1834, opened 2020-04-09
Caching & deadline displays

### LINKED ISSUE #1711: Confusing icons for overdue series
Series for which the deadline has been passed, appear with a red clock on the homepage, which is confusing for those who finished said series. Is there an option to replace the red clock with another icon so it's clear that the deadline has passed, but the student has finished the exercise in time.

### LINKED ISSUE #1805: Add per student series status and exercise status to cache
This would allow us to use the status on the homepage (so that we don't display a red clock for a missed deadline when all exercises were correctly solved before).

thepieterdc:
This pull request adds the completion status of an exercise and a series for a given user to the cache and updates the display of a completed deadline.

**Already completed deadline in the future**
image

**Completed deadline in the past**
image

**Missed deadline in the past**
image

**Not yet completed deadline in the future:**
image

**TODO:**
- [x] Cache series completion status
- [x] Cache exercise completion status
- [x] Update homepage
- [x] Invalidate series cache
- [x] Invalidate exercise cache
- [x] Remove transient models

Fixes #1711. Fixes #1805.


thepieterdc:
I cannot seem to get cache invalidation to work.

If a submission to an exercise is made, that exercise needs to be invalidated for every deadline in that course (since it's possible that one exercise occurs twice in a course, in separate series with their own deadline). This requires a partial match (`Rails.cache.delete_matched`), which is not supported in MemCache.
image

thepieterdc:
Invalidations have been fixed using loops instead of partial matches, this does not impose a problem since the amount of iterations is very small.

app/models/series.rb
```diff
@@ -106,4 +121,16 @@ def regenerate_exercise_tokens
...
     end
   end
+
+  def invalidate_caches
+    # Deadline might have been changed -> Invalidate the exercise completion
+    # status of every exercise in this series, for all users. Afterwards,
+    # invalidate the completion status of the series itself for every user.
+    course.users.each do |user|
+      exercises.each do |exercise|
+        exercise.invalidate_accepted_for(course: course, deadline: deadline, user: user)
```
charvp, line 131: Does this need to happen? The cache string uses the deadline right? So the old values won't be found anyway. 

Also, if the deadline changed you're using the new value here.

app/models/series.rb
```diff
@@ -21,8 +21,11 @@
 
 class Series < ApplicationRecord
   include ActionView::Helpers::SanitizeHelper
+  include Cacheable
   include Tokenable
 
+  USER_COMPLETED_CACHE_STRING = '/series/%<id>s/user/%<user_id>s'.freeze
```
charvp, line 27: Can't we just use the deadline here as well? In that case I think no invalidation needs to happen on series update, right?

app/models/submission.rb
```diff
@@ -262,6 +262,15 @@ def invalidate_caches
 
     return if course.blank?
 
+    # Invalidate the completion status of this exercise, for every series in
+    # the current course that contains this exercise, for the current user.
+    # Afterwards, invalidate the completion status of the series itself as well.
+    exercise.series.where(course_id: course_id).find_each do |ex_series|
+      exercise.invalidate_accepted_for(course: course, deadline: ex_series, user: user)
```
charvp, line 269:
```suggestion
      exercise.invalidate_accepted_for(course: course, deadline: ex_series.deadline, user: user)
```

app/models/series.rb
thepieterdc: I have included the deadline in the cache string as you suggested in the next comment, which indeed removes the need for this invalidation.

thepieterdc:
@bmesuere directed towards you ^

app/models/post.rb
```diff
@@ -0,0 +1,22 @@
+class Post < ApplicationRecord
```
bmesuere, line 1: to be removed

app/models/series.rb
```diff
@@ -37,6 +43,7 @@ class Series < ApplicationRecord
 
   before_create :generate_access_token
   before_save :regenerate_exercise_tokens, if: :visibility_changed?
+  after_save :invalidate_exercise_statuses
```
bmesuere, line 46: this will always remove the statuses, even if only the description is changed

app/views/series/_series.html.erb
```diff
@@ -40,7 +26,7 @@
           <i class="mdi mdi-dots-vertical"></i>
         </a>
         <ul class="dropdown-menu dropdown-menu-right">
-          <% if current_user&.course_admin?(series.course) %>
```
bmesuere, line 43: here we check for permissions in the dropdown menu, thus this should stay current_user. This is also the case for the other permission checks in this file.

app/models/exercise.rb
```diff
@@ -332,10 +333,13 @@ def last_submission(user, deadline = nil, course = nil)
...
   end
 
-  def accepted_for(user, deadline = nil, course = nil)
-    last_submission(user, deadline, course).try(:accepted)
+  def accepted_for(options)
+    last_submission(options[:user], options[:deadline], options[:course]).try(:accepted)
   end
 
+  invalidateable_instance_cacheable(:accepted_for,
+                                    ->(this, options) { format(USER_ACCEPTED_CACHE_STRING, user_id: options[:user].id.to_s, course_id: options[:course] ? options[:course].id.to_s : 'global', deadline: options[:deadline] ? options[:deadline].to_s : 'global', id: this.id.to_s) })
```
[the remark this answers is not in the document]
thepieterdc: Not used anymore

app/assets/stylesheets/models/series.css.less
```diff
@@ -20,6 +20,10 @@
   text-align: right;
 }
 
+.deadline-future {
```
rien, line 23: Maybe the `.colored-secondary` class can be re-used from #1855, but it should be moved to another CSS file then.

app/models/exercise_status.rb
```diff
@@ -0,0 +1,61 @@
...
+class ExerciseStatus < ApplicationRecord
+  belongs_to :exercise
+  belongs_to :series, optional: true
+  belongs_to :user
+
+  scope :in_series, ->(series) { where(series: series) }
+  scope :for_exercise, ->(exercise) { where(exercise: exercise) }
+  scope :for_user, ->(user) { where(user: user) }
+
+  after_create :refresh_values
```
rien, line 26: I think this will cause two database operations (create & update) quickly after each other. Maybe this can be a `before_create` and the `refresh_values` can then set the values instead of performing an update.

Or is there a specific reason to do it this way?

app/assets/stylesheets/models/series.css.less
thepieterdc: This is possible if that has been merged :)

app/models/exercise_status.rb
thepieterdc: The `refresh_values` method is also invoked from other models, but I can work something out.

bmesuere:
This is also the case if the exercise has a submission. Every refresh 8 additional and identical exercise status entries are created.

thepieterdc:
> There still seems to be a bug in the code: Too many (duplicate) ExerciseStatus entries are created in the database.
> 
> To reproduce: open the show series page containing exercises without submissions for that user. Each refresh will create an/multiple additional ExerciseStatus in the database:
> image
> 
> I'll try to figure out if this is also the case if there is at least one submission for that exercise.

This bug should be fixed, the unique constraint has apparently gone missing after changing from course-based to series-based caching.

thepieterdc:
> The branch is currently deployed on naos and everything looks good. I think future optimizations are possible. For example, it seems that the query for the "last submission" is always being done while this is not needed if we already know the user hasn't attempted that exercise.
> 
> If performance would still be a problem, we could also include the id of the last submission (before the deadline) in the table. This would not require extra queries when storing the ExerciseStatus but could reduce the number of queries later on.

I was not in favour of storing the id of the last submission, since this would not make sense for content pages.

app/models/exercise.rb
```diff
@@ -307,8 +307,10 @@ def users_tried(options)
                                     ->(this, options) { format(USERS_TRIED_CACHE_STRING, course_id: options[:course] ? options[:course].id.to_s : 'global', id: this.id.to_s) })
 
   def exercise_statuses_for(user, course)
-    series_memberships.in_course(course).each do |series_membership|
-      ExerciseStatus.create_or_find_by(exercise: self, series: series_membership.series, user: user)
+    return [exercise_status_for(user, nil)] unless course
```
rien, line 310:
```suggestion
    return [exercise_status_for(user, nil)] if course.nil?
```
More explicit.

PR #1837, opened 2020-04-10
Allow multiple code listings on same page

niknetniko:
The `FeedbackCodeRenderer` uses an id for the copy button; on the exercise information page there are multiple code listings in the same page, which breaks since there can't be multiple elements with the same id.
Additionally, `add_messages` is not called on the exercise information page, so the javascript is not initialised.

What I've done:
- The `FeedbackCodeRenderer` tracks an instance number to make unique ids. If there is a better solution, feel free to apply it.
- I've moved the copy functionality out of the code listing class. The `FeedbackCodeRenderer` will now always load it.
- I've renamed `parse` to `add_code` and added a new `parse` method, to better reflect their actual use (the old `parse` method did more than just parse the code).

### DIFF
--- a/app/helpers/renderers/feedback_code_renderer.rb
+++ b/app/helpers/renderers/feedback_code_renderer.rb
@@ -2,10 +2,38 @@ class FeedbackCodeRenderer
   require 'json'
   include Rails.application.routes.url_helpers
 
+  @instances = 0
+
+  class << self
+    attr_accessor :instances
+  end
+
   def initialize(code, programming_language)
     @code = code
     @programming_language = programming_language
     @builder = Builder::XmlMarkup.new
+    self.class.instances += 1
+    @instance = self.class.instances
+  end
+
+  def add_code
+    @builder.div(class: 'code-listing-container') do
+      parse
+
+      # Not possible to use clipboard_button_for here since the behaviour is different.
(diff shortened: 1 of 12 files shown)


app/helpers/renderers/feedback_code_renderer.rb
```diff
@@ -2,30 +2,50 @@ class FeedbackCodeRenderer
...
 
       # Not possible to use clipboard_button_for here since the behaviour is different.
-      @builder.button(class: 'btn btn-default', id: 'copy-to-clipboard', title: I18n.t('js.code.copy-to-clipboard'), 'data-toggle': 'tooltip', 'data-placement': 'top') do
+      @builder.button(class: 'btn btn-default copy-btn', id: "copy-to-clipboard-#{@instance}", title: I18n.t('js.code.copy-to-clipboard'), 'data-toggle': 'tooltip', 'data-placement': 'top') do
         @builder.i(class: 'mdi mdi-clipboard-text mdi-18') {}
       end
+      @builder.script(type: 'application/javascript') do
+        @builder << <<~HEREDOC
+          $(() => {
+            window.dodona.submissionCode#{@instance} = #{@code.to_json};
```
charvp, line 30: Why is this extra variable created? I can see that it is also used on line 85, but just repeating 
`#{@code.to_json}` would also work right? And that creates less variables in the global namespace. My opinion isn't very strong on this, but I feel like it would be cleaner without this variable.

niknetniko: This was done to prevent the code from being present in its json form twice on the resulting page, which seemed inefficient, but my opinion is also not very strong on this.

PR #1840, opened 2020-04-12
Setting up integration tests via capybara

MrKickkiller:
This pull request adds basic integration tests via capybara

- [x] Tests were added


app/views/layouts/application.html.erb
```diff
@@ -13,7 +13,8 @@
   <meta property="og:url" content="<%= request.original_url %>">
   <meta property="og:title" content="<%= @title || 'Dodona' %>">
   <%= yield :meta_tags %>
-  <meta property="og:image" content="<%= root_url(locale: nil) %>social-image.png">
+  <%= csrf_meta_tags %>
```
charvp, line 16: Already included on line 90.

app/views/layouts/application.html.erb
```diff
@@ -13,7 +13,8 @@
...
+  <meta property="og:image" content="<%= root_url(locale: nil) %>icon.png">
```
charvp, line 17: This line shouldn't be changed?

test/system/generic_system_test.rb
```diff
@@ -0,0 +1,8 @@
+require 'application_system_test_case'
+require 'capybara/rails'
+
+class GenericSystemTest < ApplicationSystemTestCase
```
charvp, line 4: Why this extra intermediate class? Seems like this is something you can do in `test/application_system_test_case.rb`?

test/system/generic_system_test.rb
```diff
@@ -0,0 +1,8 @@
...
+  setup do
+    Rails.application.config.default_host = '127.0.0.1'
```
charvp, line 6: Can't this be set in `config/environments/test.rb`? Can you also check if using `dodona.localhost` and `sandbox.localhost` works? (Like in develop?)

.github/workflows/capybara.yml
```diff
@@ -0,0 +1,89 @@
+name: Capybara
```
charvp, line 1: Maybe name this differently? If we ever switch away from capybara to do this type of tests I feel like the CI job name shouldn't need to change.

test/system/generic_system_test.rb
MrKickkiller: When setting the default_host to `dodona.localhost`, CSP blocks the connection between the JS and API
image

Unless I'm mistaken I don't think adding the `http://dodona.localhost` to `config/initializers/content_security_policy.rb` is a possible solution for this, since the port number is random. While setting the default_host to 127.0.0.1 is working as the self in the content_security_policy is able to figure out what port it is running on.

rien: The `sandbox_url` and `default_url`  should use the correct port dynamically: https://github.com/dodona-edu/dodona/blob/32b645590ef5d80fcac69ed5d6086a7005ce11ef/app/controllers/application_controller.rb#L70

So I'm not sure what is going wrong here?

test/system/annotations_test.rb
```diff
@@ -0,0 +1,235 @@
...
+    sign_in @zeus
+    @code_lines = Faker::Lorem.sentences(number: 5)
+    @instance = create :correct_submission, result: File.read(Rails.root.join('db/results/python-result.json')), code: @code_lines.join("\n")
+    @instance.exercise.judge.renderer = PythiaRenderer
+    @instance.exercise.judge.save
+  end
+
+  test 'Can view submission page' do
+    visit(submission_path(id: @instance.id))
+    assert_text 'Correct'
```
bmesuere, line 22: these asserts are indeed very broad. See if you can narrow them down. The capybara docs have examples like `expect(find('#navigation')).to have_button('Sign out')` so it should be possible

bmesuere:
I opened a separate issue for the remaining remarks but will merge this to get this code in develop.

PR #1843, opened 2020-04-13
Content pages

### LINKED ISSUE #1387: Add content pages
Create the concept of repository-backed content pages. In its most basic form, this would be something like an exercise, but without the possibility to hand in.

This would be the first step in becoming a learning platform instead of just an exercise platform. 

There are already a few requests to also use notebooks for this. While it is relatively easy to just render those, making them interactive will require a backend.

Internal changes
- I suggest adding a "type" field to the config file. If this value is not present (i.e. all current exercises), it defaults to "exercise". The only other option right now is "content".
- Create a new "Content" model and extend Exercise from that. Move all common code from Exercise to Content. Alternatively, Content and Exercise share a common super class (name to be discussed)
- We must track the user status of the content page (within a course). No such thing currently exists for Exercise (it entirely depends on the submission status). A new model must probably be added for this. While at it, we could also use this same model to track the user status for exercises (instead of putting it in the cache which is another open issue).

UI changes
- Content type must be indicated in the exercise indexes
- We must be able to search for the content type (when adding exercises)
- Content type must be indicated in an exercise series
- On the content show page, don't show the option to hand in and the top left "hand in" link, instead add a "mark as read" UI below the exercise. The "submissions" navbar link should not be present
- The info page should not show solutions and number of submissions

thepieterdc:
This PR adds support for content pages, which are essentially the same as exercises, but without the possibility to submit.

**Design decisions:**
- Single-table inheritance was used due to the lack of big differences between `ContentPage` and `Exercise`.
- A base model `Activity` was added for convenience, and to allow `ContentPage` to evolve separately from `Exercise`.
- The default type was set to `Exercise` to allow backwards compatibility with existing exercises.
- A `type`-field should be added to the configuration file, indicating it is either a
  - `ContentPage` (type: `content`)
  - `Exercise` (type: `exercise` - default)

**Subtasks:**
- [x] Implement migrations and models (#1893)
- [x] Fix `rails db:reset` (#1898)
- [x] Mark a content page as read using a button (creates a new `ActivityReadState`) (#1907)
- [x] Implement users correct etc for content pages (#1905)
- [x] Refactor the exercises controller to activities controller, but make sure the old routes keep working (#1897)
- [x] Refactor the exercise table view (#1896)
- [x] Fix series edit (#1903)
- [x] Check scoresheet (#1910 )
- [x] Check indianio export (#1904)
- [x] Check export (#1904)
- [x] Refactor the `exercises/show` view to not show a submission box if the activity is not an exercise
- [x] `activities/edit` is erroring (#1909)
- [ ] Clean-up models in which `activities`, `content_pages` and `exercises` are all associated, to only use `activities`, if relevant (e.g. not in the `Submission` model)
- [ ] Scan for usages of e.g. `best_submission` or other exercise specific methods all over the `app/` folder to ensure nothing is missed.
- [x] Add progress bar for content pages to course page (#1906)
- [x] Add status for content pages to course page (#1913)
- [ ] Ensure that the json views are still working and include the type of the activity.
- [ ] Possibility to change the type of an activity by clicking the edit-pencil.
- [x] Possibility of searching by activity type (#1900)
- [x] proper activity factory for tests (#1912)
- [x] on an activity page, the right hand navigation is broken (no active marker) (#1914)
- [x] add tests for read state  (#1912)
- [x] update db:reset to generate some read states for content pages (#1912)
- [ ] split translations into an exercise and reading activity version by using 2 subkeys and using string interpolation on only the subkey

**Test edge cases:**
- [x] Changing the type of an exercise to a content page and back should not remove associated submissions. (#1893)
- [x] It should not be possible to submit to a content page ( #1912).

Fixes #1387


bmesuere:
@thepieterdc kan je clean up the subtasks in the PR?

app/views/activities/_series_activities_table.html.erb
```diff
@@ -0,0 +1,72 @@
...
+            <i class="mdi mdi-reorder-horizontal mdi-18"></i>
+          </div>
+        <% end %>
+      </td>
+      <td class="status-icon">
+        <%= raw "<i class='mdi mdi-eye-off mdi-18' title='#{t("activities.index.private")}'></i>" if activity.access_private? %>
+        <%= raw "<i class='mdi mdi-delete mdi-18' title='#{t("activities.index.removed")}'></i>" if activity.removed? %>
+        <%= raw "<i class='mdi mdi-alert-circle mdi-18' title='#{t("activities.index.invalid")}'></i>" if activity.not_valid? %>
+
+        <%= raw "<i class='mdi mdi-file-document-outline mdi-18' title='#{t("activities.index.type.exercise")}'></i>" if activity.exercise? %>
```
bmesuere, line 24:
```suggestion
        <%= raw "<i class='mdi mdi-file-document-edit-outline mdi-18' title='#{t("activities.index.type.exercise")}'></i>" if activity.exercise? %>
```

PR #1853, opened 2020-04-15
Fix intra page links

### LINKED ISSUE #1810: Support intra page links in exercise descriptions
Links to another section within the same description page don't work as expected since the introduction of the iframes. Since there is no reference to the outer page, the links lead to a page only showing the iframe content.

```
<a href="#id-locator">xxx</a>
```

You can see it in action by following the footnote links in [this description](https://dodona.ugent.be/nl/courses/293/series/2535/exercises/1562705175/).

niknetniko:
This pull request enables intra page links and applies some CSS to footnotes.

image

Enabling the links to work was just an option in the iframe-resizer library, but I've also added some code to account for the permanent navbar: when going back from a footnote, it would cause the footnote mark to be under the navbar.

Closes #1810.

### DIFF
--- a/app/helpers/exercise_helper.rb
+++ b/app/helpers/exercise_helper.rb
@@ -73,7 +73,7 @@ module ExerciseHelper
           onResized: dodona.afterResize,
           onMessage: dodona.onFrameMessage,
           onScroll: dodona.onFrameScroll,
-          inPageLinks:true,
+          inPageLinks: true,
         },
         '##{id}')
     }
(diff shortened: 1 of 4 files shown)


app/helpers/exercise_helper.rb
```diff
@@ -72,6 +72,8 @@ def description_iframe(exercise)
           heightCalculationMethod: 'bodyScroll',
           onResized: dodona.afterResize,
           onMessage: dodona.onFrameMessage,
+          onScroll: dodona.onFrameScroll,
+          inPageLinks:true,
```
charvp, line 76:
```suggestion
          inPageLinks: true,
```

PR #1857, opened 2020-04-17
Fix empty actions menu at searchbar

thepieterdc:
This PR fixes an odd visual bug, where the actions menu next to a sidebar shows up empty if there are no actions.

image

### DIFF
--- a/app/views/layouts/_searchbar.html.erb
+++ b/app/views/layouts/_searchbar.html.erb
@@ -12,13 +12,15 @@
         <div class="auxbar bar bar3" style="width: 0%;"></div>
       </div>
     </form>
-    <div class="btn-group hidden actions" id="kebab-menu">
-      <a class="btn btn-icon dropdown-toggle" data-toggle="dropdown">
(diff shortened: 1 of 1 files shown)


app/views/layouts/_searchbar.html.erb
```diff
@@ -12,13 +12,15 @@
...
       </div>
     </form>
-    <div class="btn-group hidden actions" id="kebab-menu">
-      <a class="btn btn-icon dropdown-toggle" data-toggle="dropdown">
-        <i class="mdi mdi-dots-vertical"></i>
-      </a>
-      <ul class="dropdown-menu dropdown-menu-right">
-      </ul>
-    </div>
+    <% unless local_assigns[:actions].empty? %>
```
bmesuere, line 15:
```suggestion
    <% if local_assigns[:actions].any? %>
```
Prefer `if` over `unless` if it's possible. I don't know the type of the value, but I think `any?` or `present?` can be used?

PR #1875, opened 2020-04-21
Fix copy button location

ninewise:
It sticks to the top of the listing now.

Before:
before

After:
after

### DIFF
--- a/app/assets/stylesheets/components/code_listing.css.less
+++ b/app/assets/stylesheets/components/code_listing.css.less
@@ -8,6 +8,7 @@
 
     .copy-btn {
       margin-right: 5px;
+      margin-top: 5px;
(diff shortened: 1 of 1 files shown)


thepieterdc:
Maybe the copy button requires additional fixing: (exercise: https://dodona.ugent.be/nl/exercises/1807370355/)

image

charvp:
> Maybe the copy button requires additional fixing: (exercise: https://dodona.ugent.be/nl/exercises/1807370355/)
> 
> Even though this will be replaced by a content page in the future, it might be applicable to empty solutions in general.
> 
> Suggestion: don't display the copy button if there is nothing to copy?
> 
> image

Seems like an issue for a different PR to me.

PR #1876, opened 2020-04-21
Fix attempting to display non existent deadlines

thepieterdc:
This pull request patches exercise status showing deadlines that do not exist.

### DIFF
--- a/app/views/exercises/_user_status.html.erb
+++ b/app/views/exercises/_user_status.html.erb
@@ -6,7 +6,7 @@
     <%= link_to submission, class: 'deadline-ok' do %>
       <%= t 'exercises.exercise.correct' %>
     <% end %>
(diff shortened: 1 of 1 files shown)


app/views/exercises/_user_status.html.erb
```diff
@@ -6,7 +6,7 @@
     <%= link_to submission, class: 'deadline-ok' do %>
       <%= t 'exercises.exercise.correct' %>
     <% end %>
-    <% unless exercise.accepted_before_deadline_for?(user, series) %>
+    <% unless series&.deadline? && exercise.accepted_before_deadline_for?(user, series) %>
```
charvp, line 9: Is this correct? It's an `unless`, so this will always be triggered if there is no deadline.

thepieterdc: Replaced

PR #1885, opened 2020-04-22
Also see SIGABRT as a judge error that shouldn't result in an internal error

charvp:
The JS judge exits with SIGABRT in the case of an out of memory error. This should thus fix the recent internal errors due to out of memory errors in the JavaScript judge (the internal errors we get since https://github.com/dodona-edu/judge-javascript/pull/56 are due to high memory usage in the student code). The current signals we whitelist are SIGKILL and SIGTERM, so SIGABRT seems reasonable to me as well.

I also propose that we hotfix this, due to this happening quite often recently.

### DIFF
--- a/app/runners/submission_runner.rb
+++ b/app/runners/submission_runner.rb
@@ -179,7 +179,7 @@ class SubmissionRunner
       ]
     end
 
-    if [0, 137, 143].exclude? exit_status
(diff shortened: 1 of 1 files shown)


pdawyndt:
@charvp are SIGABRTs now reported as memory errors or general judge errors?

charvp:
Memory errors.

PR #1893, opened 2020-04-23
Activities core

thepieterdc:
This PR adds the core of activities for the content pages.

- [x] Database migrations
- [x] Fix the models
   - [x] Activity model
   - [x] Content model
   - [x] Exercise model
   - [x] ExerciseLabel model
   - [x] ExerciseStatus model
- [x] Fix the seeds

**Tests**
- [x] Changing the type of an exercise to a content page and back should retain all submissions.
- [x] Removing the ActivityStatus of a Content-activity should not be possible.


thepieterdc:
This should be done, apart from the test case that ensures activity statuses cannot be removed for content pages (as there is no way to recover them, since submissions are not a thing).

I was not sure how to cope with this, as ActivityStatus is currently used exclusively as a cache (therefore allowing removal), but this behaviour should be changed.

charvp:
> This should be done, apart from the test case that ensures activity statuses cannot be removed for content pages (as there is no way to recover them, since submissions are not a thing).
> 
> I was not sure how to cope with this, as ActivityStatus is currently used exclusively as a cache (therefore allowing removal), but this behaviour should be changed.

ActivityStatus records are only removed on series deadline change right? Can't we just call `update_values` there? (Maybe in a delayed job though.)

rien:
Isn't the problem that the ActivityStatus is used for two things?
- As an ephemeral cache item to track what the current status is for a user - series - exercise combination
- and as a permanent record to store whether a user has seen a content page.

Maybe it can be beneficial to split this into the two separate things: the ephemeral cache object and the permanent record (which is kind of like a solution, and maybe we could generalize submission here as well).

app/models/activity_status.rb
```diff
@@ -34,17 +33,17 @@ def wrong?
...
 
   private
 
-  def initialise_values
-    best = exercise.best_submission(user, nil, series&.course)
-    best_before_deadline = exercise.best_submission(user, series&.deadline, series&.course)
-    last = exercise.last_submission(user, nil, series&.course)
-    last_before_deadline = exercise.last_submission(user, series&.deadline, series&.course)
+  def initialise_values_for_exercise
+    best = activity.best_submission(user, nil, series&.course)
```
bmesuere, line 43: maybe add a return unless activity.exercise? as first line to make clear that activity is an exercise here?

app/models/activity_read_state.rb
```diff
@@ -0,0 +1,49 @@
...
+#  solved                   :boolean          default(FALSE), not null
+#  started                  :boolean          default(FALSE), not null
+#  solved_at                :datetime
+#  exercise_id              :integer          not null
+#  series_id                :integer
+#  user_id                  :integer          not null
+#  created_at               :datetime         not null
+#  updated_at               :datetime         not null
+#
+class ActivityReadState < ApplicationRecord
```
bmesuere, line 17: Hmm I'm confused. Is this for exercises, content pages or activities?

app/models/content_page.rb
```diff
@@ -0,0 +1,34 @@
...
+#  access                  :integer          default("public"), not null
+#  programming_language_id :bigint
+#  search                  :string(4096)
+#  access_token            :string(16)       not null
+#  repository_token        :string(64)       not null
+#  allow_unsafe            :boolean          default(FALSE), not null
+#  type                    :string(255)      default("Exercises"), not null
+#
+
+class ContentPage < Activity
```
bmesuere, line 24: I don't have a lot of experience with STI, but is it common practice to override stuff like "judge" in here to throw an error?

app/models/activity_read_state.rb
thepieterdc: I have used the name `Activity`ReadState because it simply keeps track of whether something was read or not. This is not content page specific and can therefore be used for other types of activities as well, should they ever be added in the future.

app/controllers/series_controller.rb
```diff
@@ -157,18 +157,18 @@ def add_exercise
         return
       end
     end
-    SeriesMembership.create(series: @series, exercise: @exercise)
```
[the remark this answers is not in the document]
thepieterdc: Controllers are for later PR's since this PR does not break them and this PR focusses on the core architecture.

app/models/activity.rb
```diff
@@ -0,0 +1,411 @@
+# == Schema Information
+#
+# Table name: exercises
```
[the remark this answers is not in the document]
thepieterdc: Done

app/models/content_page.rb
thepieterdc: The judge is not associated in the activity, but rather in the exercise itself. This should therefore be fine.

PR #1899, opened 2020-04-25
Nullify user email

rien:
This pull request fixes a bug where a user's `update_from_oauth` would set their email to the empty string (instead of `nil`).   PR #1793 made a `index_users_on_email` which required uniqueness, throwing an error when a smartschool user without email would log in. A test was added to reproduce this behavior.

This bug is fixed by adding a `nullify_empty_email` callback before a record is saved.

When this is merged, I suggest clearing users with an empty string email by running `User.where(email: '').update(email: nil)`.


charvp:
Do we want to hotfix this?

rien:
> Do we want to hotfix this?

I think this is preventing smartschool users to log in, so I would say yes.

PR #1903, opened 2020-04-25
Fix series edit

bmesuere:
This pull request fixes the series edit page.


app/views/activities/_repository_status.html.erb
```diff
@@ -1,8 +1,8 @@
 <% if activity.access_private? && local_assigns[:series].nil? %>
-  <i class="mdi mdi-eye-off mdi-18" title="<%= t('activitys.index.private') %>"></i>
+  <i class="mdi mdi-eye-off mdi-18" title="<%= t('activities.index.private') %>"></i>
```
charvp, line 2: This is already fixed in `feature/content-pages`, but if there are no merge conflicts 🤷‍♀️

PR #1904, opened 2020-04-25
Fix submission export

charvp:
Not much to fix here, IMO. Only some translation keys that weren't updated.

Export selection of a series that looks like this:
image
Looks like this:
image

Which I would think is the correct behaviour. Export itself uses exercises as well, and only uses submissions.


charvp:
Also checked Indianio export, that still works as expected (only exercises are used).

bmesuere:
don't merge this till it's rebased with the series stuff

PR #1907, opened 2020-04-26
Allow marking content page as read

thepieterdc:
This pull request adds a `Mark as read`-button to the bottom of a content page, as well as navigation buttons.

**Tasks:**
- [x] Implement mark-as-read controller action
- [x] Add button
- [x] Button is not centered
- [x] Chevron at "Next activity" is probably better off placed on the right
- [x] Add some visual feedback to indicate that the activity was successfully marked as read
- [x] Hide the button if the exercise was already read

**Consideration:**
If we were to implement "un-read", this would have to be implemented by destroying the ActivityReadState. However, this would imply that a user can miss a deadline after passing it first, or the other way around, since the ActivityStatus is just an ephemeral cache.

### DIFF
--- a/app/controllers/activities_controller.rb
+++ b/app/controllers/activities_controller.rb
@@ -123,7 +123,10 @@ class ActivitiesController < ApplicationController
                                        user: current_user
     can_read = @activity.read_state_for(current_user, @course).blank?
     if can_read && read_state.save
-      render json: { status: 'ok', activity_id: read_state.activity_id, course_id: read_state.course_id }
+      respond_to do |format|
+        format.js { render 'activities/read', locals: { activity: @activity, course: @course, user: current_user } }
+        format.json { head :ok }
+      end
     else
       read_state.errors.add(:activity, 'already read') unless can_read
       render json: { status: 'failed', errors: read_state.errors }, status: :unprocessable_entity
--- a/app/controllers/activities_controller.rb
+++ b/app/controllers/activities_controller.rb
@@ -122,6 +122,7 @@ class ActivitiesController < ApplicationController
                                        course: @course,
                                        user: current_user
     can_read = @activity.read_state_for(current_user, @course).blank?
+    can_read &&= @activity.accessible?(current_user, course)
     if can_read && read_state.save
       respond_to do |format|
         format.js { render 'activities/read', locals: { activity: @activity, course: @course, user: current_user } }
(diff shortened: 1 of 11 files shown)


rien:
I would maybe change de `read` action by a `change_read_status` which would `find_or_create` the status and update it accordingly? So that the same action can be used to mark as unread.

app/controllers/activities_controller.rb
```diff
@@ -117,6 +117,19 @@ def edit
     @labels = Label.all
   end
 
+  def read
+    read_state = ActivityReadState.new activity: @activity,
+                                       course: @course,
+                                       user: current_user
+    can_read = @activity.read_state_for(current_user, @course).blank?
```
charvp, line 124: Submissions#create does some extra checks that a user is allowed to submit to an exercise (https://github.com/dodona-edu/dodona/blob/develop/app/controllers/submissions_controller.rb#L66). Something similar should probably be done here.

thepieterdc: This is not necessary here. The submission controller needs to check that since the exercise is passed in the body (the request is sent to `POST /submissions`). However since this is done in the activity controller, this is already checked when `set_activity` is invoked.

charvp: Not true. `set_activity` only does the `authorize` check. We need to do an extra check because the access rights to an activity are dependent on the context they are loaded in. See line 70 for example. (Note that this can't just be moved to `set_activity`, because `media` needs to have a slightly different authentication flow.)

thepieterdc:
Everything has been fixed except for this visual bug:

image

pdawyndt:
Why does the navigation say "previous/next exercise", not "previous/next activity"?

thepieterdc:
> Why does the navigation say "previous/next exercise", not "previous/next activity"?

The translation string has not been updated yet apparently, thanks for spotting

bmesuere:
> Why does the navigation say "previous/next exercise", not "previous/next activity"?

We noticed that using activity everywhere is not a great experience, since almost all "activities" are exercises which complicates the application for the user. We will try to make all translations as specific as possible. For example, these buttons will read next exercise if the next activity is an exercise and next reading activity if it's a content page.

bmesuere:
Button alignment is fixed, previous and next button text is not (later pull request).

PR #1909, opened 2020-04-26
Fix the activities edit page

bmesuere:
This pull request fixes the activities edit page.

The JS and edit button itself are still broken, but waiting for a rebase on the work of @thepieterdc on the activity js bundle to fix that.


bmesuere:
This is without the type edit in the form. My brain has melted down for now.

PR #1913, opened 2020-04-26
Show content page status in activities table

charvp:


charvp:
image

config/locales/views/activities/en.yml
```diff
@@ -9,6 +9,9 @@ en:
       count_tooltip: Number of students that solved this exercise correctly
       read_tooltip: Number of students that read this
       read_by: Read %{count} times
+      read: Read
+      not_read: Not read yet
```
bmesuere, line 13:
```suggestion
      not_read: Not yet read
```

PR #1922, opened 2020-04-26
Fix content_page read within and without course

rien:
- Adds controller tests for showing content pages and marking them as read
- Fix `activity#read` raising an error when course was `nil`
- Added redirect to activity path for format html
- Moved accessibility check to read state validation

### DIFF
--- a/app/controllers/activities_controller.rb
+++ b/app/controllers/activities_controller.rb
@@ -3,7 +3,7 @@ class ActivitiesController < ApplicationController
 
   before_action :set_activity, only: %i[show description edit update media info read]
   before_action :set_course, only: %i[show edit update media info read]
-  before_action :set_series, only: %i[show edit update info]
+  before_action :set_series, only: %i[show edit update info read]
   before_action :ensure_trailing_slash, only: :show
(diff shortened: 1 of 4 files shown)


app/controllers/activities_controller.rb
```diff
@@ -122,19 +122,17 @@ def edit
   end
 
   def read
-    @course = nil unless @course.subscribed_members.include?(current_user)
+    @course = nil if @course.blank? || !@course.subscribed_members.include?(current_user)
     read_state = ActivityReadState.new activity: @activity,
                                        course: @course,
                                        user: current_user
-    can_read = @activity.read_state_for(current_user, @course).blank?
-    can_read &&= @activity.accessible?(current_user, @course)
```
charvp, line 130: Not sure about moving this to a validation. The accessibility can change after being read, which isn't necessarily wrong (e.g. a course admin making the containing series closed). That shouldn't make the model invalid, IMO.

rien: You're right. I think there is an option to only perform this validation on new records.

PR #1925, opened 2020-04-27
Fix viewing activity removed from repository

niknetniko:
It is assumed that full_path will always return a Pathname, but it returned a string when there is no path.

- [x] Tests were added

### DIFF
--- a/app/models/activity.rb
+++ b/app/models/activity.rb
@@ -85,9 +85,7 @@ class Activity < ApplicationRecord
   end
 
   def full_path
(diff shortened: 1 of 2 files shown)


app/models/activity.rb
```diff
@@ -85,9 +85,7 @@ def exercise?
   end
 
   def full_path
-    return '' unless path
```
charvp, line 88: I think I would prefer a `return Pathname.new '' unless path` here instead of the inline condition.

PR #1931, opened 2020-04-27
Make sure homepage can be loaded when user has submitted to content page

charvp:
- [x] Tests were added.


thepieterdc:
Who would do such a thing 🙄

PR #1945, opened 2020-04-28
Performance optimizations of the course show page

### LINKED ISSUE #1892: Improve course page performance
The median load time of a course page is 1 second. This is more than expected, so it is worth to investigate this.

image

The rise of the 95th percentile is due to the (empty) new cache.

bmesuere:
This pull request should  significantly improve the performance of the course show page. This is done in multiple ways:

* The submission id's of relevant submissions are now also stored in the ActivityStatus model. They were calculated anyway when initializing the ActivityStatus, so there is no overhead there. This prevents several expensive queries when rendering a series card where we link to the last submission.
* The ActivityStatus objects that are used on the course page were loaded one by one. The code was modified to first hit a local store to check if we already retrieved the status before hitting the database. In addition, a method was added to preload activity statuses in batch (for all activities in a series) which reduces the number of queries.
* The batch loading described above was expanded with optional eager loading for submissions which further reduces the number of queries.
* The number of activities in a series is now stored in the database. We use a built-in rails mechanism for this so rails handles all the invalidation. This eliminates a query for each of the skeleton series loaded.

Closes #1892

### DIFF
--- a/app/models/activity.rb
+++ b/app/models/activity.rb
@@ -368,15 +368,20 @@ class Activity < ApplicationRecord
   private
 
   def activity_status_for(user, series = nil)
-    attempts = 0
     Current.status_store ||= {}
-    Current.status_store[[user.id, series.id, id]] ||= begin
-      ActivityStatus.find_or_create_by(activity: self, series: series, user: user)
-                                                       rescue StandardError
-                                                         # https://github.com/dodona-edu/dodona/issues/1877
-                                                         raise unless (attempts += 1) <= 1
+    Current.status_store[[user.id, series&.id, id]] ||= activity_status_for!(user, series)
+  end
 
-                                                         retry
+  def activity_status_for!(user, series = nil)
+    first_try = true
+    begin
+      ActivityStatus.find_or_create_by(activity: self, series: series, user: user)
+    rescue StandardError
+      # https://github.com/dodona-edu/dodona/issues/1877
+      raise unless first_try
+
+      first_try = false
+      retry
     end
   end
 
--- a/app/models/activity.rb
+++ b/app/models/activity.rb
@@ -369,13 +369,14 @@ class Activity < ApplicationRecord
 
   def activity_status_for(user, series = nil)
     attempts = 0
-    begin
+    Current.status_store ||= {}
+    Current.status_store[[user.id, series.id, id]] ||= begin
       ActivityStatus.find_or_create_by(activity: self, series: series, user: user)
-    rescue StandardError
-      # https://github.com/dodona-edu/dodona/issues/1877
-      raise unless (attempts += 1) <= 1
+                                                       rescue StandardError
+                                                         # https://github.com/dodona-edu/dodona/issues/1877
+                                                         raise unless (attempts += 1) <= 1
 
-      retry
+                                                         retry
     end
   end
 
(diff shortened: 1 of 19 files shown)


app/models/activity.rb
```diff
@@ -369,13 +369,14 @@ def types
 
   def activity_status_for(user, series = nil)
     attempts = 0
-    begin
+    Current.status_store ||= {}
+    Current.status_store[[user.id, series.id, id]] ||= begin
```
charvp, line 373: I think `series` can be `nil` here?

app/models/activity_status.rb
```diff
@@ -39,6 +50,13 @@ def update_values
     save
   end
 
+  def self.add_status_for(user, series, eager = [])
+    Current.status_store ||= {}
+    ActivityStatus.where(series: series, user: user).includes(eager).find_each do |as|
```
charvp, line 55: `find_each` seems unnecessary here, you need to load all the objects into memory anyway (and they won't be garbage-collected).

bmesuere: I think it's a best practice to always use find_each. As far as I know it has no drawbacks over find and might prevent future issues.

app/models/activity.rb
bmesuere: Yes, this was fixed by using`series&.id`

bmesuere:
Measuring the before state:
- /courses/175/ as myself: 670-700ms, 100 queries, 30% is sql
- /courses/175/ as BD: 1100-1300ms, 98 queries, 22% in sql
- /courses/175/ as CA: 800-900ms, 118 queries, 22% in sql
- /courses/293/ as myself: 550-700ms, 75 queries, 20% in sql
- /courses/293/ as PD: 750-900ms, 77 queries, 11% in sql
- /series/1709/ as myself: 250-300ms, 60 queries, 17% in sql
- /series/1709/ as BD: 250-300ms, 58 queries, 19% in sql
- /series/1709/ as CA: 270-350ms, 61 queries, 22% in sql
- / as myself: 300-350ms, 53 queries, 50% in sql
- / as PD: 1000-1200ms, 123 queries, 45% in sql

bmesuere:
> Any idea how much time we've shaved off with these optimizations?

I will let you know once this is deployed in production ;) Based on testing on naos, there will be a 60% reduction in number of queries.

bmesuere:
- /courses/175/ as myself: 670-700ms, 100 queries, 30% is SQL -> 450ms, 34q, 34%
- /courses/175/ as BD: 1100-1300ms, 98 queries, 22% in SQL -> 800-850ms, 34q, 17%
- /courses/175/ as CA: 800-900ms, 118 queries, 22% in SQL -> 500-550ms, 37q, 14%
- /courses/293/ as myself: 550-700ms, 75 queries, 20% in SQL -> 300ms, 33q, 17%
- /courses/293/ as PD: 750-900ms, 77 queries, 11% in SQL -> 600-650ms, 35q, 9%
- /series/1709/ as myself: 250-300ms, 60 queries, 17% in SQL -> 130-180ms, 24q, 11%
- /series/1709/ as BD: 250-300ms, 58 queries, 19% in SQL -> 160-240ms, 24q, 11%
- /series/1709/ as CA: 270-350ms, 61 queries, 22% in SQL -> 200-250ms, 27q, 10%
- / as myself: 300-350ms, 53 queries, 50% in SQL -> 280-350ms, 56q, 50%
- / as PD: 1000-1200ms, 123 queries, 45% in SQL -> 880-930ms, 128q, 40%

PR #1946, opened 2020-04-28
Fix performance regression on course page load

bmesuere:
This pull request fixes a performance regression when loading a course page.
The deadline was formatted for every series, even if the series itself was not loaded.

Formatting of the deadline needs the series status, which can be cached. If it is not cached, the status for all exercises must be retrieved. If the activity status is not in the database, this causes a lot of submission loads.


thepieterdc:
Actually, performance can be further (albeit minor) increased if the partial is only rendered if loaded is true.

PR #1950, opened 2020-04-29
Submission index: periodically reload to watch for new submissions

### LINKED ISSUE #1874: Option to reload submissions page periodically
This would be useful when waiting for new submissions to come in, so we don't have to refresh ourselves.

Since this page already uses the index table to load submissions on search, this is simple to implement by calling the search function periodically when the user has requested to do so.

I propose a reload time of 15 seconds.

rien:
This pull request

- [ ] Tests were added

Closes #1874.

### DIFF
--- a/app/controllers/submissions_controller.rb
+++ b/app/controllers/submissions_controller.rb
@@ -24,29 +24,24 @@ class SubmissionsController < ApplicationController
     @title = I18n.t('submissions.index.title')
 
     return unless stale? @submissions
+    return unless format.html?
 
-    respond_to do |format|
-      format.html do
-        @crumbs = []
-        if @user
-          @crumbs << if @course.present?
-                       [@user.full_name, course_member_path(@course, @user)]
-                     else
-                       [@user.full_name, user_path(@user)]
-                     end
-        elsif @series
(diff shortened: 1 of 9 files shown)


app/controllers/submissions_controller.rb
```diff
@@ -22,22 +22,31 @@ def index
...
-    elsif @series
-      @crumbs << [@series.course.name, course_path(@series.course)] << [@series.name, @series.hidden? ? series_path(@series) : course_path(@series.course, anchor: @series.anchor)]
-    elsif @course
-      @crumbs << [@course.name, course_path(@course)]
-    elsif @judge
-      @crumbs << [@judge.name, judge_path(@judge)]
+
+    return unless stale? @submissions
+
+    respond_to do |format|
```
bmesuere, line 28: Is there a reason why you use this (in my opinion very complex double block) construction over simply wrapping it with a `request.format.html?` check?

rien: The reason is probably because `respond_to` is most used for this. But I agree, `request.format.html?` is much simpler indeed.

niknetniko:
Superseded by #2310.

Issue #1954, opened 2020-04-30
General overhead on Dodona pages


bmesuere:
Loading an exercise should only require minimal calculations. This is a good candidate page to look for general overhead that is present on all pages. Loading this pages takes 100-200ms so it's easy to calculate the time corresponding to the percentages.
![image](https://user-images.githubusercontent.com/481872/80700660-64fe9480-8ade-11ea-9e58-43981adf8ebe.png)

Note that the "gaps" are not included in the % time calculation. During these gaps, the server is waiting for the results of a query. For example, the 8% of setLocale is a lot higher in reality because it requires several queries.

bmesuere:
A first issue is the time spent in `set_locale`. We here calculate the locale out of the default, de explicit locale set in the URL and the value set in the user profile. After we determine which locale to use, the values is stored in the database.

Since this hardly changes, there is almost never a need to actually update the value. Rails is smart enough to not perform an update in that case, but it does perform all validations on the user model. Because this includes a few uniqueness checks, there are a few unnecessary queries.

By only performing the update if the value has changed, we can shave 20+ms and 5 queries from EVERY pageload. With 100K pageloads a day, this is definitely worth it.

PR #1958, opened 2020-04-30
optimize drawer partial

bmesuere:
This pull request hopefully improves the drawer partial performance. Right now on an exercise page on my account, it takes 12ms, 3 queries and 3ms in SQL to render the partial.

#1954 

In a later PR when cleaning up the home page, the grouped functions in the user model can be removed.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -237,9 +237,9 @@ class User < ApplicationRecord
 
   def drawer_courses
     actual_memberships = course_memberships.includes(:course).to_a.select(&:subscribed?)
-    favorites = actual_memberships.select(&:favorite)
-
     return [] if actual_memberships.empty?
+
+    favorites = actual_memberships.select(&:favorite)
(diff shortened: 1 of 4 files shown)


app/models/user.rb
```diff
@@ -235,6 +235,17 @@ def grouped_recent_courses(number_of_years)
     subscribed_courses.group_by(&:year).first(number_of_years)
   end
 
+  def drawer_courses
+    actual_memberships = course_memberships.includes(:course).to_a.select(&:subscribed?)
+    favorites = actual_memberships.select(&:favorite)
+
+    return [] if actual_memberships.empty?
```
charvp, line 242:
```suggestion
    actual_memberships = course_memberships.includes(:course).to_a.select(&:subscribed?)
    return [] if actual_memberships.empty?

    favorites = actual_memberships.select(&:favorite)
```
For some reason, this looks a little better to me.

PR #1964, opened 2020-05-03
Use fragment cache for markdown descriptions

bmesuere:
This pull request enables the rails fragment cache for series and course descriptions. On the course page, 30% of the time is spent on markdown rendering and sanitization. The fragment cache auto-generates a cache key which changes if the object itself (`updated_at` value) or the code within the cache block changes so we don't have to worry about invalidation.

The skeleton series could also be cached as a whole in a future pull request since they don't contain any dynamic content.

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -38,7 +38,7 @@ Rails.application.configure do
 
     config.action_mailer.perform_caching = false
 
-    # config.cache_store = :mem_cache_store, { namespace: :"2" }
+    config.cache_store = :mem_cache_store, { namespace: :"2" }
     config.public_file_server.headers = {
       'Cache-Control' => "public, max-age=#{2.days.to_i}"
     }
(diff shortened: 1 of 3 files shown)


config/environments/development.rb
```diff
@@ -38,7 +38,7 @@
 
     config.action_mailer.perform_caching = false
 
-    config.cache_store = :mem_cache_store, { namespace: :"2" }
+    # config.cache_store = :mem_cache_store, { namespace: :"2" }
```
charvp, line 41: Did you mean to do this?

bmesuere: I don't think we want to run/use memcache for local development?

charvp: The reason I originally homogenised the cache configurations is because not all cache backends support the same set of functions so using something different seemed dangerous to me.

PR #1995, opened 2020-05-14
Add system tests using capybara

charvp:
Split off from #1777 to make the diff there more readable and because the review sessions don't depend on this (and this doesn't depend on the review sessiosns).

### DIFF
--- a/.github/workflows/system-test.yml
+++ b/.github/workflows/system-test.yml
@@ -4,12 +4,10 @@ on: [push]
 
 jobs:
   test:
+    env:
+      RAILS_ENV: "test"
+      TEST_DATABASE_URL: "mysql2://root:dodona@127.0.0.1:3306/dodona_test"
     runs-on: ubuntu-latest
-    container:
-      image: ruby:2.6.6
-      env:
-        RAILS_ENV: "test"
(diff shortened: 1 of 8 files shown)


.github/workflows/system-test.yml
```diff
@@ -0,0 +1,94 @@
...
+    services:
+      mysql:
+        image: mariadb
+        env:
+          MYSQL_DATABASE: "dodona_test"
+          MYSQL_ROOT_PASSWORD: "dodona"
+          MYSQL_HOST: "localhost"
+    steps:
+    - uses: actions/checkout@v1
+    - name: Install node and yarn
```
bmesuere, line 22: This probably still uses the old way to install node and yarn: https://help.github.com/en/actions/language-and-framework-guides/using-nodejs-with-github-actions
Yarn should be installed on all runners, node can be installed with an action according to the documentation.

charvp: We had to install this way because we used ruby from a docker image instead of the `setup-ruby` action. I've changed that, so this is now a lot cleaner. I also took the opportunity to go through our actions and clean them up a bit, since some things were outdated/not done in the recommended way.

PR #2000, opened 2020-05-16
Remove datadog RUM

bmesuere:
This pull request removes the Datadog RUM. The results from this data are limited and don't outweigh the additional data gathering.

After this gets deployed, we will gather less data than before since firebase was removed as well. Performance monitoring now only happens using anonymous server-side metrics.


bmesuere:
The reason we added that was to keep an eye on the page load time of Dodona. Now that we track the server-side render times, there's not a huge need to do that anymore. From a performance perspective, something like this is many times more useful:
image

We do lose tracking of the impact of css/js, but firebase was not great for that in the first place. There might be better ways to track this, for example by logging the bundle sizes for every release or running a lighthouse check every week on github actions.

PR #2012, opened 2020-05-20
Improve sanitize performance

bmesuere:
This pull request should in prove the performance of the sanitize functions which is called many many times when constructing a feedback table. I made 2 changes:

- The list of allowed tags and attributes is now constructed only once
- The addition of noopener and noreferrer to target blanks links was removed. For this last check, the entire string had to be parsed a second time. The removal of this opens a small attack surface: course admins and exercise creators could add a link to a malicious site. This site could then change the "parent" page to a look-alike of the sign in page. When the user closes the newly opened tab, he then sees this sign in page and could be tricked into trusting that page because Dodona was previously in that tab. The risk for this seems minimal:
  - Only semi-trusted users can create courses and exercises
  - Dodona has no sign in page, so it would have to target a page from the institution of the user
  - Browsers are working to preventing this exploit. This is already fixed in Safari and the Firefox nightly. A fix is also planned for Chromium.

I will amend this PR with some timings.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -137,12 +137,7 @@ module ApplicationHelper
     sanitized = ActionController::Base.helpers.sanitize html,
                                                         tags: tags,
                                                         attributes: attributes
-    # If an anchor has a target, disable the referer
-    doc = Nokogiri::HTML::DocumentFragment.parse(sanitized)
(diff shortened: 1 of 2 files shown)


bmesuere:
https://dodona.ugent.be/en/submissions/5349671/ takes 2.3-2.5 seconds to load with 36% of the time spent in the `safe` method of the feedback table renderer.

PR #2061, opened 2020-06-15
Allow to expand individual annotations

### LINKED ISSUE #2037: Show specific code-annotation when annotations are minimized
Currently, when code-annotations are minimized, Dodona shows a small dot next to lines that have one. It could be a useful feature to be able to click on these dots to show that specific annotation.

image

thepieterdc:
This pull request allows the user to show individual annotations by clicking on the dots. Currently, the dots do not disappear when the annotation is shown, to allow the user to re-hide it.

**Limitation:** Since the `annotate-this-line` button appears on top of the dot, it is only possible to click the dot as a student.

visual

Closes #2037  .

### DIFF
--- a/app/assets/javascripts/code_listing/code_listing.ts
+++ b/app/assets/javascripts/code_listing/code_listing.ts
@@ -203,6 +203,15 @@ export class CodeListing {
             dot.id = `dot-${line}`;
             row.querySelector<HTMLTableDataCellElement>(".rouge-gutter").prepend(dot);
 
+            // Add a click listener to the dot.
+            dot.addEventListener("click", () => {
+                if (annotation.html.classList.contains("hide")) {
+                    annotation.show();
+                } else {
+                    annotation.hide();
+                }
+            });
+
             // Create annotation groups.
             ANNOTATION_ORDER.forEach((type: string) => {
                 const group = document.createElement("div") as HTMLDivElement;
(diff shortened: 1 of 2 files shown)


bmesuere:
> * The toggle button at the top should probably be put in an indeterminate state
> * This will only work with the first annotation added for a line, as far as I can tell?
> * I do think this should also work for teachers, but I'm not sure how sadly

Regarding that last one: if we will let users ask a question using the same UI, this will be a problem for normal users as well.

pdawyndt:
While reviewing I was paying attention to the issue of overlap between the colored dots as indicators for hidden annotations and the dynamic button for adding a manual annotation to a line. I guess I wouldn't mind if the both were at opposing sides of the editor. Google Docs also puts the dynamic button for adding remarks in the right margin. Since this is the most frequently used option (from my side), we might as well put the dynamic button in the left margin and the colored dots in the right margin.

pdawyndt:
While code reviewing exams, I again got to this issue. I still would move the markers for line annotations to the right margin. Making them clickable there would not interfere with adding manual line annotations using the pop-up button in the left margin.

chvp:
Putting things in the right margin wouldn't work for us IMO. All our lines are outlined to the left and (almost) never take up the full width, leaving it harder to see what indicator matches which line.

pdawyndt:
But the markers in the right margin is not something to be set manually, so I don't think this is an issue.

jorg-vr:
This is no longer relevant with the changes in #4596 
It is now possible to see annotations on hover when they are hidden, which reduces the use case for this feature

PR #2065, opened 2020-06-18
Authentication rework [depends on #2072]

thepieterdc:
This pull request reworks the inner authentication mechanisms to decouple institutions from login providers.

- [x] Providers
  - [x] SAML
  - [x] SmartSchool
  - [x] Google OAuth2
  - [x] Office365
- [x] Identities
- [x] Migrate all existing institutions to providers
  - [x] SAML
  - [x] SmartSchool
  - [x] Google OAuth2
  - [x] Office365
- [x] Creation of new institutions
- [x] Migrate all existing users to identities
- [x] Update fallback behaviour for usernames
- [x] UserController#new
- [x] Tests

Future work in https://github.com/dodona-edu/dodona/issues/2067


db/migrate/20200620113027_extract_usernames_to_identities.rb
```diff
@@ -0,0 +1,11 @@
+class ExtractUsernamesToIdentities < ActiveRecord::Migration[6.0]
+  def change
+    User.find_each do |user|
+      if user.institution.present?
+        Identity.create provider: user.institution.providers.first,
+                        identifier: user.username,
```
bmesuere, line 6: I assume this is just a temp value that will be overwritten by a provider specific identifier when the user first signs in? Or do we already store the uid as username right now?

thepieterdc: > I assume this is just a temp value that will be overwritten by a provider specific identifier when the user first signs in? Or do we already store the uid as username right now?

The uid is actually already stored in the username field.

thepieterdc: https://github.com/dodona-edu/dodona/blob/7e993008927284cead4441402b0bd1c9c779a13e/app/models/user.rb#L261

app/controllers/application_controller.rb
```diff
@@ -46,9 +46,12 @@ def after_sign_out_path_for(_resource)
   end
 
   Warden::Manager.after_authentication do |user, auth, _opts|
+    # SAML authentication.
     if user.institution.nil?
       idp = Institution.find_by(short_name: auth.env['rack.session'][:current_idp])
-      user.update(institution: idp)
+      user.institution = idp
+      user.identities.build(provider: idp.providers.first, identifier: user.username)
```
charvp, line 53: Why is `idp.providers.first` used here? In this phase each institution only has one provider, I guess, but don't we want to find the correct provider using the same data as we use to find the institution here?

charvp, line 53: Right now the user's email address is used to find users. Is switching to `username` something that will work for all (SAML) institutions?

thepieterdc: > Right now the user's email address is used to find users. Is switching to `username` something that will work for all (SAML) institutions?

This is not true: https://github.com/dodona-edu/dodona/blob/da323f58c9f35ecc82969d544a6a367f869a6817/lib/SAML/my_resource_validator.rb#L13

thepieterdc: I have added this to the list of todo's in #2067

PR #1807, opened 2020-04-02
Disable CSRF protection for API requests

thepieterdc:
This pull request disables CSRF protection for API requests, as discussed in #1765. My latest reply in that issue contains the reason as to why the current implementation is used.

- [x] Tests were added.

Closes #1765.

PR #2072, opened 2020-06-22
Rework SAML authentication

thepieterdc:
This pull request replaces the current SAML gem to use omniauth.


app/controllers/application_controller.rb
```diff
@@ -4,7 +4,7 @@ class ApplicationController < ActionController::Base
 
   rescue_from Pundit::NotAuthorizedError, with: :user_not_authorized
 
-  protect_from_forgery with: :null_session
+  protect_from_forgery with: :exception, prepend: true
```
bmesuere, line 7: We recently switched from exception to null to prevent API calls from breaking, methinks. @charvp probably knows more about this.

app/controllers/application_controller.rb
```diff
@@ -45,11 +45,7 @@ def after_sign_out_path_for(_resource)
     stored_location_for(:user) || root_path
   end
 
-  Warden::Manager.after_authentication do |user, auth, _opts|
-    if user.institution.nil?
```
bmesuere, line 49: why was this removed?

app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -8,23 +8,40 @@ class OmniauthCallbacksController < Devise::OmniauthCallbacksController
...
     oauth_login
   end
 
-  def google_oauth2
-    oauth_hash[:info][:institution] = oauth_hash.extra[:raw_info][:hd]
+  def saml
+    user = User.find_by(username: auth_hash.uid)
+    user ||= User.from_email(oauth_email)
+    if user.blank?
+      institution = auth_hash.extra.institution
```
bmesuere, line 29: I probably looked over it, but where was this institution set? A few lines below, `info.institution` is used. Is that the same?

app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -8,23 +8,40 @@ class OmniauthCallbacksController < Devise::OmniauthCallbacksController
...
   end
 
   def office365
     oauth_login
   end
 
-  def google_oauth2
-    oauth_hash[:info][:institution] = oauth_hash.extra[:raw_info][:hd]
+  def saml
+    user = User.find_by(username: auth_hash.uid)
```
bmesuere, line 26:
```suggestion
    # try to fetch the user by username, if that fails try by email, otherwise create a new user
    user = User.find_by(username: auth_hash.uid)
```

app/views/layouts/_drawer.html.erb
```diff
@@ -143,7 +143,7 @@
           <% end %>
         </li>
         <li>
-          <%= link_to destroy_user_session_path, method: :delete do %>
+          <%= link_to users_sign_out_path, method: :delete do %>
```
bmesuere, line 146: I assume this change was made to use the saml sign out? If I recall correctly, we explicitly chose not to use that. For some idp's it was simply broken, for others (like ugent) it resulted in an error if the sessions timed out at the ugent side. In any case, the user was navigated away from dodona with no obvious way to get back.

config/initializers/devise.rb
```diff
@@ -1,7 +1,13 @@
-require_relative('../../lib/SAML/metadata.rb')
-require_relative('../../lib/SAML/saml_controller.rb')
-require_relative('../../lib/SAML/idp_settings_adapter.rb')
-require_relative('../../lib/SAML/my_resource_validator.rb')
+#require_relative('../../lib/SAML/metadata.rb')
+#require_relative('../../lib/SAML/saml_controller.rb')
+#require_relative('../../lib/SAML/idp_settings_adapter.rb')
+#require_relative('../../lib/SAML/my_resource_validator.rb')
```
bmesuere, line 4: remove these?

lib/SAML/settings.rb
```diff
@@ -0,0 +1,41 @@
...
+        ASSERTION_ERROR_INSTITUTIONS_ENTITY_IDS = %w(https://idp.hogent.be/idp https://idp.howest.be/idp/shibboleth)
+
+        def self.base
+          # Load the certificate and private key if on staging/production.
+          certificate = IO.read('/home/dodona/cert.pem') if File.file?('/home/dodona/cert.pem')
+          private_key = IO.read('/home/dodona/key.pem') if File.file?('/home/dodona/key.pem')
+
+          {
+              assertion_consumer_service_binding: 'urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST',
+              assertion_consumer_service_url: "https://#{Socket.gethostbyname(Socket.gethostname).first.downcase}/users/saml/auth",
```
bmesuere, line 14: move hostname to variable

lib/SAML/settings.rb
```diff
@@ -0,0 +1,41 @@
+module OmniAuth
+  module Strategies
+    class SAML
+      class Settings
+        ASSERTION_ERROR_INSTITUTIONS_ENTITY_IDS = %w(https://idp.hogent.be/idp https://idp.howest.be/idp/shibboleth)
```
bmesuere, line 5: why are these here?

lib/SAML/settings.rb
charvp: I think we have the external address in the environment config as well, it's probably cleaner to get this value from there if possible.

app/controllers/application_controller.rb
charvp: See #1807 (which @thepieterdc actually created).

thepieterdc:
> * are the files in the `lib/SAML/old` directory still needed?
I had removed those but apparently restored them in a rebase; these are no longer necessary indeed.

> * are the "backward compatible" paths used in the metadata?
This is only one path; and this is used, to ensure that the metadata is still equivalent to the current one. I am not sure if changing this would break IDP's, but we can never be sure. Fact is that the complete authentication flow works without any metadata at all as well.

> * If I'm correct, some files were copied/modified from another saml implementation, can you link to those files in a comment at the top of the respective files to be able to track down the origin in case of problems?
That's correct; I will do so.

app/controllers/application_controller.rb
thepieterdc: Correct; that's a remnant from when I was wondering why the session was being nullified.

app/controllers/application_controller.rb
thepieterdc: This was required because the previous SAML gem did not set the institution upon logging in. However, since we are now using omniauth, this is performed when https://github.com/dodona-edu/dodona/blob/744f3a3fbec4e794576f733835fd075f1260c474/app/models/user.rb#L259 is called during the login flow at https://github.com/dodona-edu/dodona/blob/09cc20fa01db452bfd47376bda20d4bd5c040f91/app/controllers/omniauth_callbacks_controller.rb#L123.

app/views/layouts/_drawer.html.erb
thepieterdc: This is an interesting one actually.

This route was previously implemented by the `saml_sessions_controller`, included in the previous devise saml gem. This controller extends the base Devise session-controller to allow _IDP-initiated sign out_.

However, the base controller also implements a standard Devise sign-out, not specific to SAML. Since the other authentication providers are also using Devise, this sign-out action works for those as well. By removing the saml gem, there was no way to sign-out of Dodona anymore, regardless of which authentication provider was used. Therefore, I have implemented this action myself in a session controller, but this will be refactored in #2065 .

**TL;DR:** Sign-out was currently working coincidentally.

app/controllers/omniauth_callbacks_controller.rb
thepieterdc: This was set by https://github.com/dodona-edu/dodona/blob/09cc20fa01db452bfd47376bda20d4bd5c040f91/lib/SAML/strategy.rb#L60,which was in turn set by https://github.com/dodona-edu/dodona/blob/09cc20fa01db452bfd47376bda20d4bd5c040f91/lib/SAML/strategy.rb#L77.

The difference is that `auth_hash.info.institution` is an identifier, rather than an `Institution`-instance. In a subsequent refactor of this controller (#2065 ), this discrepancy should ideally be streamlined.

lib/SAML/settings.rb
thepieterdc: image

lib/SAML/settings.rb
thepieterdc: This has been resolved by using the `config.default_host` var.

Gemfile
```diff
@@ -66,7 +66,7 @@ gem 'autoprefixer-rails', '~>9.8.1'
 
 # saml authentication
 gem 'devise', '~>4.7.2'
-gem 'devise_saml_authenticatable', '~> 1.5.0'
+gem 'ruby-saml'
```
bmesuere, line 69: pin version

PR #2093, opened 2020-07-01
LTI authentication

thepieterdc:
This pull request implements LTI 1.3 authentication through OpenID Connect, as described in https://www.imsglobal.org/spec/security/v1p0/#openid_connect_launch_flow.

If a user logs in using a link provider for the first time, they wil be redirected to the default provider for that institution and the id of the current LTI provider will be stored in the session. Upon returning from the default provider, both identities will be linked together for future use.

This implementation has been tested using Moodle. The CSP violation errors can be "solved" by including the below fragment in the application controller. However, this has not been implemented here, as we will require that the LMS does not embed us but rather opens Dodona in an existing/new window. This will be required for Deep Linking support, but there it can be added in the upcoming LtiController.

```ruby
  content_security_policy do |policy|
    policy.frame_ancestors '*'
  end
```

### DIFF
--- a/app/controllers/auth/omniauth_callbacks_controller.rb
+++ b/app/controllers/auth/omniauth_callbacks_controller.rb
@@ -22,6 +22,10 @@ class Auth::OmniauthCallbacksController < Devise::OmniauthCallbacksController
     generic_oauth
   end
 
+  def lti
+    try_login!
+  end
+
   def saml
     try_login!
   end
@@ -46,7 +50,8 @@ class Auth::OmniauthCallbacksController < Devise::OmniauthCallbacksController
 
   def try_login!
     # Ensure the preferred provider is used.
-    return redirect_to_preferred_provider! unless provider.prefer?
+    # TODO add link providers.
+    # return redirect_to_preferred_provider! unless provider.prefer?
 
     # Find the identity.
     identity, user = find_identity_and_user
--- a/app/controllers/auth/omniauth_callbacks_controller.rb
+++ b/app/controllers/auth/omniauth_callbacks_controller.rb
@@ -141,7 +141,9 @@ class Auth::OmniauthCallbacksController < Devise::OmniauthCallbacksController
   def flash_failure(reason)
     return unless is_navigational_format?
 
-    set_flash_message :notice, :failure, kind: auth_provider_type || 'OAuth2', reason: reason
+    # Get the provider type.
+    provider_type = auth_provider_type || request.env['omniauth.error.strategy']&.name || 'OAuth2'
+    set_flash_message :notice, :failure, kind: provider_type, reason: reason
   end
 
   def redirect_to_preferred_provider!
@@ -162,8 +164,8 @@ class Auth::OmniauthCallbacksController < Devise::OmniauthCallbacksController
         "#{auth_hash.pretty_inspect}"
 
     ApplicationMailer.with(authinfo: auth_hash, errors: resource.errors.inspect)
-                     .user_unable_to_log_in
-                     .deliver_later
+        .user_unable_to_log_in
+        .deliver_later
 
     redirect_with_flash! resource.errors.full_messages.to_sentence
   end
@@ -231,8 +233,8 @@ class Auth::OmniauthCallbacksController < Devise::OmniauthCallbacksController
       "#{auth_hash.pretty_inspect}"
 
     ApplicationMailer.with(authinfo: auth_hash)
-                     .institution_created
-                     .deliver_later
+        .institution_created
+        .deliver_later
   end
 
   def institution_create_failed(errors)
@@ -242,8 +244,8 @@ class Auth::OmniauthCallbacksController < Devise::OmniauthCallbacksController
       "#{errors}"
 
     ApplicationMailer.with(authinfo: auth_hash, errors: errors.inspect)
-                     .institution_creation_failed
-                     .deliver_later
+        .institution_creation_failed
+        .deliver_later
   end
 
   def provider_missing!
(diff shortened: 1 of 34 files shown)


rien:
It should be easy to dynamically set the allowed `frame-ancestor`, we already do this in the `ActivitiesController`:
https://github.com/dodona-edu/dodona/blob/ebdd517bc3c3c19c99e74482aab64b612ead5f0a/app/controllers/activities_controller.rb#L19-L21

config/locales/devise.en.yml
```diff
@@ -30,6 +30,7 @@ en:
       user_not_created: "user creation failed"
       unknown_failure: "an unknown failure has occurred"
       failure: "Could not authenticate you from %{kind} because \"%{reason}\"."
+      linked: Because you had not signed in to Dodona via Ufora before, we could not execute the requested operation. Now that your UGent-account has successfully been linked, this should work. Retry clicking on the link in Ufora to try again.
```
bmesuere, line 33:
```suggestion
      linked: Because you have not signed in to Dodona via Ufora before, we could not execute the requested operation. Now that your UGent-account has successfully been linked, this should work. Retry clicking on the link in Ufora to try again.
```

lib/LTI/messages/claims.rb
```diff
@@ -0,0 +1,6 @@
+module LTI::Messages
+  module Claims
+    # Common.
+    TARGET_LINK_URI = 'https://purl.imsglobal.org/spec/lti/claim/target_link_uri'.freeze
```
bmesuere, line 4: what is this for?

thepieterdc:
> This looks very very good. Just to be clear, it will only redirect if the LTI provider is set to redirect (UGent) and not otherwise (ilearn)? The PR says it always redirects.

This PR adds a third kind of provider, next to default/redirect, there is now also `link`. If a user logs in with a `link`-provider, and there is no matching identity, they will also be redirected to the default provider for that institution.

However, after they return from logging in at the default provider, an identity will be created for the link provider they originally intended to sign in with. From now on, they can sign in using this link provider as well.

lib/LTI/messages/claims.rb
thepieterdc: The target link is an LTI claim used to forward the user after logging in. I have stored it in a file that will contain other claims once deep linking is implemented, to have all these constants in one place.

bmesuere:
Thanks, this is what I meant (but mistakenly typed redirect instead of link). However, the top post here in the PR doesn't reflect that behavior.

thepieterdc:
> Thanks, this is what I meant (but mistakenly typed redirect instead of link). However, the top post here in the PR doesn't reflect that behavior.

Fixed.

lib/LTI/auth/setup.rb
```diff
@@ -0,0 +1,61 @@
...
+          return Provider::Lti.find_by(id: params[:provider]) if params[:provider].present?
+
+          # Get the provider from the issuer parameter.
+          return Provider::Lti.find_by(issuer: params[:iss]) if params[:iss].present?
+
+          # If there is a JWT available, we can parse that to find the issuer.
+          return nil if params[:id_token].blank?
+
+          # Parse the JWT token.
+          jwt_token = JSON::JWT.decode params[:id_token], :skip_verification
```
rien, line 55: Why is the verification skipped?

rien:
[The test coverage shows there is no test for linking identities](https://codecov.io/gh/dodona-edu/dodona/compare/16eb75a7c74dab823d51e91e9f1b8699608728be...5cb6e95ea16332532cdebd641b63ed3e2d50a87f/src/app/controllers/auth/omniauth_callbacks_controller.rb?before=app/controllers/auth/omniauth_callbacks_controller.rb). Is there no way to test this?

lib/LTI/auth/setup.rb
thepieterdc: Bare with me. In order to validate the jwt token, we need a public key of the originating party. This public key is stored with the provider on our site. In order to find which LTI provider is being used (e.g. Ufora/I-Learn/...), we can find this in the url (`provider` or `iss` parameter). However, this cannot be included in the callback phase, since this would incur a cyclic dependency in the configuration. To tackle this, I parse the returned jwt token (cannot be validated since the public key is unknown at this stage), extract the `iss` from the token to find the provider and the public key and then validate the token again in the strategy on line 29.

rien: Sounds reasonable :) I would maybe add a comment warning future developers to only use the `iss` parameter and maybe change the line to
```ruby
issuer = JSON::JWT.decode(params[:id_token], :skip_verification)[:iss]
```

app/controllers/auth/omniauth_callbacks_controller.rb
```diff
@@ -151,6 +195,16 @@ def auth_provider_type
...
   end
 
+  def auth_redirect_params
+    auth_hash.extra[:redirect_params].to_h || {}
+  end
+
+  def auth_target
+    return nil if auth_hash.extra[:target].blank?
+
+    "#{auth_hash.extra[:target]}?#{auth_redirect_params.to_param}"
```
rien, line 205: In the staging server the `auth_redirect_params` seem to contain the `id_param` JWT token, causing a `CookieOverflowError` (see #2127). The `id_param` doesn't seem to be needed there, any idea how (and why) this parameter is in the `:redirect_params`?

thepieterdc: This is in fact necessary to allow deep linking with content selection. When deep linking is used, the user will be sent to `/auth/lti/callback` to handle the authentication flow. After this is completed, the user agent must be forwarded to `/lti/activities` (does not yet exist), a page which contains a form with checkboxes to select activities to link to. This form must be submitted to a URL which is encoded in the jwt token, among with various other required settings that describe the format. This is only possible if the entire token is indeed passed on to `/lti/activities`, therefore it must be added to the url.

bmesuere:
@thepieterdc can you resolve the conflict and merge the branch?

thepieterdc:
> @thepieterdc can you resolve the conflict and merge the branch?

Rebase pushed, waiting for tests to complete.

PR #2094, opened 2020-07-02
Introduce questions for students

MrKickkiller:
This pull request allows students to ask questions on their own code and introduces a new questions page for a course where course admins can see all the questions on their course. 

- [x] Tests were added


app/assets/javascripts/auto_reload.ts
```diff
@@ -10,7 +10,7 @@
  * To prevent responding to too many events, they are throttled with requestAnimationFrame.
  */
 export class InactiveTimeout {
-    timout: NodeJS.Timeout | null;
+    timout: number = 0;
```
pdawyndt, line 13: timeout?

app/assets/javascripts/auto_reload.ts
```diff
@@ -54,20 +54,20 @@ export class InactiveTimeout {
...
-        this.interactionElement.removeEventListener("touchmove", this.listener, { passive: true });
-        window.removeEventListener("scroll", this.listener, { passive: true });
+        this.interactionElement.removeEventListener("mousemove", this.listener);
+        this.interactionElement.removeEventListener("touchmove", this.listener);
+        window.removeEventListener("scroll", this.listener);
     }
 
     private startTimeout(): void {
-        this.timout = setTimeout(() => {
+        this.timout = window.setTimeout(() => {
```
pdawyndt, line 63: timeout?

app/assets/javascripts/auto_reload.ts
```diff
@@ -54,20 +54,20 @@ export class InactiveTimeout {
...
             this.callback();
             this.startTimeout();
         }, this.delay);
     }
 
     private endTimeout(): void {
-        clearTimeout(this.timout);
-        this.timout = null;
+        window.clearTimeout(this.timout);
+        this.timout = 0;
```
pdawyndt, line 71: timeout?

app/assets/javascripts/auto_reload.ts
```diff
@@ -54,20 +54,20 @@ export class InactiveTimeout {
...
+        this.timout = window.setTimeout(() => {
             this.callback();
             this.startTimeout();
         }, this.delay);
     }
 
     private endTimeout(): void {
-        clearTimeout(this.timout);
-        this.timout = null;
+        window.clearTimeout(this.timout);
```
pdawyndt, line 70: timeout?

app/policies/question_policy.rb
```diff
@@ -1,9 +1,14 @@
...
+    # If there is no course, don't allow questions.
+    return false if record.submission.course.nil?
+    return false unless record.submission.course.enabled_questions?
+    return false unless record.submission.user == user
+
     all_questions_for_submission = user.questions.where(submission: record.submission)
     total_question_count = all_questions_for_submission.count
     unanswered_question_count = all_questions_for_submission.where(question_state: :unanswered).count
-    record.submission.user == user && unanswered_question_count < 5 && total_question_count < 15
+    unanswered_question_count < 5 && total_question_count < 15
```
pdawyndt, line 11: Do we want to keep these rather arbititrary restrictions? Discussion needed.

niknetniko: We decided to remove this limit

bmesuere (changes requested):
Nice work! Some small remarks:
- On the course settings page, don't make a separate section for questions and add it just below the description. Also, enable the setting for all existing courses and make it the default for new courses
- not convinced of the question mark icon in the nav bar. Maybe use account-question? I would also expand the tooltip to "questions from students" and not just "questions"
- add a tooltip to the ask a question button in the gutter
- button to send "vraag versturen" -> "vraag stellen"
- student tooltip "als behandeld markeren" -> "als beantwoord markeren" ?
- An answered question looks the same as an open question for students (minus the buttons) maybe add an extra (textual) indicator next to the date
- I'm not happy with the icons, suggestions welcome
- On the overview questions page, I wouldn't show the "in behandeling" section if there are no questions with that status
- "(elke 2 seconden als de pagina niet gebruikt wordt)" in fact it only refreshes when the page is used, but I would remove that part of the setting.
- maybe also add the setting to enable questions for the course to the questions page?
- the column width of the tables is not fixed which causes unaligned columns over the 2 or 3 tables

niknetniko:
> * I'm not happy with the icons, suggestions welcome

I'll look at the rest tomorrow, but I did consider this briefly, and this is were I ended:

image

With the two people (`account-supervisor`) meaning in progress, from the idea that someone is next to student explaining things.

pdawyndt:
`chat-processing` (busy talking ... to student)
`comment-quote` (busy writing answer ... to student)
`message-draw` (busy writing answer ... to student)

app/assets/javascripts/code_listing/code_listing.ts
```diff
@@ -252,15 +267,20 @@ export class CodeListing {
         const globalButton = document.querySelector(annotationGlobalAdd);
         globalButton.addEventListener("click", () => this.handleAnnotateGlobal());
 
+        const type = this.questionMode? "user_question" : "user_annotation";
```
charvp, line 270:
```suggestion
        const type = this.questionMode ? "user_question" : "user_annotation";
```

app/assets/javascripts/code_listing/code_listing.ts
```diff
@@ -373,7 +395,9 @@ export class CodeListing {
         // Deletion handler.
         if (deleteButton !== null) {
             deleteButton.addEventListener("click", async () => {
-                if (confirm(I18n.t("js.user_annotation.delete_confirm"))) {
+                const type = this.questionMode? "user_question" : "user_annotation";
```
charvp, line 398:
```suggestion
                const type = this.questionMode ? "user_question" : "user_annotation";
```

app/assets/javascripts/code_listing/user_annotation.ts
```diff
@@ -29,6 +34,8 @@ export interface UserAnnotationData {
     evaluation_id: number | null;
     url: string;
     user: UserAnnotationUserData;
+    type: string;
+    question_state: QuestionState | null;
 }
 
 export class UserAnnotation extends Annotation {
```
charvp, line 41: Did you consider subtyping a `QuestionAnnotation` from `UserAnnotation`? There's a few times where you `if` on `this.type` which very much feels like it could be cleaner with inheritance.

app/controllers/annotations_controller.rb
```diff
@@ -37,6 +44,27 @@ def update
...
+      format.json { render :show, status: :ok, location: @annotation }
+    end
+  end
+
+  def resolve
+    @annotation.answered!
+    respond_to do |format|
+      format.json { render :show, status: :ok, location: @annotation }
+    end
+  end
```
charvp, line 66: These feel like updates that could/should be part of the `update` action.

app/controllers/courses_controller.rb
```diff
@@ -94,10 +94,10 @@ def create
...
         Series.new(
           series_memberships: @copy_options[:exercises] ?
-                                  s.series_memberships.map do |sm|
-                                    SeriesMembership.new(activity: sm.activity, order: sm.order)
-                                  end :
-                                  [],
+                                s.series_memberships.map do |sm|
+                                  SeriesMembership.new(activity: sm.activity, order: sm.order)
+                                end :
+                                [],
```
charvp, line 100: These are only whitespace changes?

app/helpers/application_helper.rb
```diff
@@ -219,10 +219,11 @@ def previous_or_next_page(page, text, classname)
   end
 
   def page_navigation_links(pages, remote = false, controller = '', params = {}, action = 'index')
+    param_name = params[:param_name] || 'page'
```
charvp, line 222: I would prefer a sixth argument for this. `params` is meant to pass on parameters for `will_paginate`, not for params to this method.

app/models/annotation.rb
```diff
@@ -31,6 +33,10 @@ class Annotation < ApplicationRecord
   after_save :create_notification
   after_destroy :destroy_notification
 
+  def to_partial_path
```
charvp, line 36: Doesn't look like this is used? Or is this some rails magic?

app/models/question.rb
```diff
@@ -0,0 +1,59 @@
...
+  enum question_state: { unanswered: 0, in_progress: 1, answered: 2 }
+  alias_attribute :question_text, :annotation_text
+
+  # Saves from which expected state the question should migrate, to prevent unexpected state changes.
+  attr_accessor :transition_from
+
+  default_scope { order(created_at: :desc) }
+
+  # Fix for routing. Otherwise it would require question_url instead of the annotation_url
+  def self.model_name
```
charvp, line 28: Is there really no other way to do this? This very much feels like a hack.

app/models/question.rb
```diff
@@ -0,0 +1,59 @@
...
+    superclass.model_name
+  end
+
+  # Fix the above fix
+  def self.policy_class
+    QuestionPolicy
+  end
+
+  def self.per_page
```
charvp, line 37: I think we always prefer not to have any presentation layer concerns in the models. In this case, adding the `per_page` argument to the `paginate` call in the controller seems cleaner to me.

app/views/annotations/_annotation.json.jbuilder
```diff
@@ -9,5 +10,9 @@ end
 json.permission do
   json.update policy(annotation).update?
   json.destroy policy(annotation).destroy?
+  json.unresolve policy(annotation).unresolve?
+  json.in_progress policy(annotation).in_progress?
+  json.resolve policy(annotation).resolve?
 end
 json.released AnnotationPolicy.new(annotation.submission.user, annotation).show?
+json.type annotation.type&.downcase || 'annotation'
```
charvp, line 18: I would hope that the `||` here isn't actually required?

db/migrate/20200629132522_add_questions.rb
```diff
@@ -0,0 +1,14 @@
+class AddQuestions < ActiveRecord::Migration[6.0]
+  def change
+    reversible do |dir|
+      dir.up {
+        add_column :annotations, :type, :string
```
charvp, line 5: As mentioned before, I wouldn't allow this to be null and just add a default of `annotation`. Or you could do it in three steps: add the column, fill it for all existing annotations and then change the nullability.

charvp (changes requested):
Some nitpicks, some questions, but I don't see any major problems with the code.

bmesuere:
Maybe move the notification dot to the lower left corner instead of the right? This solves the overlap problem.

bmesuere (changes requested):
- "Vragen zijn uitgeschakeld. Schakel dit in in de cursusopties." Maybe link "cursusopties" to the correct page? Or even better, make it possible to enable/disable from this page.
- Turn "auto refresh" on by default
- I think `comment-processing-outline` is clearer for use as pending icon
- The questions aren't paginated right now. If we don't add this now, we'll probably have to hotfix it after the first week. An alternative is to cap the list to the first 20/30/50 questions and add a link to a separate page where all questions can be seen (with pagination). This would make it a bit easier to implement without having to mess with the pagination in combination with the auto reload

app/assets/javascripts/code_listing/user_annotation.ts
niknetniko: I have tried this, but the subtype resulted in a bunch of `if`s inside the code listing, since all annotations (normal ones and questions) come from the same endpoint `/annotations`.

app/controllers/annotations_controller.rb
niknetniko: The current method has a big advantage: the policy for the question is relatively clean. I tried putting this all together in update, but I wasn't sure how the policy object should look in that case (e.g. course administrators are allowed to update the state of the question, but not the text of the question at the moment)

app/controllers/courses_controller.rb
niknetniko: My bad, I've reverted them

app/models/annotation.rb
niknetniko: Indeed rails magic, this determines the path for the partials for this model. I've moved it to the `Question` model.

app/models/question.rb
niknetniko: I haven't found a way 😢

niknetniko:
> Turn "auto refresh" on by default

It is on by default if questions are enabled for the course.

> The questions aren't paginated right now. If we don't add this now, we'll probably have to hotfix it after the first week. An alternative is to cap the list to the first 20/30/50 questions and add a link to a separate page where all questions can be seen (with pagination). This would make it a bit easier to implement without having to mess with the pagination in combination with the auto reload

They are paginated, but the pagination controls only show if there are enough questions (each sections gets its own, independent pagination). Perhaps we should always show them?
image

Will look at changing options from this page tomorrow.

app/assets/javascripts/code_listing/user_annotation.ts
charvp: Hmm, I think I would still prefer the if's to be in code listing, because then we can dispatch to the right class and not worry about it for the rest of the runtime. But that's not a strong opinion, so if you feel otherwise, feel free to ignore me.

app/controllers/annotations_controller.rb
charvp: `permitted_attributes` should be used to determine what elements of the model are allowed to be updated.

app/models/question.rb
charvp: How do we do this for exercises and reading activities? They both show up under `activities_url`, so it's not entirely the same, but maybe there is some inspiration to take from there.

app/assets/javascripts/code_listing/user_annotation.ts
niknetniko: I've changed it to use a subclass

app/models/question.rb
niknetniko: The comment was bit misleading; one of the main issues was that rails expects values with a name based on the class (e.g. it expects `question` vs `annotation`). This was solved relatively easily with the subclass in typescript, so I've removed these lines.

charvp (approved):

PR #2102, opened 2020-07-08
Fix institution views

thepieterdc:
This pull request fixes the institution views and adds a very basic integration test that visits these pages to ensure no error is thrown.

image

### DIFF
--- a/app/views/institutions/_form.html.erb
+++ b/app/views/institutions/_form.html.erb
@@ -17,8 +17,4 @@
     <%= f.label :short_name, :class => "col-sm-3 col-xs-12 control-label" %>
     <div class="col-sm-6 col-xs-12"><%= f.text_field :short_name, class: "form-control", disabled: !f.permission?(:short_name) %></div>
   </div>
-  <div class="field form-group">
-    <%= f.label :identifier, :class => "col-sm-3 col-xs-12 control-label" %>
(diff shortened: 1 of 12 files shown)


app/views/institutions/_form.html.erb
```diff
@@ -17,8 +17,4 @@
     <%= f.label :short_name, :class => "col-sm-3 col-xs-12 control-label" %>
     <div class="col-sm-6 col-xs-12"><%= f.text_field :short_name, class: "form-control", disabled: !f.permission?(:short_name) %></div>
   </div>
-  <div class="field form-group">
-    <%= f.label :identifier, :class => "col-sm-3 col-xs-12 control-label" %>
-    <div class="col-sm-6 col-xs-12"><%= f.text_field @institution.saml? ? :entity_id : :identifier, class: "form-control", disabled: true %></div>
-  </div>
```
charvp, line 23: It would be nice if we could show the identifiers in the form as well. They're often useful to look up institution names.

thepieterdc: My edit to the PR was apparently not saved. I have updated the first post to include a screenshot, which displays the identifiers.

charvp: That's `institutions#show` right? I was talking about `institutions#edit`. (When updating institutions, I don't use `institutions#show`.)

thepieterdc: I have added the table to the `#edit` view as well now.

PR #2125, opened 2020-07-29
Convert database character set to UTF8MB4

### LINKED ISSUE #1961: Convert database to utf8mb4 to support emoji
When trying to save a string with emoji in it, the database will throw an error `Mysql2::Error: Incorrect string value` because the database collation doesn't support four-byte utf8 characters (#1664). Currently we invalidate some models when these kind of characters are present (e.g. in submissions), but we often don't.

In order to be able to support emoji (in annotations, exercise titles, course titles, ...) the database tables should be converted to the `utf8mb4` collation. This is a very big change with a risk of potentially corrupting our database if we don't do this smart.

I've made this issue to collect some research about how we can do this migration while minimizing the risk of completely destroying our database.

rien:
This pull request

- [x] Tests were added

Closes #1961.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,11 +13,14 @@ The documentation of this project can be found at https://dodona-edu.github.io.
     ```
     sql-mode='STRICT_TRANS_TABLES,NO_ZERO_IN_DATE,NO_ZERO_DATE,ERROR_FOR_DIVISION_BY_ZERO,NO_ENGINE_SUBSTITUTION'
     ```
-3. Create a `dodona` user with access to the `dodona` and `dodona_test` databases:
+3. Create a `dodona` user with access to the `dodona` and `dodona_test-N` databases. You will need as much test databases as hou have CPU threads.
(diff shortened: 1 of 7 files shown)


rien:
I've loaded a Naos db dump into my own local environment and there we no issues.  Converting the `submissions` table was the most time consuming operation, lasting 75 seconds for almost 1M submissions. Our production database (with 5.5M submissions) would require around 7 minutes. But that's on other hardware so we should first measure the migration time on the actual Naos server to have a realistic upper bound.

MySQL has proper checks on data types, so if a utf8mb3 connection want to query or edit a string in a utf8mb4 database, it notices an throws an error. So there is a very low risk of this corrupting our database.

Once I reset the connections (by restarting the application), I was able to change courses, series and users to contain emoji and was able to search for them using only emoji as query string. :tada:

bmesuere (approved):

PR #2131, opened 2020-08-03
Institution name is now pseudonymized

### LINKED ISSUE #1300: pseudonymize institute names in demo mode
Names of institutes can be considered as personally identifiable information, especially in courses where all users come from different institutes. Therefore we should add the following pseudonimizations in demo mode:

- institute names
- email domain names

LucianosLionakis:
When in demo mode the institution names are now pseudonymized.

Closes #1300

Previous Demo mode:
image

New Demo mode:
image

### DIFF
--- a/app/models/institution.rb
+++ b/app/models/institution.rb
@@ -21,6 +21,12 @@ class Institution < ApplicationRecord
 
   scope :of_course_by_members, ->(course) { joins(users: :courses).where(courses: { id: course.id }).distinct }
 
+  def name
+    return self[:name] unless Current.demo_mode
+    
+    Faker::University.name
+  end
+
   def preferred_provider
     Provider.find_by(institution: self, mode: :prefer)
   end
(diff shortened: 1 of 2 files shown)


pdawyndt:
Are users from the same institute still assigned to the same institute in demo mode? Is this something we want?

LucianosLionakis:
> Are users from the same institute still assigned to the same institute in demo mode? Is this something we want?

No, users from the same institute are not assigned to the same institute in demo mode. I don't know if this would be something we want.

pdawyndt:
Keeping users in the same institution could be beneficial if the institution is used as a selector or as a category e.g. in learning analytics.

LucianosLionakis:
Wouldn't this create edge cases in which students can be identified? For example a course where only one student is from another institute would make him/her identifiable.
This edge case might be far fetched. In any case this would be easy to implement.

LucianosLionakis:
> @LucianosLionakis I think user _are_ assigned the same random institution. The seed is set to the id of the instution + the day. The result of this is that a certain institution will always be "randomized" in the same way for the entire day. This is what we want.

Yes, I changed this in my last push because I thought my example was far fetched.

PR #2132, opened 2020-08-03
Hide copy button on empty submission

### LINKED ISSUE #2053: Hide copy-button when empty code is submitted
When empty code is submitted, the copy button hangs out of the panel. Since there's no need to copy the code, we could hide the button in this case.

image

winniederidder:
This pull request hides the copy button when the submission is empty.

Visuals now:
empty_code
Previous visuals can be found in:
https://github.com/dodona-edu/dodona/issues/2053

- [ ] Tests:
Seed was changed to sometimes submit empty code to allow this case to be tested more easily.

Closes #2053 .

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -220,7 +220,8 @@ if Rails.env.development?
                           status: before,
                           accepted: before == :correct,
                           created_at: before_deadline,
-                          code: code,
+                          # allow Status Test to contain empty submissions for layout testing
+                          code: rand() > 0.5 ? code : '',
                           result: File.read(Rails.root.join('db', 'results', "#{exercise.judge.name}-result.json"))
       end
       if after != :none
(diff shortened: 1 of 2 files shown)


db/seeds.rb
```diff
@@ -220,7 +220,8 @@
                           status: before,
                           accepted: before == :correct,
                           created_at: before_deadline,
-                          code: code,
+                          # allow Status Test to contain empty submissions for layout testing
+                          code: rand() > 0.5 ? code : '',
```
rien, line 224: I don't think this change is needed. You can simply test this by creating an empty submission in the development environment.

Having 50% empty submissions isn't really useful :)

winniederidder: It's the Status Test course only, so it's 50% of a very small number where we don't really care for the contents. Creating empty submissions in dev, what do you mean? If you mean, simply submitting one, then that one will be 'loading' forever as it is not processed, so the screen that has been edited with this PR can't be accessed.

rien: Ah, I didn't notice it was only for `Status Test`. It is OK there.

Keep in mind that you _can_ run submissions in development, but you'll have to set up docker and run delayed_job.

PR #2138, opened 2020-08-03
Feature/programming language icons

### LINKED ISSUE #1975: Use language-specific icons for programming exercises
In a series we could replace the neutral icon for a programming exercise by an icon specific for the programming language used in the exercise. There seem to be [material design icons](https://materialdesignicons.com/) for most of the popular programming languages.

An exam series that mixes Python and JavaScript exercises now looks like this

image

but could then look something like

image

In case we later move to programming exercises that support multiple programming languages, we may then switch to showing the programming language of the latest submission, maybe still in combination with an icon for the exercise (e.g. related to the judge linked to the exercise).

image

winniederidder:
This pull request uses different icons for activities depending on their Programming Language.

Sample of a current table:
new_icons

Current issues:
Example testing exercises are always linked to Python, so e.g. Echo JS shows a Python logo.
Blank, a.k.a. Programming Language = text is also not available yet to test visually.

Exercises are created via a git repo, so I require explanation on how to work with that.
Closes #1975  .


app/views/evaluations/_exercises_progress_table.html.erb
```diff
@@ -12,7 +12,7 @@
     <% local_assigns[:metadata][:per_exercise].each do |meta| %>
       <tr>
         <td class='type-icon'>
-          <i class="mdi mdi-file-document-edit-outline mdi-18" title="<%= t "activities.index.type.exercise" %>"></i>
+          <i class=<%"mdi mdi-#{meta[:exercise].programming_language&.icon&.blank? ? "file-document-edit-outline" : meta[:exercise].programming_language&.icon&} mdi-18"%> title="<%= t "activities.index.type.exercise" %>"></i>
```
rien, line 15:
```suggestion
          <i class=<%"mdi mdi-#{meta[:exercise].programming_language&.icon&.blank? ? "file-document-edit-outline" : meta[:exercise].programming_language&.icon} mdi-18"%> title="<%= t "activities.index.type.exercise" %>"></i>
```
I think there is an `&` to many here?

db/migrate/20200803125737_add_icon_to_programming_languages.rb
```diff
@@ -0,0 +1,22 @@
+class AddIconToProgrammingLanguages < ActiveRecord::Migration[6.0]
+
+  @@icon_map = {
+    'python' => 'language-python',
+    'sh' => 'mdi-bash',
```
rien, line 5:
```suggestion
    'sh' => 'bash',
```

db/migrate/20200803125737_add_icon_to_programming_languages.rb
```diff
@@ -0,0 +1,22 @@
+class AddIconToProgrammingLanguages < ActiveRecord::Migration[6.0]
+
+  @@icon_map = {
```
rien, line 3:
```suggestion
  ICON_MAP = {
```
Creates a constant instead of class variable.

db/migrate/20200803125737_add_icon_to_programming_languages.rb
```diff
@@ -0,0 +1,22 @@
...
+    'haskell' => 'language-haskell',
+    'R' => 'language-r',
+    'csharp' => 'language-csharp',
+    'text' => nil
+  }
+  def change
+    add_column :programming_languages, :icon, :string
+
+    ProgrammingLanguage.all.each do |language|
+      language.update(icon: @@icon_map[language.name])
```
rien, line 19:
```suggestion
      language.update(icon: ICON_MAP[language.name])
```

winniederidder:
Picture with icons from different languages. 
new_icons

app/views/activities/_activities_table.html.erb
```diff
@@ -47,7 +47,7 @@
 
         <td class='type-icon'>
           <% if activity.exercise? %>
-            <i class="mdi mdi-file-document-edit-outline mdi-18" title="<%= t "activities.index.type.exercise" %>"></i>
+            <i class="<%= "mdi mdi-#{activity.programming_language&.icon&.blank? ? "file-document-edit-outline" : activity.programming_language.icon} mdi-18" %>" title="<%= t "activities.index.type.exercise" %>"></i>
```
rien, line 50:
```suggestion
            <i class="<%= "mdi mdi-#{ activity.programming_language&.icon || "file-document-edit-outline" } mdi-18" %>" title="<%= t "activities.index.type.exercise" %>"></i>
```
If `programming_language.icon` is `nil` then `programming_language&.icon&.blank?` will return `nil`, giving the opposite result that you want: a falsey value where you'd expect a truthy one.

app/views/activities/_series_activities_table.html.erb
```diff
@@ -20,8 +20,7 @@
         <%= raw "<i class='mdi mdi-eye-off mdi-18' title='#{t("activities.index.private")}'></i>" if activity.access_private? %>
         <%= raw "<i class='mdi mdi-delete mdi-18' title='#{t("activities.index.removed")}'></i>" if activity.removed? %>
         <%= raw "<i class='mdi mdi-alert-circle mdi-18' title='#{t("activities.index.invalid")}'></i>" if activity.not_valid? %>
-
-        <%= raw "<i class='mdi mdi-file-document-edit-outline mdi-18' title='#{t("activities.index.type.exercise")}'></i>" if activity.exercise? %>
+        <%= raw "<i class='mdi mdi-#{activity.programming_language&.icon&.blank? ? "file-document-edit-outline" : activity.programming_language.icon} mdi-18' title='#{t("activities.index.type.exercise")}'></i>" if activity.exercise? %>
```
rien, line 23:
```suggestion
        <%= raw "<i class='mdi mdi-#{ activity.programming_language&.icon || "file-document-edit-outline" } mdi-18' title='#{t("activities.index.type.exercise")}'></i>" if activity.exercise? %>
```

app/views/evaluations/_exercises_progress_table.html.erb
```diff
@@ -12,7 +12,7 @@
     <% local_assigns[:metadata][:per_exercise].each do |meta| %>
       <tr>
         <td class='type-icon'>
-          <i class="mdi mdi-file-document-edit-outline mdi-18" title="<%= t "activities.index.type.exercise" %>"></i>
+          <i class=<%"mdi mdi-#{meta[:exercise].programming_language&.icon&.blank? ? "file-document-edit-outline" : meta[:exercise].programming_language&.icon} mdi-18"%> title="<%= t "activities.index.type.exercise" %>"></i>
```
rien, line 15:
```suggestion
          <i class=<%"mdi mdi-#{meta[:exercise].programming_language&.icon || "file-document-edit-outline" } mdi-18"%> title="<%= t "activities.index.type.exercise" %>"></i>
```

app/views/evaluations/_exercises_table.html.erb
```diff
@@ -11,7 +11,7 @@
     <% local_assigns[:exercises].each do |exercise| %>
       <tr>
         <td class='type-icon'>
-          <i class="mdi mdi-file-document-edit-outline mdi-18" title="<%= t "activities.index.type.exercise" %>"></i>
+          <i class=<%"mdi mdi-#{exercise.programming_language&.icon&.blank? ? "file-document-edit-outline" : exercise.programming_language.icon} mdi-18"%> title="<%= t "activities.index.type.exercise" %>"></i>
```
rien, line 14:
```suggestion
          <i class=<%"mdi mdi-#{ activity.programming_language&.icon || "file-document-edit-outline" } mdi-18"%> title="<%= t "activities.index.type.exercise" %>"></i>
```

pdawyndt:
If we also attach a programming language to a submission (inherited from the exercise for now), we could also add the programming language icon to all tables showing a list of submissions.

app/views/evaluations/_exercises_table.html.erb
winniederidder: Should be exercise still, so fixed in a separate commit.

rien: Ah yes, copy-paste went wrong.

rien:
> @rien don't we want to handle the "no icon" case in the model so we don't have to add that exception each time we use it?

I would maybe put it in a helper, because otherwise we're mixing application logic with its views.

pdawyndt:
Also useful to show the icons in the overview of learning activities for a series in the right margin (navigation element) and also on the exercise page itself (after the title in the header?).

bmesuere:
> > @rien don't we want to handle the "no icon" case in the model so we don't have to add that exception each time we use it?
> 
> I would maybe put it in a helper, because otherwise we're mixing application logic with its views.

Yes, but this _is_ logic, right? It seems logical to overwrite the icon attribute in the model and return the string from the datababase or "file-document-edit-outline". That way there's one place where de default icon is set and you can trust the `icon` call to return something useful.

rien:
> Yes, but this _is_ logic, right? It seems logical to overwrite the icon attribute in the model and return the string from the datababase or "file-document-edit-outline". That way there's one place where de default icon is set and you can trust the `icon` call to return something useful.

Hmm yes, you're right. I think putting the icon in the database was mixing logic and views with each other to begin with. But adding a constant mapping in the `ProgrammingLanguage` class is also not ideal.

I would indeed write something in the model like
```ruby
def icon
  self[:icon] || "file-document-edit-outline"
end
```

bmesuere:
@winniederidder can you modify the code as proposed by @rien and add the title?

winniederidder:
Updated tooltip visual:
new_icons

winniederidder:
> The tests still seem to fail. Probably a missing `&.` somewhere.

Was working on that while also updating the i18n
Visuals now:
new_icons

app/views/evaluations/_exercises_progress_table.html.erb
```diff
@@ -12,7 +12,7 @@
     <% local_assigns[:metadata][:per_exercise].each do |meta| %>
       <tr>
         <td class='type-icon'>
-          <i class="mdi mdi-file-document-edit-outline mdi-18" title="<%= t "activities.index.type.exercise" %>"></i>
+          <i class=<%"mdi mdi-#{meta[:exercise].programming_language&.icon} mdi-18"%> title="<%= "#{t "activities.index.type.exercise_language", language: meta[:exercise].programming_language.name.titleize}" %>"></i>
```
bmesuere, line 15: Is this correct? I think it should be `<%=`

app/views/evaluations/_exercises_table.html.erb
```diff
@@ -11,7 +11,7 @@
     <% local_assigns[:exercises].each do |exercise| %>
       <tr>
         <td class='type-icon'>
-          <i class="mdi mdi-file-document-edit-outline mdi-18" title="<%= t "activities.index.type.exercise" %>"></i>
+          <i class=<%"mdi mdi-#{exercise.programming_language&.icon} mdi-18"%> title="<%= "#{t "activities.index.type.exercise_language", language: exercise.programming_language.name.titleize}" %>"></i>
```
bmesuere, line 14: Is this correct? I think it should be `<%=`

thepieterdc: The quotes are off as well

PR #2144, opened 2020-08-04
Zero submission message now uses full row

### LINKED ISSUE #2052: Use full width for zero-submission message
When a user has no submissions for an exercise, the corresponding message could use the full width of the panel, instead of half the width.

image

LucianosLionakis:
Previous:
image

Now:
image
Closes #2052.

### DIFF
--- a/app/views/feedbacks/show.html.erb
+++ b/app/views/feedbacks/show.html.erb
@@ -15,7 +15,7 @@
       </div>
       <div class="card-supporting-text card-border">
         <div class="row">
-        <% if (@feedback.submission.present? && @feedback.later_attempts + @feedback.previous_attempts > 0) || @feedback.total_attempts > 0 %>
+        <% if @feedback.total_attempts > 0 %>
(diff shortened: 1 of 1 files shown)


app/views/feedbacks/show.html.erb
```diff
@@ -15,6 +15,7 @@
       </div>
       <div class="card-supporting-text card-border">
         <div class="row">
+        <% if (@feedback.submission.present? && @feedback.later_attempts + @feedback.previous_attempts > 0) || @feedback.total_attempts > 0 %>
```
bmesuere, line 18: @rien Would it be worth it to move this large check to the model?

rien: Probably, but I've got a feeling this check can be simplified.
@LucianosLionakis would `@feedback.total_attemts > 0` be enough for this condition?

LucianosLionakis: Yes, @rien this actually seems to be enough after testing a bit.

PR #2145, opened 2020-08-04
Add media method to repository_controller

### LINKED ISSUE #1839: Support direct links to files in public directory of repository
The files in the `/public` directory of a repository are accessible in the descriptions of an exercise, as if they were symlinked in the `media` directory of that description. In addition, it would be useful if these files were also accessible via a URL that does not refer to a specific exercise. Something like

```
https://dodona.ugent.be/en/repositories/61/public/CodersApprentice.png
```

This way, the files could also be referenced for example from course and series descriptions.

winniederidder:
This pull request allows media files in the public folder of Dodona repositories to be used similar to activity media files.

Example of an image in a course description that is found in the example-exercises repository
course_description_with_picture

- [x] Tests were added
  A test was added in repositories_controller_test.rb to check if an admin can obtain the image (functionality)
  A test was added with a signed out user to check it is actually public (policy)
- [x] Documentation update can be found at  PR https://github.com/dodona-edu/dodona-edu.github.io/pull/62

Closes #1839  .


test/controllers/activities_controller_test.rb
```diff
@@ -111,7 +111,7 @@ def setup
     @instance.stubs(:media_path).returns(Pathname.new('not-a-real-directory'))
     Repository.any_instance.stubs(:full_path).returns(Pathname.new(Rails.root))
```
rien, line 112: Can you also change this path to `test/remotes/exercises/echo`?

test/controllers/repositories_controller_test.rb
```diff
@@ -13,6 +13,15 @@ class RepositoriesControllerTest < ActionDispatch::IntegrationTest
     sign_in @admin
   end
 
+  def request_public_image
+    @instance.stubs(:public_path).returns(Pathname.new('not-a-real-directory'))
+    Repository.any_instance.stubs(:full_path).returns(Pathname.new('test/remotes/exercises'))
```
rien, line 18: `test/remotes/exercises` is not meant to be one repository, it contains two repositories: `echo` (containing a single 'echo'exercise) and `lasagna` (for testing dirconfig merging and multiple exercises).

Your image should thus be moved to `test/remotes/exercises/echo/public/`.

winniederidder:
Visuals of the file listing:
public_files

app/controllers/repositories_controller.rb
```diff
@@ -16,6 +16,10 @@ def index
   def show
     @title = @repository.name
     @crumbs = [[I18n.t('repositories.index.title'), repositories_path], [@repository.name, '#']]
+    pathname_dir = Pathname.new @repository.public_path
+    @files = Dir[File.join(@repository.public_path, '**/*')] # uses entries in the locally stored Dodona copy of the repository
```
rien, line 20: Extract this function to the model.

app/views/repositories/show.html.erb
```diff
@@ -57,6 +57,26 @@
...
+          <table class="table table-index table-resource">
+            <thead>
+              <tr>
+                <th><%= t ".public_files" %></th>
+              </tr>
+            </thead>
+            <tbody>
+              <% @files.each do |file| %>
+                <tr>
+                  <td><%= link_to file.to_s, public_repository_url(@repository, file) %></td>
```
rien, line 72: I guess we want a way to quickly copy the URL. We already have a 'copy to clipboard'-button, but I fear having a button for each file would become too cluttered maybe. @bmesuere any ideas? I guess we can also just leave it like this and expect a user to be able to right click and use _copy link_.

app/views/repositories/show.html.erb
```diff
@@ -57,6 +57,26 @@
...
+      <div class="card-supporting-text">
+        <div class="table-scroll-wrapper">
+          <table class="table table-index table-resource">
+            <thead>
+              <tr>
+                <th><%= t ".public_files" %></th>
+              </tr>
+            </thead>
+            <tbody>
+              <% @files.each do |file| %>
```
rien, line 70: I would make this feature more discoverable and add a placeholder message if there are no files. Something like
> Files in the `/public/` directory of the repository are listed here and will be available to use in course, series and exercise descriptions under the url `#{ public_respository_url(@repository)}`.

app/views/repositories/show.html.erb
winniederidder: There is a lot of room in the table rows so it could be added in a small column. But right click copy doesn't sound so hard if we expect them to construct an entire repository with a lot of files in a specific structure for any of this to be relevant :P

winniederidder:
Placeholder visuals:
placeholder
May not be ideal so open to suggestions

app/views/repositories/show.html.erb
```diff
@@ -57,6 +57,31 @@
...
+            <% @files.each do |file| %>
+              <tr>
+                <td><%= link_to file.to_s, public_repository_url(@repository, file) %></td>
+              </tr>
+            <% end %>
+            </tbody>
+            <% end %>
+          </table>
+          <% if !@files.present? %>
+            <p id="no-public-files"><%= t ".public_files_placeholder", url:  public_repository_url(@repository, '') %></p>
```
rien, line 80: We already use specific styling for placeholders like this, see https://github.com/dodona-edu/dodona/blob/develop/app/views/api_tokens/_table.html.erb.

config/locales/views/repositories/nl.yml
```diff
@@ -20,6 +20,8 @@ nl:
       exercises: Oefeningen
       content_pages: Leesactiviteiten
       learning_activities: Leeractiviteiten
+      public_files: Publieke bestanden
+      public_files_placeholder: "Bestanden in de /public/ folder van de repository worden hier opgelijst en kunnen gebruikt worden in cursus-, serie- en oefeningenbeschrijvingen via de url %{url}."
```
rien, line 24:
```suggestion
      public_files_placeholder: "Bestanden in de /public/ map van de repository worden hier opgelijst en kunnen gebruikt worden in cursus-, reeks- en oefeningenbeschrijvingen onder de url %{url}."
```

app/models/repository.rb
```diff
@@ -42,7 +43,22 @@ class Repository < ApplicationRecord
   has_many :exercises, dependent: :restrict_with_error
 
   def full_path
-    Pathname.new File.join(ACTIVITY_LOCATIONS, path)
+    if path.present?
+      Pathname.new File.join(ACTIVITY_LOCATIONS, path)
+    else
+      ''
+    end
```
rien, line 50:
```suggestion
    Pathname.new File.join(ACTIVITY_LOCATIONS, path)
```
This is only something that occurs during testing, so we don't want to handle this in the model.

app/views/repositories/show.html.erb
winniederidder: So I shouldn't use the table header at all in that case, but only a p with those classes?

winniederidder:
New placeholder visuals based on @rien 's suggestions:
placeholder

app/views/repositories/show.html.erb
```diff
@@ -57,6 +57,30 @@
         <div id="activities-table-wrapper">
         </div>
       </div>
+      <% if policy(@repository).public? %>
+        <div class="card-supporting-text">
```
rien, line 61: I think a `<h3><%= t ".public_files" %></h3>` here instead of a table header is more in line with our style. Can you also try adding the `card-border` class?

app/models/repository.rb
```diff
@@ -45,6 +46,17 @@ def full_path
...
 
+  def public_path
+    full_path + PUBLIC_DIR
+  end
+
+  def public_files
+    pathname_dir = Pathname.new public_path
+    Dir[File.join(public_path, '**/*')] # uses entries in the locally stored Dodona copy of the repository
+      .select { |f| File.file? f } # skip directories such as ., icons/
+      .map { |f| Pathname.new(f).relative_path_from pathname_dir } # make them relative again to be able to create urls
```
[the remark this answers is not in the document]
rien: Context: we've had issues with unreadable files in the past, reading them gives internal server errors.

app/views/repositories/show.html.erb
rien: Yes, with the classes used here: https://github.com/dodona-edu/dodona/blob/325fc3d91b45a0ada13585031d2fa075c9d8070f/app/views/api_tokens/_table.html.erb#L1-L3

app/views/repositories/show.html.erb
```diff
@@ -57,6 +57,30 @@
         <div id="activities-table-wrapper">
         </div>
       </div>
+      <% if policy(@repository).public? %>
```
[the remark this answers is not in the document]
winniederidder: It does fit in the style and is no doubt future-proofed :P. @bmesuere would you like this to be removed aswell?

winniederidder:
Visuals after changing some HTML elements:
With files:
card-border-with-files
With placeholder:
card-border-with-placeholder

winniederidder:
With default information text:
with files present:
public-files-table
without files:
public-files-placeholder

config/locales/views/repositories/nl.yml
```diff
@@ -20,6 +20,11 @@ nl:
       exercises: Oefeningen
       content_pages: Leesactiviteiten
       learning_activities: Leeractiviteiten
+      public_files: Publieke bestanden
+      public_files_info: "Bestanden in de /public/ map van de repository worden hier opgelijst en kunnen gebruikt worden in cursus-, reeks- en oefeningenbeschrijvingen onder de url %{url}."
```
bmesuere, line 24:
```suggestion
      public_files_info: "Bestanden in de <code>/public</code> map van een repository kunnen overal op Dodona gebruikt worden, bijvoorbeeld in cursus-, reeks- en oefeningenbeschrijvingen, via onderstaande url."
```
You will probably need to add `_html` to the key to make the code tag work + add similar change to the English version.

winniederidder:
> The test remotes files probably need to be in the example exercise repo and not here? Shouldn't the .gitignore file ingore these?

The files are used in the tests to check the fallback action using stubbed urls pointing to those files.

winniederidder:
Updated text:
updated-info

winniederidder:
All changed to h4 to have uniform style
h4page

PR #2146, opened 2020-08-05
Upgrade to MathJax 3 and use it for annotations

### LINKED ISSUE #2049: Support LaTeX-style formulas in annotations
LaTeX-style formulas (`$...$`, `$$...$$`, `\[...\]`) in annotations are not rendered because MathJax is explicitly disabled on feedback pages (including the code page).

LucianosLionakis:
How it now looks:
image

Todo:

- [x] Upgrade to MathJax 3:
   - [x] ~~MathJax added as a package~~
   - [x] ~~Allow sandbox to request the mathjax fonts (.woff files) from dodona.~~
- [x] Add mathjax functionality to the annotations
   - [x] Works after edit
   - [x] Works after creation
   - [x] Works when page is first loaded
   - [x] Works now in the exercise page when looking at a submission
- [x] Add code to log an error when MathJax is undefined

Closes #2049.


app/assets/javascripts/code_listing/annotation.ts
```diff
@@ -89,6 +89,7 @@ export abstract class Annotation {
             // Body.
             this.__html.appendChild(this.header);
             this.__html.appendChild(this.body);
+            window.MathJax?.typeset();
```
rien, line 92: I think we want to log an error if `MathJax` isn't present in our current environment.

app/views/activities/description.html.erb
```diff
@@ -26,7 +29,4 @@
       <% end %>
     </footer>
   </div>
-</div>
-<script>
-  $(function(){window.dodona.initDescription();});
-</script>
+</div>
```
rien, line 32: Why was this change needed? Are you sure the description is still initialized correctly?

app/assets/javascripts/submission.js
```diff
@@ -86,4 +86,32 @@ function contextualizeMediaPaths(parentClass, exercisePath, token) {
     });
 }
 
+function initMathJax() {
```
rien, line 89: This is an exact copy of the `initMathJax` function in `exercise.js`.

LucianosLionakis: Almost, the same. The one in the `exercise.js` ignores the `feedback-table` class while the one in `submission.js` doesn't.

rien: My bad. Why the difference?

LucianosLionakis: Well, it was initially configured with the ignore but this prevented the annotations from rendering correctly since they are in the `feedback-table`. I didn't remove it from de `exercise.js` since I don't know if this was added for some specifiek reason.

app/views/activities/description.html.erb
LucianosLionakis: Localy it seemed to work correctly. I had to make two changes because otherwise MathJax didn't load correctly.
First I had to move this function above the MathJax import on [line 6](https://github.com/dodona-edu/dodona/pull/2146/files/c38f9a91e849767ca51fb8df20fba7d8c870a84a#diff-7b122ec4852c0cc33a99fdc8cdaee956R6) this sadly was not enough because of the `$()`. So the second edit I made was to remove this.

rien: You're initializing the description now before the description itself is parsed (the `$()` is executed once the document is fully loaded). I would move the script and MathJax loading back to the bottom of the page.

One way to test if this is working properly, is by slowing down your network speed artificially in the networking tab of your developer tools. If everything still works (MathJax is loaded, images are centered and have the lightbox is working) then it will probably be fine.

app/views/activities/description.html.erb
```diff
@@ -28,5 +27,6 @@
   </div>
 </div>
 <script>
-  $(function(){window.dodona.initDescription();});
+  window.dodona.initDescription();
 </script>
+<script id="MathJax-script" src='https://cdn.jsdelivr.net/npm/mathjax@3/es5/tex-mml-chtml.js'></script>
```
bmesuere, line 32: Is there a reason why you moved this to the bottom while it was in the head previously? By putting it here, mathjax will probably fail on the first pageload because it's loaded after use.

LucianosLionakis: > Is there a reason why you moved this to the bottom while it was in the head previously? By putting it here, mathjax will probably fail on the first pageload because it's loaded after use.

It need to be loaded after use otherwise it doesn't work. That how it was suggested by the MathJax converter and their docs. As I understand it you first need to set the MathJax configuration and then import their script.


> I might have missed it, but how do you prevent it from modifying code/test descriptions containing `$`'s?

This case is handled in the configuration in by [lines 123 to 124](https://github.com/dodona-edu/dodona/pull/2146/files#diff-1991fc3cd2f2a7973aab6e222c4d436bR123-R124). Firstly line 123 says that the `feedback-table` should not be processed by MathJax. Then line 124 says that in we want that anything containing `annotaion user` class should be processed even if it is in the `feedback-table`

bmesuere: > As I understand it you first need to set the MathJax configuration and then import their script. The reason it is at the bottom of the page is that the initialization of MathJax was there and @rien suggested to leave the description initialization where it is and not move it at the top of the page.

Yes, but in feedback/show.html.erb both are set in the head (using `content_for :javascripts`)

LucianosLionakis: > Yes, but in feedback/show.html.erb both are set in the head (using content_for :javascripts)

Yes, because the `feedback/show.html.erb` had no restriction to where MathJax should be initialized. I forgot to mention my justification in the case of `activities/description.html.erb`, MathJax was imported at the bottom because of fear that something would break if the description initialization was moved to the top.

app/views/pages/home.html.erb
```diff
@@ -7,7 +7,7 @@
       <%= render "getting_started_card" %>
       <%= render 'privacy_disclaimer' %>
     <% else %>
-      <div class="row favorites-row">
+      <div class="card-columns favorites-card-columns">
```
bmesuere, line 10: this probably shouldn't be in this PR

PR #2148, opened 2020-08-06
Added missing pythia pack to the feedback view

### LINKED ISSUE #2051: Python Tutor disabled in evaluations
The Python Tutor can't be used while evaluating solutions.

LucianosLionakis:
This pull request fixes the python tutor during feedback process of an evaluation.

Closes #2051.

### DIFF
--- a/app/views/feedbacks/show.html.erb
+++ b/app/views/feedbacks/show.html.erb
@@ -1,5 +1,6 @@
 <% content_for :javascripts do %>
   <%= javascript_pack_tag 'submission' %>
+  <%= javascript_pack_tag 'pythia_submission' if @feedback.submission.judge.renderer == PythiaRenderer %>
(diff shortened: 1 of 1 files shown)


app/views/feedbacks/show.html.erb
```diff
@@ -73,4 +74,4 @@
       $(".nav.nav-tabs li:last-child a").tab("show");
     });
   </script>
-<% end %>
+<% end %>
```
rien, line 77: Your editor seems to remove trailing newlines, can you turn this off and restore the newline? It is adding a bit of unneeded clutter to the PR.

LucianosLionakis: That's weird because the option is already disabled O.o, so I probably did that...

PR #2150, opened 2020-08-07
Improve about page

### LINKED ISSUE #1656: About page
We should have page with more information about Dodona and the team behind it (what, who, why, ...).

LucianosLionakis:
This pull request enriches the Dodona about page.

Currently:
image
image
image

Closes #1656 .


LucianosLionakis:
> For such text-heavy pages, translating the entire page/partial might be a better option: https://guides.rubyonrails.org/v4.2/i18n.html#localized-views

Okay, I am on it. Is the content sufficient or do you have any suggestions of things I could/should add/edit?

bmesuere:
This is a good first start, but the copy should be improved. I would leave out the second card and expand the last one. I don't have time to rewrite this myself right now, but maybe @pdawyndt has some suggestions.

niknetniko:
Superseded by #2372.

PR #2151, opened 2020-08-10
Added light-only and dark-only css classes to hide when needed

### LINKED ISSUE #1397: Add classes to show elements only in light/dark mode
We may consider inverting images while in dark mode. One option using CSS:

```css
img {
   -webkit-filter: invert(1);
   filter: invert(1);
}
```

Don't know if this would work better for all images, but at least it turns the ISBN image

image

into

image

Might even look better if I make the background of the image transparant.

LucianosLionakis:
This pull request add the `light-only` and `dark-only` css classes that can be used to hide elements in dark and light mode. As the names suggest `light-only` will only be visible in light mode while `dark-only` will only be visible in dark mode.

- [x] Added documentation at [dodona-edu/dodona-edu.github.io#69](https://github.com/dodona-edu/dodona-edu.github.io/pull/69)

Closes #1397.

### DIFF
--- a/app/assets/stylesheets/theme/theme-dark.css.less
+++ b/app/assets/stylesheets/theme/theme-dark.css.less
@@ -111,6 +111,6 @@
     }
 
     .dark-only {
-        display: block;
+        display: unset;
     }
 }
--- a/app/assets/stylesheets/theme/theme-dark.css.less
+++ b/app/assets/stylesheets/theme/theme-dark.css.less
@@ -105,12 +105,12 @@
 
 @table-hover-bg: @body-bg;
 
+.activity-description {
+    .light-only {
+        display: none;
+    }
 
-.light-only {
-    visibility: hidden;
-    display: none;
+    .dark-only {
+        display: block;
+    }
 }
-
-.dark-only {
-    visibility: visible;
-} 
--- a/app/assets/stylesheets/theme/theme-dark.css.less
+++ b/app/assets/stylesheets/theme/theme-dark.css.less
@@ -104,3 +104,13 @@
 @punchcard-fg: @light-blue-900;
 
 @table-hover-bg: @body-bg;
+
+
+.light-only {
+    visibility: hidden;
+    display: none;
+}
+
+.dark-only {
+    visibility: visible;
+} 
--- a/app/assets/stylesheets/theme/theme-light.css.less
+++ b/app/assets/stylesheets/theme/theme-light.css.less
@@ -177,6 +177,6 @@
     }
 
     .light-only {
-        display: block;
+        display: unset;
     }
 }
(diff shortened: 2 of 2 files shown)


app/assets/stylesheets/theme/theme-dark.css.less
```diff
@@ -103,4 +103,14 @@
...
\ No newline at end of file
+@table-hover-bg: @body-bg;
+
+
+.light-only {
+    visibility: hidden;
+    display: none;
+}
+
+.dark-only {
```
rien, line 114: Why is there no equivalent of this in `theme-light.css.less`?

LucianosLionakis: Because the `theme-dark.css.less` imports `theme-light.css.less` so `.dark-only` needs to be overwritten. Since `theme-light.css.less` does not import `theme-dark.css.less` the `.light-only` doesn't have to be overwritten. All that aside, do you think would be better to add this in the `theme-light.css.less` for clarity purposes?

bmesuere:
Documentation can be added to this page under images similar to Lightboxes and centered groups https://dodona-edu.github.io/nl/references/exercise-description/#lightboxes + to the English version of the page.

LucianosLionakis:
> Why use both visibility and display? I think display alone should be enough.

Yes, you are right display is enough. So I removed the visibility

> Please also scope in exercise description

Okay, I did this using the `activity-description` class.

app/assets/stylesheets/theme/theme-dark.css.less
rien: Ok, nothing has to change. I was asking out of curiosity.

LucianosLionakis:
> Documentation can be added to this page under images similar to Lightboxes and centered groups https://dodona-edu.github.io/nl/references/exercise-description/#lightboxes + to the English version of the page.

Just opened a [pull request](https://github.com/dodona-edu/dodona-edu.github.io/pull/69) adding the documentation in both English and Dutch.

app/assets/stylesheets/theme/theme-dark.css.less
```diff
@@ -103,4 +103,14 @@
...
\ No newline at end of file
+@table-hover-bg: @body-bg;
+
+.activity-description {
+    .light-only {
+        display: none;
+    }
+
+    .dark-only {
+        display: block;
```
bmesuere, line 114: This will break if the element on which it is applied is not a block element (eg inline span). Can you verify this? Potential alternatives might be "unset" or "initial".

LucianosLionakis: Yes, are right some parts do break (even images aren't centered anymore) because of setting display to block. It seems that both `initial` and `unset` work fine but the description of `unset` seemed more general so I changed it to that.

pdawyndt:
Included separate dark/white images in [this exercise](https://dodona.ugent.be/nl/activities/2066754027/) and it works, apart from the fact that both images are picked up by the carrousel showing all images on the page.

PR #2157, opened 2020-08-11
Improve homepage so that cards fit nicely 

### LINKED ISSUE #1551: homepage cards do not fit nicely
@pverscha mentioned the course cards on his homepage were shown in a weird way:

image

The 'Programming' course card should ideally fit closer to the card above, like the 'Computergebruik' card does.

LucianosLionakis:
This pull request fixes a problem were homepage cards did not fit nicely.

Trying to fix this:

Currently is like this:
image

- [x] Need to trigger masonry when favorite courses are altered.
- [x] Do not allow cards to fill the whole row if they are alone in a two-column setup. 
- [x] Change the number of columns if the screen size changes (just like before).
Closes #1551.


app/assets/javascripts/masonry.ts
```diff
@@ -0,0 +1,84 @@
...
+
+    constructor() {
+    // subscribe to load and resize events
+        window.addEventListener("load", () => this.onLoad());
+        window.addEventListener("resize", () => this.onResize());
+    }
+
+    onLoad(): void {
+        const rootElements = document.getElementsByClassName("masonry-root");
+        this.roots = Array.prototype.map.call(rootElements, (rootElement: HTMLElement) => {
```
rien, line 20: Is there a reason why `rootElements.map(...)` isn't used here?

app/assets/javascripts/masonry.ts
```diff
@@ -0,0 +1,84 @@
...
+            return { element: rootElement, columnNumber: 0, cells: cells };
+        });
+        // do the first layout
+        this.onResize();
+    }
+
+    onResize(): void {
+        for (const root of this.roots) {
+            // only layout when the number of columns has changed
+            const newColumnNumber = window.innerWidth > 1024 ? 2 : 1;
```
rien, line 39: I would extract `1024` to a class constant and make sure this value agrees with the bootstrap change points.

app/assets/stylesheets/pages/home.css.less
```diff
@@ -161,3 +161,17 @@
...
+
+.masonry-root {
+  display: flex;
+  flex-direction: column;
+  flex-wrap: wrap;
+  .masonry-cell {
+    flex: 1;
+    margin-left: 4px;
+    margin-right: 4px;
+    @media (min-width: 1023px){
```
rien, line 173: I think the bootstrap `lg` class (where we start using two columns for our cards) uses `min-width: 1200px`.

thepieterdc: `lg` is `1200px` indeed. However, rather than hardcoding this, I'm sure that has been defined as a variable in the bootstrap source, which you can simply import.

`@screen-lg-min` probably

app/assets/javascripts/masonry.ts
LucianosLionakis: Not that I know of. I will change it and see if it has any effect (which it probably won't).

LucianosLionakis: Apparently the `HTMLCollectionOf<>`  type has no `map` property.

LucianosLionakis:
> Can you also try to restore the padding/margin between the cards? I think this is called the "gutter" in bootstrap.

I tried using padding-left and -right in the CSS of the masonry-items class. This leads to larger paddings around the course card listing as seen below (the red arrows indicate the problem).
image

I tried finding out how bootstrap handles the padding of columns and rows. Unfortunately, I didn't find anything helpful so I decided to solve this problem in the masonry script since it is easy to find the cards in the first and last columns and give them the appropriate padding.

I edited the description to reflect the current state of changes.

bmesuere (changes requested):
I don't think a homepage pack + binding to the global context is needed. Instead, 
- import Masonry inside favorite_course_buttons just like the toast
- rename `initFavoriteButtons` to something more general (initHomePageCards?) and do the initialization that's in home.html.erb over there

app/assets/javascripts/masonry.ts
```diff
@@ -0,0 +1,90 @@
...
+
+    roots: CustomElement[];
+
+    constructor() {
+    // subscribe to load and resize events
+        window.addEventListener("load", () => this.onLoad());
+        window.addEventListener("resize", () => this.onResize());
+    }
+
+    onLoad(): void {
```
bmesuere, line 20: Rename `onLoad` and `onResize` to what it does and not when it is called. It doens't make sense to call a function named `onLoad` when adding a favorite, or calling `onResize` at the end of `onLoad`

app/assets/javascripts/masonry.ts
bmesuere: I don't like this construction. If it stays in the code, a comment should be added. Can't it be solved by using `Array.from(rootElements).map` ?

app/assets/javascripts/masonry.ts
```diff
@@ -0,0 +1,90 @@
...
+        const rootElements = document.getElementsByClassName("masonry-root");
+        this.roots = Array.prototype.map.call(rootElements, (rootElement: HTMLElement) => {
+            const cellElements: HTMLCollectionOf<Element> = rootElement.getElementsByClassName("masonry-cell");
+            const cells: CustomElement[] = Array.prototype.map.call(cellElements, (cellElement: HTMLElement) => {
```
bmesuere, line 24: same here

app/assets/javascripts/masonry.ts
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    onResize(): void {
+        for (const root of this.roots) {
+            // only layout when the number of columns has changed
+            const newColumnNumber = window.innerWidth >= Masonry.screenLgMin ? 2 : 1;
+            if (newColumnNumber != root.columnNumber) {
+                // initialize
+                root.columnNumber = newColumnNumber;
+                const columns = Array.from(new Array(root.columnNumber)).map( () => {
```
bmesuere, line 45: What is the goal of this `Array.from(new Array(root.columnNumber))` construct?

app/assets/javascripts/masonry.ts
```diff
@@ -0,0 +1,90 @@
...
+                    const rootElement: CustomElement = { outerHeight: 0 };
+                    rootElement.cells = [];
+                    return rootElement;
+                });
+
+                // divide...
+                for (const cell of root.cells) {
+                    const minOuterHeight = Math.min(...columns.map(column => column.outerHeight));
+                    const column = columns.find(column => column.outerHeight == minOuterHeight);
```
bmesuere, line 54:
```suggestion
                    const column = columns.find(column => column.outerHeight === minOuterHeight);
```

app/assets/javascripts/masonry.ts
```diff
@@ -0,0 +1,90 @@
...
+                        cell.element.style.order = String(order++);
+                        // set the cell's flex-basis to 0
+                        cell.element.style.flexBasis = "0";
+                    }
+                    // set flex-basis of the last cell to fill the
+                    // leftover space at the bottom of the column
+                    // to prevent the first cell of the next column
+                    // to be rendered at the bottom of this column
+                    if (column.cells.length !== 0) {
+                        column.cells[column.cells.length - 1].element.style.flexBasis = String(column.cells[column.cells.length - 1].element.offsetHeight + masonryHeight - column.outerHeight - 1) + "px";
```
bmesuere, line 75: use template string (with backticks) instead of wrapping it in String and using a +

app/assets/javascripts/masonry.ts
```diff
@@ -0,0 +1,90 @@
...
+                    }
+                }
+
+                // set the masonry height to trigger
+                // re-rendering of all cells over columns
+                // one pixel more than the tallest column
+                root.element.style.maxHeight = String(masonryHeight) + "px";
```
bmesuere, line 82: template string

app/assets/javascripts/masonry.ts
```diff
@@ -0,0 +1,90 @@
...
+                if (root.element.className.includes("favorite")) {
```
bmesuere, line 83: these logs probably shouldn't be here?

bmesuere (changes requested):
👍 the cards now flow as expected and this way of importing is cleaner. I have few small remarks about the code. Since this js file and pack are only used on the homepage, it might be better to rename them from `favorite_course_buttons.js` to `home_page.js`

app/assets/javascripts/masonry.ts
```diff
@@ -0,0 +1,90 @@
...
+        for (const root of this.roots) {
+            // only layout when the number of columns has changed
+            const newColumnNumber = window.innerWidth >= Masonry.screenLgMin ? 2 : 1;
+            if (newColumnNumber != root.columnNumber) {
+                // initialize
+                root.columnNumber = newColumnNumber;
+                const columns = Array.from(new Array(root.columnNumber)).map( () => {
+                    const rootElement: CustomElement = { outerHeight: 0 };
+                    rootElement.cells = [];
+                    return rootElement;
```
thepieterdc, line 48: What happens if you do `return { cells: [], outerHeight: 0}` (can subsequently remove the return as well to truly oneline it)

app/assets/javascripts/masonry.ts
```diff
@@ -0,0 +1,84 @@
...
+        });
+        // do the first layout
+        this.setCellLayout();
+    }
+
+    setCellLayout(): void {
+        for (const root of this.roots) {
+            // only layout when the number of columns has changed
+            const newColumnNumber = window.innerWidth >= Masonry.screenLgMin ? 2 : 1;
+            if (newColumnNumber != root.columnNumber) {
```
bmesuere, line 42:
```suggestion
            if (newColumnNumber !== root.columnNumber) {
```

bmesuere (approved):
looks good 👍

PR #2164, opened 2020-08-13
Add language and labels to the activities index

### LINKED ISSUE #2163: Improve exercise listing
Right now, we have 3 exercise listings:
- on the page to add exercises to a series, we display extended info about the exercise (programming language, labels, etc)
- on the all exercises page, we display only the name and the number of people who solved the exercise
- on the course page itself within a series

I think the last two use the same partial, but I might be wrong.

The general listing (second one in the list) should be expanded to look more like the first one. The last one should stay as it is, so if it uses the same partial, that should be split into a separate one.

We want to display all information from the first type (maybe limit repo and directory to users with staff permissions) and the current information about how many users solved it and your personal status.

winniederidder:
This pull request updates the activity listings.

For comparison: the upper image is the original series edit. Below is the new activities index page with now more data added.
series-edit
activities-index

The labels, repository names etc. are clickable. For this I created an activities pack and moved initTokensClickable from series to util as it is no longer series only.

Closes #2163  .


bmesuere (commented):
I would remove the "labels" (and potentially "languages") header since the column contains both the programming language and the labels.

app/javascript/packs/activities.js
```diff
@@ -0,0 +1,3 @@
+import { initTokenClickables } from "util.js";
+
+window.dodona.initTokenClickables = initTokenClickables;
```
bmesuere, line 3: Is there a reason to bind this to the global dodona variable?

bmesuere:
The activities index view should be set to use 12 columns instead of 10 right now.

app/javascript/packs/activities.js
winniederidder: I did what other packs did. I'll see for a different way.

winniederidder:
activities-index
Changed to 12 columns and removed labels header

app/views/activities/index.html.erb
```diff
@@ -1,5 +1,5 @@
 <div class="row">
-  <div class="col-sm-10 col-sm-offset-1 col-xs-12">
+  <div class="col-sm-12 col-sm-offset-1 col-xs-12">
```
bmesuere, line 2:
```suggestion
  <div class="col-xs-12">
```

bmesuere (approved):

PR #2175, opened 2020-08-19
Allow exercises to be filtered by judge and link to them from the judge/show view

### LINKED ISSUE #2166: List exercises on judge page
The judge page mentions the total number submissions for this judge (and links to those submissions), but could also mention the total number of exercises for this judge (and link to those exercises).

winniederidder:
This pull request adds an exercise link to the judges page and allows filtering activities by used judge.

Judge page:
judge-show

Following the link:
judge-label

Filtering by judge:
judges-searchbar

Regular users cannot use this feature, only staff and above:
judges-filtered-student

- [x] Tests were added

Closes #2166 .


db/seeds.rb
```diff
@@ -138,14 +138,14 @@
 
   puts 'Create & clone judge'
 
-  python_judge = Judge.create name: 'python', image: 'dodona-python', remote: 'git@github.com:dodona-edu/judge-pythia.git', renderer: PythiaRenderer
+  python_judge = Judge.create name: 'python-judge', image: 'dodona-python', remote: 'git@github.com:dodona-edu/judge-pythia.git', renderer: PythiaRenderer
```
rien, line 141: The judges used in Dodona right now don't have 'judge' in their name. So I would keep it like this in the seeds. See my comment in `_searchbar.html` for a solution.

app/views/layouts/_searchbar.html.erb
```diff
@@ -132,11 +132,23 @@
...
         <% end %>
-
+        <% if local_assigns.fetch :judges, false %>
+        filterCollections["judges"] = {
+            param: "judge_id",
+            paramVal: function(judge) {
+                return judge.id;
+            },
+            multi: false,
+            data: <%= raw local_assigns.fetch(:judges, []).map{|judge| {id: judge.id, name: judge.name}}.to_json %>,
```
rien, line 146:
```suggestion
            data: <%= raw local_assigns.fetch(:judges, []).map{|judge| {id: judge.id, name: "#{judge.name}-judge"}}.to_json %>,
```

app/models/activity.rb
```diff
@@ -77,6 +77,7 @@ class Activity < ApplicationRecord
   scope :by_labels, ->(labels) { includes(:labels).where(labels: { name: labels }).group(:id).having('COUNT(DISTINCT(activity_labels.label_id)) = ?', labels.uniq.length) }
   scope :by_programming_language, ->(programming_language) { includes(:programming_language).where(programming_languages: { name: programming_language }) }
   scope :by_type, ->(type) { where(type: type) }
+  scope :by_judge, ->(judge_id) { where(judge_id: judge_id) }
```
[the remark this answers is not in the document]
winniederidder: Does this work for the has_scope aswell, where the judge_id is present in the url parameter?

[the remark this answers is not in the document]
rien: Yes, if you pass an integer it will use that as the id.

rien (approved):

app/controllers/activities_controller.rb
```diff
@@ -16,6 +16,7 @@ class ActivitiesController < ApplicationController
   has_scope :by_type, as: 'type'
   has_scope :in_repository, as: 'repository_id'
   has_scope :by_description_languages, as: 'description_languages', type: :array
+  has_scope :by_judge, as: 'judge_id', if: ->(this) { Pundit.policy!(this.current_user, Judge).index? }
```
bmesuere, line 19: I'm not sure if we want this policy check here?

bmesuere (changes requested):

app/controllers/activities_controller.rb
rien: I don't know if there is a better place? We _could_ just omit the check, because it's not really an issue if students can request activities for a specific judge.

bmesuere: Sorry if I wasn't clear. My proposal was indeed to remove this check here (since it just defines a scope) and simply not list the judges in the autocomplete for normal users. The reason not to show it is because it isn't relevant for them and would only add confusion not because they shouldn't be allowed to see it.

rien: The judge search filter won't show up for students because the `policy_scope` is empty for them. Or do you also want to hide this filter for teachers?

bmesuere: The judge/index view is visible to zeus+staff so it should probably be hidden for course admins without staff privileges?

rien: Ah yes, I meant staff. There won't be any judge filter unless `user.admin?` is true. If there are no judges to filter on, the autocomplete won't show them. We could also explicitly disable that behavior by placing a policy check around the `filterCollection["judges"]` call in [the searchbar](https://github.com/dodona-edu/dodona/pull/2175/files#diff-608825a8bb59cd289c36a2c1510cd4a5R140).

winniederidder: Does this imply the policy scope change for judges should also be reverted?

bmesuere (approved):

Issue #737, opened 2019-08-01
LTI support


pdawyndt:
Add LTI support to Dodona.

Official documentation: https://documentation.brightspace.com/EN/integrations/ipsis/LTI%20Advantage/intro_to_LTI.htm?tocpath=Administrators%7CIntegrations%7CLTI%20Advantage%7C_____0

LTI Advantage is the next iteration of LTI, consisting of a new launch protocol and a package of optional extensions:

* Launch: The LTI 1.3 core launch flow follows a more modern security approach by adopting OpenId Connect (OIDC) specifications, OAuth2 and JSON Web Tokens.
* Assignment and Grades Services v2.0 (AGS): A replacement for the Basic Outcome service. AGS allows the LTI tool to create and manage grade objects and exchange scores and instructor comments. AGS includes additional capabilities two-way gradebook integration, where the LTI tool can request updates to grade values linked to the tools Assignments; allowing the Brightspace gradebook and tool to stay in sync.
* Names and Roles Provisioning Services v2.0: The service that allows the LTI tool to request a list of participants in a course offering in a safe and secure fashion. Typically used to repopulate the classlist in the tool.
* Deep Linking 2.0: A replacement for Content Item Message (CIM) service. Provides an intuitive way to add content from a publisher directly into Brightspace content, creating a seamless learner experience.

_[Original issue](https://github.ugent.be/dodona/dodona/issues/737) by @pdawyndt on Thu Oct 18 2018 at 14:57._

pdawyndt:
DICT has managed to link Curios to Ufora. We could ask if this was done through 
LTI or directly through the Brightspace API.

pdawyndt:
Ufora-interface for setting up LTI connections with external tools (provided by Patrick Cool from DICT)

![LTI_ufora](https://user-images.githubusercontent.com/5736113/69233084-89982d00-0b8c-11ea-81ce-14c94ba5c51e.png)

pdawyndt:
Veerle Fack on integration of Dodon in external learning platforms (from MS Teams):

> Peter Dawyndt Ik denk dat er toch eens werk moet van gemaakt worden om Dodona ook in Ufora te integreren. Wat we misschien allemaal nog als minpuntjes aan Bongo VC gaan ontdekken, het feit dat het in Ufora geintegreerd is, maakt het wel erg handig: studenten zitten er automatisch in, ingeplande sessies komen ook automatische in agenda, e.d.

rien:
[...]

PR #2177, opened 2020-08-23
LTI Content Selection

thepieterdc:
This PR adds LTI Deep Linking support with Content Selection.

**Tests:**
- Moodle (https://sandbox.moodledemo.net) and local instance
- Brightspace via Ufora (https://ufora.ugent.be); can only be done by deploying to naos.

**Roadmap:**
- [x] Select 1 activity on Moodle
- [x] Select 1 activity on Brightspace
- [x] Improve layout for content selection (#2231)
- [x] Allow multiple items to be selected (#2231)
- [x] Fix cookies (see below)
- [ ] Decide if we want analytics in the iframes (https://github.com/dodona-edu/dodona/pull/2214#discussion_r486623107)

See #2214 for the UI when logging in for the first time.

**Remarks:**
We need to verify that the SameSite policy of our cookies is correct. Because of an update in Chrome, the security for cookies in iframes has been tightened and this might break content selection. Can be temporarily disabled in `chrome://flags` by disabling everything related to SameSite cookies.

Grouping issue: #737


app/controllers/auth/omniauth_callbacks_controller.rb
```diff
@@ -5,10 +5,15 @@ class Auth::OmniauthCallbacksController < Devise::OmniauthCallbacksController
...
-    redirect_to root_path
+    # Find the error message and log it for analysis.
+    error_message = failure_message ||
+                    request.params['error_description'] ||
+                    I18n.t('devise.omniauth_callbacks.unknown_failure')
+    logger.error error_message
+
+    # Show a flash message and redirect.
+    flash_failure error_message
+    redirect_to root_path, status: :bad_request
```
niknetniko, line 16: @thepieterdc What does the `status: :bad_request` do? I ask since it breaks the tests

charvp (changes requested):
There's quite a lot of duplication between the top-level layout erb files. Can we extract some of that to partials?

Otherwise looks good to me, but I'm not familiar with the LTI standard, so I didn't really check for implementation mistakes there. I assume that since things are working on naos, we're good in that department though.

niknetniko:
> There's quite a lot of duplication between the top-level layout erb files. Can we extract some of that to partials?

I intended to rename "embedded" to "lti", but somehow ended up with both. I've removed the "embedded" one (they were identical).

charvp:
> > There's quite a lot of duplication between the top-level layout erb files. Can we extract some of that to partials?
> 
> I intended to rename "embedded" to "lti", but somehow ended up with both. I've removed the "embedded" one (they were identical).

I do think there's still some duplication between `application.html.erb` and `lti.html.erb` (and maybe `frame.html.erb`). Can you look into removing some of that? (It might not be worth it, I'll leave that up to your judgment.)

niknetniko:
> I do think there's still some duplication between `application.html.erb` and `lti.html.erb` (and maybe `frame.html.erb`). Can you look into removing some of that? (It might not be worth it, I'll leave that up to your judgment.)

I've extracted some more common code into partials. I think the remaining code is either different or too small (a few lines) to extract into a partial.

charvp (approved):

bmesuere (approved):

PR #2187, opened 2020-08-27
Fix/math jax crash on feedback

LucianosLionakis:
This pull request fixes a crash when feedback was present for a submission.


winniederidder:
I wouldn't start from the bot-branch but from develop ;)

PR #2189, opened 2020-08-27
Fix MathJax

LucianosLionakis:
This pull request fixes an error when feedback is present for a submission. To fix this the initialization of MathJax was moved inside the feedback_code_rendere. 

Previously:
image

Now:
image


bmesuere (changes requested):
The promise is a new global variable, can't we bind to the dodona var?

bmesuere (approved):

PR #2214, opened 2020-09-03
[LTI] don't login via iframe

niknetniko:
The flow happening here is:

1. We know server side we are not in an iframe. In this case we redirect to the main provider like normal, and redirect to the target happens like normal (no changes).
2. We don't know if we are in an iframe (Firefox & safari) or we are in an iframe. In this case we redirect to a web page.
3. On the web page, we do another check to see if we are in an iframe.
   a. If we are not in an iframe, we redirect immediately to the next step, with `browser = true`.
   b. If we are in an iframe, we display a button the user has to click, which will open the next step, with `browser = false`.
4. If this is the first time we hit the page, we redirect to the main provider, which redirect to this page after authentication.
5. If this is not the first time we hit this page,  this means we returned from the main provider. In this case, we redirect to the original target if we are not in an iframe, but show a page if we are in an iframe.
	
A lot of the complexity comes from the case where we don't know if it is an iframe on the server, but is is in fact not an iframe. In this case we simulate the normal process, for which we need to save the original URL before all redirection.

This flow should work for all LTI logins, independent of where they come from (e.g. if a platform opens Dodona in an iframe, this should also apply).

Tip: testing locally is a lot easier if SAML works, use https://capriza.github.io/samling/samling.html and create a new SAML provider for it.

### Visuals ("Select content" title is part of Moodle)
image

image


config/routes.rb
```diff
@@ -21,6 +21,9 @@
       end
 
       get '/users/saml/metadata' => 'saml#metadata'
+
+      get '/users/lti/redirect', to: 'redirect#redirect'
+      get '/users/lti/do_redirect', to: 'redirect#do_redirect'
```
thepieterdc, line 26: A bit confused about the inconsistency; there's a notion of `LTI` in the route, but not in the name of the controller. I'd suggest going a step further and removing the `user` part as well. Move these methods to the `LtiController`, since they are tied to LTI.

```suggestion
      get '/lti/redirect', to: 'lti#redirect'
      get '/lti/do_redirect', to: 'lti#do_redirect'
```

config/locales/views/auth/nl.yml
```diff
@@ -14,3 +14,10 @@ nl:
       errors:
         institution-creation: Er ging iets mis tijdens het aanmaken van een nieuwe onderwijsinstelling.
         missing-provider: Enkel accounts die gelinkt zijn aan een onderwijsinstelling worden ondersteund.
+    redirect:
+      first_time: Dit is de eerste keer dat je bij Dodona aanmeldt met een leerplatform. Daarom moet je nu eenmalig aanmelden in een nieuw venster.
```
thepieterdc, line 18:
```suggestion
      first_time: Aangezien dit de eerste keer is dat je je aanmeldt bij Dodona via dit leerplatform, dien je eenmalig aan te melden in een nieuw venster.
```

config/locales/views/auth/nl.yml
```diff
@@ -14,3 +14,10 @@ nl:
...
+      login: Aanmelden in een nieuw venster
+      after: Na het aanmelden moet je dit venster sluiten (en opnieuw openen).
```
thepieterdc, line 20:
```suggestion
      after: Nadat je aangemeld bent, moet je dit venster sluiten en de actie opnieuw uitvoeren.
```

thepieterdc (changes requested):
Tested and verified using the samling tool, which is very cool! :+1: 

Added some remarks and updated the description with visuals.

PR #2230, opened 2020-09-07
[LTI] Some clean-up

niknetniko:
- As discussed, we save the whole JWT in the selection HTML, enabling us to access the data and send it back (currently the `nonce`, the `data` and the `deployment_id`). The deployment ID was hardcoded, but I assume it can change.
- I've included a bunch more claims that are used / can be used / will be used in the content selection UI, like the supported item types, the language of the user, etc.
- I've attempted to reduce some code duplication by introducing more classes in the LTI Message namespace, but not entirely sure if this is the Ruby way.

TODO:
- [ ] Test on Naos to see nothing breaks


thepieterdc:
Merging this into #2177

thepieterdc (approved):

PR #2231, opened 2020-09-07
[LTI] Content selection UI

niknetniko:
Depends on  #2214. 

- If the platform does not support `LtiResourceLink`s, we display that we don't support any content. In the future, we might expand to also support links or other content types.
- Multiple selection is implemented, but only for activities, not for courses or series. This because we plan to rewrite this view in Vue in the future, where we'll have a better UI to support this.
- Technically: all courses are loaded, and the series & activities for one courses are inserted when the course is selected (using the same technique as the search pages).


app/assets/javascripts/lti.ts
```diff
@@ -6,22 +6,105 @@ const activitySelectId = "lti_content_selection_activity";
 const courseSelectId = "lti_content_selection_course";
 const seriesSelectId = "lti_content_selection_series";
 
+function showElement(bar: HTMLElement): void {
+    bar.style.visibility = "visible";
+}
+
+function hideElement(bar: HTMLElement): void {
+    bar.style.visibility = "hidden";
+}
```
thepieterdc, line 15: These functions are not LTI specific and can therefore be refactored to a more general util file

app/assets/javascripts/lti.ts
```diff
@@ -6,22 +6,105 @@ const activitySelectId = "lti_content_selection_activity";
...
+function activateSelect(container: HTMLElement | null): void {
+    if (!container) {
+        return;
+    }
+    container.classList.remove("hidden");
+    container.classList.add("active");
+    const element = container.getElementsByTagName("select")[0];
+    element.id = activitySelectId;
+}
+
```
thepieterdc, line 36: Same comment as above

app/assets/javascripts/lti.ts
```diff
@@ -6,22 +6,105 @@ const activitySelectId = "lti_content_selection_activity";
...
+        const id = (event.target as HTMLSelectElement).value;
+        seriesContainer.innerHTML = "";
+
+        if (id === "") {
+            hideElement(confirmButton);
+            return;
+        }
+        showElement(confirmButton);
+        showElement(progressBar);
+        const url = new URL(baseUrl);
```
thepieterdc, line 58: Didn't know about this, cool!

app/assets/javascripts/lti.ts
```diff
@@ -6,22 +6,105 @@ const activitySelectId = "lti_content_selection_activity";
...
+        url.searchParams.append("multiple", multiple.toString());
+
+        const response = await fetch(url, {
+            headers: {
+                "accept": "text/javascript",
+                "x-csrf-token": $("meta[name=\"csrf-token\"]").attr("content"),
+                "x-requested-with": "XMLHttpRequest",
+            },
+            credentials: "same-origin",
+        })
```
thepieterdc, line 70: I might be wrong, but I believe that the Dodona `fetch` implementation already does this (`assets/javascripts/util.js` ~ line 160). Optionally add the `same-origin` there using a parameter or so

app/assets/javascripts/lti.ts
```diff
@@ -6,22 +6,105 @@ const activitySelectId = "lti_content_selection_activity";
...
+            .then(resp => resp.text())
+            .then(data => {
+                eval(data);
+                hideElement(progressBar);
+            });
+        console.log(response);
+        console.log(`Selected ${id}`);
```
thepieterdc, line 77: Remove console logs

app/controllers/lti_controller.rb
```diff
@@ -11,25 +11,26 @@ class LtiController < ApplicationController
...
   def content_selection
-    # TODO: For testing purposes, the course, series and activity are chosen at
-    #       random. This should be replaced by a form that allows the user to
-    #       select something themself.
-    @series = Series.first
-    @course = @series.course
-    @activity = @series.activities.first
+    @supported = @lti_message.accept_types.include?(LTI::Messages::Types::DeepLinkingResponse::LtiResourceLink::TYPE)
+    courses = policy_scope(Course.all)
+    @grouped_courses = courses.group_by(&:year)
```
thepieterdc, line 16: Not using the `courses` variable (local scope), since it's not too long I'd just inline it

```suggestion
    @grouped_courses = policy_scope(Course.all).group_by(&:year)
```

app/controllers/lti_controller.rb
```diff
@@ -47,4 +48,44 @@ def set_lti_message
   def set_lti_provider
     @provider = Provider::Lti.find(params[:provider_id])
   end
+
+  def create_items_from_params(params)
```
thepieterdc, line 52: Move out of controller and into lti_helper (create if not yet exists). Try to keep controllers concise.

app/controllers/lti_controller.rb
```diff
@@ -47,4 +48,44 @@ def set_lti_message
...
+    end
+  end
+
+  def get_series_url(course_id, series)
+    if series.hidden?
+      series_url(series, token: series.access_token)
+    else
+      course_url(course_id, anchor: series.anchor)
+    end
+  end
```
thepieterdc, line 82: Move to series helper

app/controllers/lti_controller.rb
```diff
@@ -47,4 +48,44 @@ def set_lti_message
...
+
+  def get_course_url(course)
+    if course.secret_required?
+      course_url(course, secret: course.secret)
+    else
+      course_url(course)
+    end
+  end
```
thepieterdc, line 90: Move to course helper

app/views/lti/series_and_activities.js.erb
```diff
@@ -0,0 +1 @@
+$("#series").html("<%= escape_javascript(render partial: 'series_and_activities', locals: {series: @series, multiple: @multiple}) %>");
```
thepieterdc, line 1: Can de-jQuery-fi

```suggestion
document.getElementById("series").innerHTML = "<%= escape_javascript(render partial: 'series_and_activities', locals: {series: @series, multiple: @multiple}) %>";
```

thepieterdc (changes requested):
Looking good! Some minor remarks, mostly refactorings.

app/assets/javascripts/lti.ts
niknetniko: I agree with the previous comment, but these are pretty specific to this page (e.g. removing/adding the ID attribute of the first `select` element to make elements active and inactive).

app/assets/javascripts/lti.ts
niknetniko: Same origin is the default value, so I'll just remove it

PR #2257, opened 2020-09-15
Fix long activity navigations

### LINKED ISSUE #2047: Too many exercises in 1 block makes the rest not visible
When I open an exercises from a big block (see below) the bottom ones can get blocked off in the exercise sideview. I can't scroll down anymore therefor I can't click on the bottom exercises in the picture, here it stops at "Chromosomale crossover" while there are more exercises visible on the first picture.
image
image

niknetniko:
All styles for the Bootstrap affix stuff were in place, but it was not
activated.

See https://getbootstrap.com/docs/3.4/javascript/#affix

Fixes #2047

### DIFF
--- a/app/assets/stylesheets/models/course.css.less
+++ b/app/assets/stylesheets/models/course.css.less
@@ -26,6 +26,12 @@
   }
 }
 
+.activity-sidebar {
+  .nav {
+    padding-bottom: 80px;
+  }
+}
+
 .activity-sidebar,
 .series-sidebar {
   padding: 0;
--- a/app/views/activities/show.html.erb
+++ b/app/views/activities/show.html.erb
@@ -47,7 +47,7 @@ end %>
 
   <% if @series.present? %>
     <nav class="col-md-1 hidden-xs hidden-sm hidden-print activity-sidebar">
-      <ul class="nav affix" data-offset-bottom="80">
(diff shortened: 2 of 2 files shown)


bmesuere:
I'm confused. The affix stuff is to update the active state depending on the in-page scroll position of the page (as we do on the course page). On an activity page, however, this should never be updated because only a single exercise is shown on that page.

niknetniko:
When I apply it on the navigation, it activates when I scroll to the bottom and the footer comes into view. Once activated it then applies the offset for the footer (otherwise the footer overlaps with the series navigation if it is long). I thought it was intended to apply the affix stuff, otherwise the `data-offset-bottom="80"` on the series navigation does nothing as far as I can see.

charvp:
This does indeed fix the issue, but I think a simple `margin-bottom=80px` on the element should suffice as well? As @bmesuere says, we don't actually need the scrollspy functionality in this case. The `data-offset-bottom="80"` is probably copy-paste from the course page, where we do use the scrollspy functionality.

niknetniko:
Margins are ignored since the element is fixed, but some padding also fixes the issue.

charvp (approved):

bmesuere (approved):

PR #2260, opened 2020-09-16
Add in-page setting to enable/disable questions

niknetniko:
I've made it similar to the hidden series:

image

### DIFF
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -131,6 +131,7 @@ class CoursesController < ApplicationController
       if @course.update(permitted_attributes(@course))
         format.html { redirect_to @course, notice: I18n.t('controllers.updated', model: Course.model_name.human) }
         format.json { render :show, status: :ok, location: @course }
+        format.js { render :reload_questions }
       else
         format.html { render :edit }
         format.json { render json: @course.errors, status: :unprocessable_entity }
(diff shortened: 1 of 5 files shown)


app/controllers/courses_controller.rb
```diff
@@ -129,7 +129,10 @@ def create
   def update
     respond_to do |format|
       if @course.update(permitted_attributes(@course))
-        format.html { redirect_to @course, notice: I18n.t('controllers.updated', model: Course.model_name.human) }
+        format.html do
+          target = params[:questions] ? questions_course_path(@course) : @course
+          redirect_to target, notice: I18n.t('controllers.updated', model: Course.model_name.human)
+        end
```
niknetniko, line 135: Not sure if this is the best way to do this in rails

niknetniko: To recap: this link sends a request to the update method in the controller, which always redirects to the course overview path.

- The code now uses an additional parameter to redirect to the questions page if the parameter is true.
- Using `remote: true` on the question page might work, but requires `format.js { }` in the course controller and probably some js to actually reload the questions page.
- We could use `redirect_back()`  in the course controller, which will redirect the user to the page they came from. This works fine for the questions page, but has a side effect: the course edit page will no longer redirect to the course overview page, but will redirect back to the course edit page.
- We could not use a built-in link and instead send the request using javascript (`fetch` and stuff) and manually refresh the page once the request completes (since we will get a result in json back then).

charvp: I think we usually opt for `format.js` in this case?

niknetniko: I have changed it to format.js

charvp (approved):

bmesuere (approved):

Issue #2266, opened 2020-09-21
Dodona should honour locale if included in LTI request


niknetniko:
Most LTI requests (ie. the launch request & deep linking request) include optional "launch params", which in turn include an optional `locale`. If present, Dodona should honour this locale. (Brightspace & Moodle do include this)

PR #2285, opened 2020-09-22
Respect LTI language if present

niknetniko:
This pull request implements #2266.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -122,7 +122,7 @@ class ApplicationController < ActionController::Base
   end
 
   def set_locale
-    helpers.set_locale(params[:locale] || current_user&.lang || I18n.default_locale, current_user)
+    helpers.locale = params[:locale] || current_user&.lang || I18n.default_locale
   end
 
   def default_url_options
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -122,12 +122,7 @@ class ApplicationController < ActionController::Base
   end
 
   def set_locale
-    begin
-      I18n.locale = params[:locale] || current_user&.lang || I18n.default_locale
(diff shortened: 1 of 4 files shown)


app/helpers/application_helper.rb
```diff
@@ -227,4 +227,15 @@ def page_navigation_links(pages, remote = false, controller = '', params = {}, a
     end
   end
   # rubocop:enable Metrics/ParameterLists
+
+  def set_locale(locale, user)
```
thepieterdc, line 231: Is the `user` argument required here if this is only called for `current_user`?

app/controllers/application_controller.rb
```diff
@@ -122,12 +122,7 @@ def user_not_authorized
   end
 
   def set_locale
-    begin
-      I18n.locale = params[:locale] || current_user&.lang || I18n.default_locale
-    rescue I18n::InvalidLocale
-      I18n.locale = I18n.default_locale
-    end
-    current_user&.update(lang: I18n.locale.to_s) if current_user&.lang != I18n.locale.to_s
+    helpers.set_locale(params[:locale] || current_user&.lang || I18n.default_locale, current_user)
```
charvp, line 125: I don't think `helpers.` is required here?

app/helpers/application_helper.rb
niknetniko: This was done to be able to call the helper from the tests without setting up the whole authentication stuff. There is probably a way to do this, but I haven't found it yet.

app/controllers/application_controller.rb
niknetniko: This doesn't seem to work without, since the controller function is also called `set_locale`

charvp: Ah, good point.

bmesuere (approved):

PR #2289, opened 2020-09-23
Small refactor in questions

niknetniko:
This pull request address [this comment](https://github.com/dodona-edu/dodona/pull/2094#discussion_r488618617) on the original PR, by incorporating the `question_state` changes into the `update` action.

### DIFF
--- a/app/assets/javascripts/code_listing/annotation.ts
+++ b/app/assets/javascripts/code_listing/annotation.ts
@@ -1,4 +1,5 @@
 export type AnnotationType = "error" | "info" | "user" | "warning" | "question";
+export type QuestionState = "unanswered" | "answered" | "in_progress";
 
 export abstract class Annotation {
     private static idCounter = 0;
@@ -30,18 +31,6 @@ export abstract class Annotation {
     }
 
     protected edit(): void {
-        // Do nothing.
-    }
-
-    protected resolve(): void {
-        // Do nothing.
-    }
-
-    protected inProgress(): void {
-        // Do nothing.
-    }
-
-    protected unresolve(): void {
         // Do nothing.
     }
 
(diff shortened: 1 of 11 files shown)


charvp (commented):
I feel like things could be simpler if you use the record in the `permitted_attributes_for_update` method. Did you try this?

niknetniko:
> I feel like things could be simpler if you use the record in the `permitted_attributes_for_update` method. Did you try this?

Unless I'm missing something, do you mean something like this?

```ruby
def update?
  return false if record.answered?

  super || user.course_admin?(record.submission.course)
end

def permitted_attributes_for_update
  if transitioning? and transition?(record.transition_to, record.transition_from)
    %i[question_state]
  else
    super
  end
end
```

Some more thoughts:

- The `transition?` function is useful as a standalone function to be able to check if the question can transition to a specific state (which is used in the question table and in the json builder).
- The `transition_from` is useful to detect if the question is still in the expected state
- The `transition_to` is needed because you can't access the params in the policy

charvp:
> > I feel like things could be simpler if you use the record in the `permitted_attributes_for_update` method. Did you try this?
> 
> Unless I'm missing something, do you mean something like this?
> 
> <snip>

That is what I meant indeed. Looking at it with a fresh perspective though, I guess you lose the ability to tell the user that their update failed with that construction? That is something that would be nice to have I think.

niknetniko:
> I guess you lose the ability to tell the user that their update failed with that construction

Just tried it, and that's exactly what happens. This behaviour is currently used to notify the user if the question was already changed by someone else, so it seems useful to keep it.

charvp (approved):

bmesuere (approved):

PR #2292, opened 2020-09-24
Fixes for LTI series URL

niknetniko:
- Ufora just appends query params to whatever URL we give it, so we can't give anchors. Just link to the series URL instead.
- The second commit is to prevent issues when the form sends `activities: [""]`, which happens if you select an activity and then select nothing again.


charvp (approved):

charvp:
(Tests still need to be updated.)

bmesuere (approved):
ok, except for tests

niknetniko:
Tests are now fixed

PR #2302, opened 2020-09-28
Show number of unanswered questions in title of question page

niknetniko:
This pull request adds the number of unanswered questions in the title & shows a dot in the favicon if there are any unanswered questions (exactly like the notification dot).

The dot in the favicon now serves two purposes, but since course administrators currently don't receive a lot of notifications (I assume), I don't think it's a big problem.

More technically, the PR adds a `FaviconManager`, which tracks why a dot is shown in the favicon. For example, if you have unread notifications but you resolve all questions, we can't hide the dot yet (and vice versa).

Screenshot:

image

Closes #2275.

PR #2308, opened 2020-09-29
Check notifications every minute

charvp:
Depends on #2302.


niknetniko (changes requested):
After a refresh has happened, toggling the read state of a notification is broken for me. Clicking the dot of a notification doesn't seem to do anything.

bmesuere (approved):

bmesuere:
I didn't run it locally, so my approval is not because it works for me

charvp:
Apparently contained JavaScript isn't executed when using `innerHtml = `. jQuery does some special stuff for that in its `html()` method. There's [some ways](https://stackoverflow.com/questions/2592092/executing-script-elements-inserted-with-innerhtml) of doing it, but they're all pretty involved, so I just opted for using jQuery for now, while we still have it. If/when we ever get rid of it, this will have to be fixed in a lot of places anyway.

niknetniko (approved):

PR #2310, opened 2020-10-01
Periodically reload to watch for new submissions II

### LINKED ISSUE #1874: Option to reload submissions page periodically

niknetniko:
This pull request supersedes #1950. It is mainly the same PR, but updated. There are some differences:
- The toggle to enable refresh is now an option instead of an action. The options have been renamed from "Filter options" to "Options". Screenshot when there are other options:
image

- The refresh uses the interaction detection mechanism, so it won't refresh while scrolling, typing or otherwise interacting with the results.

- [x] Tests were added

Closes #1874.


app/assets/javascripts/index.js
```diff
@@ -410,4 +436,16 @@ function initFilterButtons() {
     init();
 }
 
-export { initFilterButtons, initFilterIndex, setBaseUrl };
+
+function toggleIndexReload() {
+    const loader = window.dodona.index.periodicReload;
+    if (loader !== null) {
```
charvp, line 442: This shouldn't happen right? The `toggleIndexReload` function is (should) only be used on pages where `refreshElement` wasn't `null`, which means that `window.dodona.index.periodicReload` will have been set by `initRefresh()`.

I'm not necessarily saying it should be changed, justed wanted to check.

charvp: Although, if someone forgets to pass `refreshElement` in the future, this will probably make it harder to debug why things aren't working. Maybe we should just let it error out in that case?

config/locales/views/submissions/nl.yml
```diff
@@ -17,6 +17,7 @@ nl:
       title: Alle oplossingen
       for: voor
       from: van
+      watch_submissions: Nieuwe oplossingen inladen
```
bmesuere, line 20:
```suggestion
      watch_submissions: Oplossingen automatisch herladen
```

config/locales/views/submissions/en.yml
```diff
@@ -18,6 +18,7 @@ en:
       for: for
       from: from
       most_recent_correct: Most recent correct submissions per user
+      watch_submissions: Watch for new submissions
```
bmesuere, line 21:
```suggestion
      watch_submissions: Automatically reload submissions
```

bmesuere (changes requested):

app/assets/javascripts/index.js
niknetniko: Correct on both accounts, I've changed it (and moved the toggle code to the class instead).

bmesuere (approved):

charvp:
This has a merge conflict, but otherwise can be merged?

PR #2314, opened 2020-10-03
Add course description to json

stijndcl:
This pull request adds a course's description to it's json file.

### DIFF
--- a/app/views/courses/_course.json.jbuilder
+++ b/app/views/courses/_course.json.jbuilder
@@ -1,3 +1,3 @@
-json.extract! course, :id, :name, :teacher, :color, :year, :visibility, :registration, :created_at, :updated_at
+json.extract! course, :id, :name, :description, :teacher, :color, :year, :visibility, :registration, :created_at, :updated_at
 json.url course_url(course, format: :json)
 json.series course_series_index_url(course, format: :json)


bmesuere:
I really appreciate the work that has been done on the plugins that make it easier to write your code in a real IDE instead of the Dodona editor. However, I don't think the goal should be to create an alternative "Dodona client". We do want people the use the Dodona website for several reasons.
Such alternative client will never have feature parity with the main application, and we don't want to commit to this. New features such as asking questions and receiving feedback on your code are completely invisible if you only rely on IDE plugins. In addition, we have no way to communicate with users using banners or flash messages. Finally, since Dodona is used for examinations and tests, we need to have a certain level of confidence that stuff works in all environments. Including extra clients over which we have limited control is not something we want.

To summarize:
- exercise selection and one click submissions: 👍 
- easy links to exercise descriptions and test results: 👍 
- rendering exercise descriptions: gray zone. It would seem that the added value is limited and could potentially break future new features such as new exercise types
- rendering the course description and results table: 👎

stijndcl:
@bmesuere I definitely understand that. I was just looking into implementing some of the feature requests we've been getting, it's definitely not a requirement to have this in the plugin (and it wasn't even supposed to be possible in the first place). The thought of this slowly becoming a replacement for Dodona if I keep going with these requests has also come into my mind, and I must admit I wasn't 100% behind that idea either.

I've closed the PR, I don't think review from anyone else is needed anymore. Thank you for the reply, though.

BTWS2:
@bmesuere thank you for your remarks. I proposed an alternative solution that sends users to dodona.ugent.be and has the same effect as the original feature request for the user. What's your opinion on the [new proposal](https://github.com/thepieterdc/dodona-plugin-vscode/issues/59#issuecomment-703851925)?

> Maybe an alternative solution is linking to the series description (and course?) on dodona.ugent.be with the edit.svg icon, instead of displaying it in VS Code itself? This is in alignment with the vision of Bart and sends users to the Dodona website. The icon would appear when the mouse hovers over the name.
> 
> This would example below would link to https://dodona.ugent.be/nl/courses/293/ and https://dodona.ugent.be/nl/courses/293/#series-2561-0-proloog.

image

stijndcl:
* quoting my response to this as well, as it provides a possible solution

> I can see how that is a better solution than integrating it into the plugin itself, but that would still mean students would have to click on that button to check the link & see if there is any important info on that page. As a student myself I think I can confidently say that you can't guarantee people will do that, even if the button is there.
>
> I think some sort of in-your-face popup might be required that checks whether or not the series actually *has* a description attached to it, and notifying the user when it does. That might be a viable solution (combining it with your proposal, so students can always click on it themselves). When a student opens a new exercise in a series with a description, and they have never made an exercise in that series before (to avoid spamming it), they should be notified that the description might contain important information?

PR #2328, opened 2020-10-09
Update to ruby 2.7.2

charvp:
Ruby 2.7.2 turns off deprecation warnings by default, so I also removed our custom code to disable the warnings in production. The deprecation warnings can only be turned on again using a command-line option.


bmesuere (approved):

bmesuere:
Maybe wait till Github actions supports this version?

charvp:
Apparently `ruby/setup-ruby` is prefered over `actions/setup-ruby`. It is faster to update in any case. I've switched to using that.

niknetniko (approved):

bmesuere (approved):

PR #2342, opened 2020-10-14
Include course_labels in export

### LINKED ISSUE #2341: User labels are not exported in the submission-export-overview
It would be useful to also find a column `labels` in the overview (`info.csv`) of an export of all student-submission.

rien:
This pull request adds course labels to series or course exports for course admins.

I confirmed manually that labels are included as expected. Adding a test to check if the correct course labels are present would complicate and slow down the already complex code in `test/testhelpers/export_zip_helper.rb`.

Closes #2341.

### DIFF
--- a/app/controllers/exports_controller.rb
+++ b/app/controllers/exports_controller.rb
@@ -78,9 +78,10 @@ class ExportsController < ApplicationController
 
     if @user.blank? && (item.class == Series || item.class == Course)
       authorize item, :course_admin?
-      params[:include_labels] = true
+      params[:with_labels] = true
     else
-      params.remove(:include_labels)
+      # only course admins should be able to export labels
+      params.delete(:with_labels)
     end
 
     Export.create(user: current_user).delay(queue: 'exports').start(item, list, ([@user] if @user), params)
--- a/app/controllers/exports_controller.rb
+++ b/app/controllers/exports_controller.rb
@@ -75,7 +75,13 @@ class ExportsController < ApplicationController
   def create(item, list)
     authorize item, :export?
     authorize @user, :export? if @user
-    authorize item, :course_admin? if @user.blank? && (item.class == Series || item.class == Course)
+
+    if @user.blank? && (item.class == Series || item.class == Course)
+      authorize item, :course_admin?
+      params[:include_labels] = true
+    else
+      params.remove(:include_labels)
+    end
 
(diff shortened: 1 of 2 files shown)


app/helpers/export_helper.rb
```diff
@@ -208,7 +223,9 @@ def csv_submission(csv, user, exercise, submission, filename)
       csv << if indianio?
                [filename, submission&.status, submission&.id, exercise.name_en, exercise.name_nl, exercise.id]
              else
-               [filename, user.full_name, user.id, submission&.status, submission&.id, exercise.name_en, exercise.name_nl, exercise.id, submission&.created_at]
+               row = [filename, user.full_name, user.id, submission&.status, submission&.id, exercise.name_en, exercise.name_nl, exercise.id, submission&.created_at]
+               labels? && row << @users_labels[user].map(&:name).join(';')
```
charvp, line 227: Why not this?
```suggestion
               row << @users_labels[user].map(&:name).join(';') if labels?
```

app/controllers/exports_controller.rb
```diff
@@ -75,7 +75,14 @@ def set_user
   def create(item, list)
     authorize item, :export?
     authorize @user, :export? if @user
-    authorize item, :course_admin? if @user.blank? && (item.class == Series || item.class == Course)
+
+    if @user.blank? && (item.class == Series || item.class == Course)
```
charvp, line 79: Why not make this an actual option that the user can choose to have?

rien: Adding an actual option will add a bit more complexity to the user interface (a checkbox that is conditionally disabled) and will make the permission check a bit more complex. 

I can add the option if you want. But I don't think it would be a nuisance to have the labels included in the export,  apart from breaking very badly written scripts

charvp: Sure.

bmesuere (approved):

niknetniko (approved):

rien:
Is a documentation update needed here? This seems like a tiny detail that can be added to the release notes, but does not need special mention on the docs.

charvp:
> Is a documentation update needed here? This seems like a tiny detail that can be added to the release notes, but does not need special mention on the docs.

I don't think this needs special mention in the docs.

PR #2350, opened 2020-10-19
Invalidate series status cache on deadline change

### LINKED ISSUE #2218: Series status not updated after deadline removal
This is the view I got after removing the deadline of a series. The status of the series has not been updated. Before the deadline there was one exercise where I had an incorrect solution before the deadline, and a correct solution after the deadline. After the deadline was remove, the status of the exercise itself was updated, but not the status of the series (or updated incorrectly (*)).

image

(*) If I recall well, the status before removing the deadline was actually showing a clock as an indication that the deadline was missed.

niknetniko:
This pull request modifies the series status cache to:

- No longer include the deadline in the cache key
- Invalidate all caches on deadline change

The problem this solves is e.g. this scenario:

1. The user submits something in a series with a deadline
2. Someone changes the deadline. Old series status is not removed, but another cache key is used => no problem
3. The user does a submission. Series status cache is removed, but for current deadline => no problem
4. Someone changes deadline back. Now the old cache is used again, since the key is the same. The new submission from step 3 is thus ignored.

Suggestions welcome if there is a better way of doing this.

- [x] Tests were added. I've introduced a helper that enables an in-memory cache for the specified tests.

(Should hopefully) close #2218.

### DIFF
--- a/app/models/series.rb
+++ b/app/models/series.rb
@@ -25,7 +25,7 @@ class Series < ApplicationRecord
   include Cacheable
   include Tokenable
 
-  USER_COMPLETED_CACHE_STRING = '/series/%<id>s/deadline/%<deadline>s/user/%<user_id>s/completed'.freeze
+  USER_COMPLETED_CACHE_STRING = '/series/%<id>s/user/%<user_id>s/completed'.freeze
   USER_STARTED_CACHE_STRING = '/series/%<id>s/user/%<user_id>s/started'.freeze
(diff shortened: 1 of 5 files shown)


rien (approved):
This seems like a fix for the problem you describe.

I was afraid the invalidation would take some time, but it looks like just one database query will be performed and then just `3*#users` cache invalidation requests (which will probably be fast enough?).

charvp (commented):
Can we do the invalidation asynchronously? The largest courses have upward of 400 users, which would mean 1200 cache calls, which doesn't feel so good to me.

charvp (approved):

PR #2351, opened 2020-10-19
Clear ActivityStatus cache when adding/removing exercises to series

### LINKED ISSUE #2079: ActivityStatus is not invalidated when adding/removing exercises in series
Currently, the status is only updated when changing the deadline

niknetniko:
This pull request invalidates the status of the series when adding/removing exercises to the series.

I don't think it is necessary to clear the actual activity statuses themselves (as happens when the deadline of the series changes), since they are still valid.

- [x] Tests were added

Closes #2079.

### DIFF
--- a/app/models/series_membership.rb
+++ b/app/models/series_membership.rb
@@ -21,11 +21,16 @@ class SeriesMembership < ApplicationRecord
   validates :series_id, uniqueness: { scope: :activity_id }
   after_create :invalidate_caches
   after_destroy :invalidate_caches
+  after_destroy :invalidate_status
   after_destroy :regenerate_activity_token
 
   def invalidate_caches
     course.invalidate_activities_count_cache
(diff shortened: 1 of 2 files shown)


charvp:
Regarding your last point, wouldn't they be invalid in the following scenario?
 * Exercise solved by user at `t`.
 * Exercise removed from series.
 * Deadline changed to `t - 1`.
 * Exercise re-added to series.
Note that I'm not actually sure about this, but maybe we should add a test anyway to make sure we don't break that scenario in the future, even if it works now?

niknetniko:
I've:
- added a test for the scenario; it was already covered, because activity statuses are deleted when the deadline is changed
- added code the remove relevant `ActivityStatus`es when removed from a series anyway, since it is more "correct" and guards against cases where other code can manually retrieve all statuses directly.

charvp (approved):

Issue #2265, opened 2020-09-19
Additional student question features


pdawyndt:
### general

- [ ] direct navigation from submission to question dashboard
- [ ] direct navigation from submission linked to question to latest submission (after posing question, student may have newer submissions for the same exercise) or other submissions for the same exericise in general
- [ ] allow students to ask a question on an exercise, a series or a course (not a specific submission); would also allow to ask question on reading assignments (#2329)
- [ ] link question and answer in discussion thread; upon answering question, course admin is asked if state needs to be changed (start of answering > mark as in progress if no already the case; end of answering > mark as answered if not already the case)
- [ ] course admins should see the “ask a question” features on their own submissions, instead of the “code review” features (this way they can also demo the feature)
- [ ] if student asks multiple questions on the same submission, we still might have multiple course managers working on the same submission (student might e.g. ask a second question after course manager started answering the first question)
- [ ] students indicate that they have asked a question, but no answer has arrived yet; this is due to the fact that they have no clue where they are in the queue; some indication about "number of questions waiting in front of your" or an ETA would be nice to have
- [ ] give students a "help me I'm stuck" button (both at exercise and submission level; the proverbial "ask the duck" button); some students already seem to freeze when comfronted with an empty edit box on how to ask a proper question; this might bypass this freeze-moment and make it easier to ask for help (anonymously from other students)
- [ ] support [text shortcuts to include emoji](https://emojipedia.org/emoji/%F0%9F%A6%86/) as in MS Teams (e.g. `:duck:`) (#2311)
- [ ] immediately inform course admin when student has marked questions as "resolved"; if a course admin is busy answering a question, there's currently no way to get signaled
- [ ] start video chat immediately from student question (moonshot)
- [ ] activate "ask questions" for a selection of exercises (e.g. series where deadline is not expired, selected series (e.g. only exam series); suggested by Dries Benoit
- [ ] allow sending emails to course admins when a new question was asked; maybe we should bundle multiple questions in a single digest; we should allow course admins to switch on/of the option to receive emails upon questions (suggested by Dominiek Vandewalle; more info: #1788)

### question dashboard

- [x] replace "Er zijn geen vragen beschikbaar" by "Er zijn op dit moment geen onbeantwoorde vragen."
- [x] show name of course admin answering question
- [ ] combine actions of claiming a question and navigation to the submission page as a single step
- [x] add visual clue to browser tab if there are unanswered questions (c.f. new notifications on twitter) (#2275)
[...]

PR #2354, opened 2020-10-21
Question overview page

niknetniko:
This pull request adds an overview page for questions, similar to the one for submissions:

image

Some remarks:

- Search is possible on exercise & student.
- Filtering is possible on status & course.
- Filter options are:
  - Only show questions from courses where you are an admin (especially useful for Zeus)
  - Auto-refresh
- Searching on question content is not implemented, since that would become very slow; if we want this, we'll probably need a full text index.

The page is accessible from:
- The sidebar for zeus & staff
image

- The user profile dropdown
image

With the filters and options, this page implements some suggestions from #2265 related to the question dashboard, e.g. the overview of questions across courses, easier browsing of all questions or students being able to view all their own questions.

- [x] Tests were added


app/views/annotations/question_index.html.erb
```diff
@@ -0,0 +1,39 @@
...
+            <% end %>
+          </span>
+        </h2>
+      </div>
+      <div class="card-supporting-text" id="refresh-element">
+        <% actions = [] %>
+        <% if current_user&.admin? %>
+          <%
+            actions << { icon: 'file-eye', text: t('questions.index.watch'), search: { refresh: true }, click: 'window.dodona.toggleIndexReload()' }
+            actions << { icon: 'done', text: t('questions.index.strict'), search: {administrating: true} }
```
charvp, line 29: `a_course_admin?` might be a more accurate method for determining who might want this? And can we enable this by default for Zeus?

charvp (commented):
What's the reason for filtering by course in two different ways? Can't we just link to the questions page with the label prefilled in the search bar?

niknetniko:
> What's the reason for filtering by course in two different ways? Can't we just link to the questions page with the label prefilled in the search bar?

This was from a previous iteration, and was no longer relevant. I've removed the duplicate filtering.

charvp (approved):

PR #2363, opened 2020-10-27
Clear activity status store when performing a new job

charvp:
This should prevent the workers from updating deleted ActivityStatus objects.


bmesuere (approved):
Great detective work! Hopefully, this solves the issue we're seeing with incorrect statuses. Do we need to clear all non-read states to be sure?

niknetniko (approved):

PR #2365, opened 2020-10-29
Create read state within course if content page is opened in course without series

charvp:
This pull request fixes read states being created outside of the course when the content page is opened without a series id in the URL. (This is used in evaluations right now to allow access to and 'reading of' content pages without giving access to the full series yet.)


niknetniko (approved):

bmesuere (approved):
lgtm

PR #2392, opened 2020-11-16
Fix nil when converting exercises to reading activities

### LINKED ISSUE #2364: Crash for question on submission without exercise
If an exercise is converted to a reading activity, the submissions lose their exercise. Question-related code currently assumes all submissions have an exercise.

niknetniko:
Fixes issues with nil in the submission and question related pages when an exercise with existing submissions is converted to a reading activity. This PR renames the existing column `submissions.excerise_id` to `submissions.activity_id`, and changes to the code to the the linked entity as an activity wherever possible, and making the code that requires an exercise nil-proof:

- `exercise` and `exercise_id` have been renamed everywhere it was possible to use the linked entity as an activity instead of an exercise.
- `exercise` is kept as a nillable attribute:
   - When submissions lose their programming language, they become `text`.
   - When submissions lose their judge/renderer, the default one is used.
   - When attempting to evaluate a submission without exercise, an error is thrown. This allows us to catch those cases, since they shouldn't happen.
    - Submissions without exercise are excluded from mass rejudges and exercise-dependent cache invalidation stuff.
- `exercise_id` continues to be accepted and returned via the API, next to `activity_id`. This to prevent breaking clients.

- [x] Tests were added.
- [ ] Documentation was updated.

Closes #2364.

### DIFF
--- a/app/assets/javascripts/exercise.js
+++ b/app/assets/javascripts/exercise.js
@@ -221,7 +221,7 @@ function initExerciseShow(exerciseId, programmingLanguage, loggedIn, editorShown
         return $.post("/submissions.json", {
             submission: {
                 code: code,
-                exercise_id: exerciseId,
+                activity_id: exerciseId,
                 course_id: courseId,
             },
         });
--- a/app/controllers/annotations_controller.rb
+++ b/app/controllers/annotations_controller.rb
@@ -7,7 +7,7 @@ class AnnotationsController < ApplicationController
   has_scope :by_course, as: :course_id
 
   has_scope :by_filter, as: 'filter' do |controller, scope, value|
-    scope.by_filter(value, skip_user: controller.params[:user_id].present?, skip_exercise: controller.params[:exercise_id].present?)
+    scope.by_filter(value, skip_user: controller.params[:user_id].present?, skip_activity: controller.params[:exercise_id].present?)
   end
 
   def index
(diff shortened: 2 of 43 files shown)


charvp (approved):
We will need to schedule downtime when deploying this. It makes changes to the submissions table that will be problematic if running code that doesn't know about these changes interacts with the database. The migration will also probably take a while, since it is on the submissions table.

bmesuere (commented):
Sorry, this one escaped my attention. I didn't review it right away because of its size.

I'm not super fond of these changes, only to accomodate an edge case. Are there other benefits?

An alternative would be to not allow such conversion, e.g. by generating a new id. The old exercise in Dodona would then be marked as removed and the old submissions would stay available that way.

niknetniko:
> I'm not super fond of these changes, only to accomodate an edge case. Are there other benefits?

Sadly not.

> An alternative would be to not allow such conversion, e.g. by generating a new id. The old exercise in Dodona would then be marked as removed and the old submissions would stay available that way.

In #2364 @chvp remarked:

> The problem with option 4 is that if the conversion is reversed the submissions are not linked to the original exercise anymore. I don't really have a better solution right now, but it seems like an important downside.

However, since we haven't had this error since then, perhaps this is so rare that we can ignore this downside (or manually fix it if it ever occurs)?

bmesuere:
I think it's even trivial to fix manually if the downside of option 4 would occur by changing the generated id's.

@chvp what to you think? Revert on our previous decision and fix this another way?

chvp:
Sure, converting between activity types is already rare, so converting back will probably be exceedingly rare.

bmesuere:
Sorry for the unnecessary work, @niknetniko!

PR #2393, opened 2020-11-16
Use Rails logger for Omniauth

niknetniko:
This pull request integrates omniauth logging with the rails logger, getting rid of this:

image

Main rational is that I'm tired of having these logs when I run tests locally.

Since we also use the rails logger for the rest, I don't think it's necessary to make this a conditional change (e.g. only in test or dev environment), but not sure.


charvp (approved):
🎆

PR #2394, opened 2020-11-17
Link to student & exercise in evaluation

### LINKED ISSUE #2384: Link submission in evaluation to exercise
When a submission is viewed in an evaluation, there's no direct link to the corresponding exercise. This seems like the logical place to add a link in the UI.

image

niknetniko:
Fixes #2384:

image

The links open in a new tab (`target='_blank'`).

### DIFF
--- a/app/views/feedbacks/show.html.erb
+++ b/app/views/feedbacks/show.html.erb
@@ -16,7 +16,13 @@
   <div class="col-xs-12 col-lg-pull-3 col-lg-9 feedback-submission">
     <div class="card feedback-show">
       <div class="card-title card-title-colored">
-        <h2 class="card-title-text"><%= t '.submission_for_by', for: @feedback.exercise.name, by: @feedback.user.full_name %></h2>
+        <h2 class="card-title-text"><span>
+          <%= t '.submission_for' %>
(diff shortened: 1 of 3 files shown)


pdawyndt:
Funda Blokhuis: goeie naam voor een student!

charvp (approved):
Do we have headers as links anywhere else?

niknetniko:
> Do we have headers as links anywhere else?

I did a quick search, but couldn't find any. Perhaps an alternative would be to make the name/exercise clickable in the "page" title, like on the submission page:
image

charvp:
Hmm, but on the submission page it's not in the card title header either right? It's a message at the top of the card content. I don't really have any better ideas though.

niknetniko:
> Hmm, but on the submission page it's not in the card title header either right? It's a message at the top of the card content. I don't really have any better ideas though.

Ah oops I meant to say the "submissions index" page, e.g: (it's not very discoverable though, the underline is only when hovering over the name)

image

niknetniko:
Changed the locations of the links to be in the card title, as discussed.

charvp (approved):

PR #2462, opened 2021-01-15
Grading

### LINKED ISSUE #2401: Grading in Dodona
Thoughts/proposal for a first version:

#### Terminology

- Evaluation: existing term in Dodona.
- Feedback: existing term in Dodona.
- Submission: existing term in Dodona.
- ScoreItem: the things we want to give a student a score on. ~~Proposed is to make this a tree structure, with the root of the tree being the grade item for an exercise. It is then possible to construct a tree to reflect how the scores should be assigned: x points for function A, y points for function B, etc. This is up to the evaluator. (Every ScoreTemplate would be linked to an EvaluationExercise in Dodona)~~  This is a list of items for each exercise.
- Score: A single score: the link between a ScoreItem and a Submission (and thus student). (A Score links a ScoreTemplate to a Feedback in Dodona)

#### Flow

The intended flow of grading would be like this:

1. The evaluator creates an evaluation for a series, as is done now.
2. After selecting the relevant exercises and users, there is a third step. In this step, the evaluator must specify the ScoreTemplates per exercise. This can be very simple (e.g. x/10 for each exercise), but a more detailed tree is possible. Automating this is difficult, because an exercise has no notion of the result that will be produced by the judge. We could allow a grading suggestion in the config of the exercise though (e.g. if you author an exercise with two functions, you can suggest splitting the grading into two grade items).
3. When actually doing the evaluation, the evaluator can only mark a submission as evaluated if each leaf in the ScoreTemplate tree is filled in. We'll need to think were we put this: above or below the submission code, or in a new tab (next to the "code" tab).
4. When the evaluation is released, the students will be able to see their grades. (Again, perhaps the total for a submission can be above the code, while the ScoreTemplates are displayed in a new tab.)

#### Future

Things that are not for this issue, but can be done later if we determine the basic version is deemed OK

##### LTI integration

In LTI, there are some different concepts:

- A line item is usually a column in the gradebook, and holds the result of a specific activity.
- A result is a single cell in the gradebook; it is unique for the line item and user. (read-only, reading scores from the platform)
- A score is the latest score a student has, together with a status (write-only, sending scores to the platform)

Some problems need investigating:

- How should we handle the case of not all students being in tyhe same series, e.g. in the course Programming? Ideally, Dodona would "group" these series in a way, allowing us to publish line items like "Question 1" for a certain evaluation, even if the "Question 1" are different exercises.
- It is possible to link line items to resources. However, the resources are not necessarily the same: e.g. we link to a series in Ufora (so the series is the "resource"), but the line items are the "exercises" in the series.

##### Penalties

Another useful addition would be the concept of penalties: annotations with an assigned score that is deducted from the total score for an exercise.

The main difference with the scenario where the evaluator fills in the scores is that here, each student starts with the maximum available points, and the annotations are subtracted from that score. (e.g. 10/10 for a function, but -1 penalty because of "break/continue" => score of 9/10).

This will need some thought on how both modes can be integrated together, and how the penalties act when the evaluator wishes to override the final score. One possibility is allowing a custom start score (e.g. the student starts with 9/10, from which penalties are deducted). This might feel arbitrary to students though (e.g. why do I have 7/10, when I only have two penalties of -1 each?). Another option is to disallow mixing both modes: either the student starts with 0 and the evaluator fills in the score, or the students starts with the max score and the evaluator must use penalties to deduct points.

niknetniko:
This pull request adds grading to Dodona.

- [x] Tests were added

Closes #2401.
Follow up issues: #2516 #2525


app/policies/feedback_policy.rb
```diff
@@ -19,8 +21,23 @@ def update?
...
+  def refresh?
+    show?
+  end
+
+  def complete?
+    required_rubrics = record.rubrics.map(&:id).to_set
+    found_rubrics = record.scores.map(&:rubric_id).to_set
+
+    # For every template, there should be a score
+    required_rubrics.subset?(found_rubrics)
```
charvp, line 33: Maybe I'm missing something, but isn't this just an equality check? It would seem weird to me that a score can be linked to a non-existent rubric.

app/views/evaluations/_evaluation_grade_table.html.erb
```diff
@@ -0,0 +1,45 @@
+<div class="table-scroll-wrapper status-table-wrapper score-table">
+  <table class="table table-resource">
+    <thead>
+    <tr>
+      <th class="user-name"><%= t('.user') %></th>
+      <% feedbacks[:evaluation_exercises].each do |exercise| %>
+        <th class="status-header"><%= exercise.exercise.name %> (/<%= format_score exercise.maximum_score %>)</th>
+      <% end %>
+      <th class="status-header">Totaal</th>
```
charvp, line 9: Add sum of maxima here?

charvp: Also, i18n.

app/views/evaluations/_grade_status.html.erb
```diff
@@ -0,0 +1,27 @@
...
+
+<% if feedback.rubrics.present? %>
+  <span class='feedback <%= submission_status %> <%= "active" if local_assigns[:current] == feedback %>'>
+    <%= link_to evaluation_feedback_path(evaluation, feedback), class: "btn btn-icon" do %>
+      <% if feedback.completed? %>
+        <span title="<%= feedback.exercise.name %>: <%= t "evaluations.feedback_status.submission_" + submission_status %><%= t "evaluations.feedback_status.feedback_finished" %>">
+            <span><%= format_score feedback.score %></span>
+        </span>
+      <% else %>
+        <span class="provisional" title="<%= feedback.exercise.name %>: <%= t "evaluations.feedback_status..submission_" + submission_status %><%= t "evaluations.feedback_status.feedback_unstarted" %>">
```
charvp, line 19: Is the `..` on purpose? I've never seen that before.

app/views/feedbacks/_score_actions.html.erb
```diff
@@ -0,0 +1,40 @@
...
+           end
+    %>
+    <div class="form-group input <%= e %>">
+      <div class="control-wrapper">
+        <%= f.number_field :score, required: true,
+                           class: "form-control score-input",
+                           step: 0.01,
+                           tabindex: index + 1,
+                           value: format_score(f.object.score, 'en') %>
+        <div class="dodona-progress dodona-progress-indeterminate" style="visibility: hidden;">
```
charvp, line 20: We usually don't do inline styles.

app/views/rubrics/_copy.html.erb
```diff
@@ -0,0 +1,18 @@
+<%= form_for(:copy, url: copy_evaluation_rubrics_path(@evaluation), html: { class: 'form-horizontal' }, namespace: evaluation_exercise.id, remote: true) do |f| %>
+  <%= f.hidden_field :to, value: evaluation_exercise.id %>
+  <div class="field form-group">
+    <%= f.label t("rubrics.copy.exercise"), :class => "col-sm-3 control-label" %>
+    <div class="col-sm-6">
+      <%= f.collection_select :from,
+                              evaluation.evaluation_exercises.includes(:exercise).filter { |o| o.id != evaluation_exercise.id },
+                              :id,
+                              ->(e) { e.exercise.name },
+                              { include_blank: "-- Kies een oefening --" }, { class: "form-control" } %>
```
charvp, line 10: I18n

app/views/rubrics/_table.html.erb
```diff
@@ -0,0 +1,66 @@
...
+    <th><%= t ".maximum" %></th>
+    <th><%= t ".visible" %></th>
+    <th></th>
+  </tr>
+  </thead>
+  <tbody>
+  <% if amount >= 2 %>
+    <tr>
+      <td>Totaal</td>
+      <td class="description">Som van alle rubrics</td>
```
charvp, line 19: I18n

charvp (changes requested):
The migrations give a different `schema.rb` file for me than is included in this pull request. E.g. `t.index ["rubric_id", "feedback_id"], name: "index_scores_on_rubric_id_and_feedback_id", unique: true` is added to `rubrics`. I'm pretty sure this also hides test failures, since they don't all succeed for me locally.

UI stuff:
 * Top row showing total (when adding) should be styled a little differently to show that it is a second header row.
 * Copying a rubric is a little confusing, since it adds all rubrics of another exercise, but doesn't remove the current rubrics.
 * Pressing enter to save a score results in errors.
 * It should be clearer that `Score: ../..` on the submission page is a dropdown. I'm also not entirely sure on the placement, but I'll let @bmesuere handle that part.

app/assets/javascripts/evaluation.ts
```diff
@@ -1,81 +1,375 @@
...
+
+    public markBusy(): void {
+        this.disableInputs();
+        this.input.classList.add("in-progress");
+        this.spinner.style.visibility = "visible";
+    }
+
+    public disableInputs(): void {
+        this.input.disabled = true;
+        this.disabled = false;
```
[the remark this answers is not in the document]
niknetniko: `disabled` was not actually used anywhere, so I removed it.

app/assets/javascripts/score_template.ts
```diff
@@ -0,0 +1,11 @@
...
+    document.querySelectorAll(".add-new-button").forEach(e => {
+        e.addEventListener("click", event => {
+            const evaluationExerciseId = (event.target as HTMLButtonElement).dataset.exerciseId;
+            document.getElementById(`new-template-exercise-${evaluationExerciseId}`)
+                .classList.toggle("hidden");
+        });
+    });
+}
+
+export { initAddScoreTemplates };
```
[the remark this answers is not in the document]
niknetniko: Good catch

app/controllers/concerns/rescue_json_response.rb
```diff
@@ -0,0 +1,11 @@
+module RescueJsonResponse
+  extend ActiveSupport::Concern
+
+  included do
+    rescue_from ActiveRecord::RecordInvalid do |exception|
```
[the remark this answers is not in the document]
niknetniko: I don't there is anything wrong with always doing this, but I don't think it will have a lot of impact. I added this for scenario's like this:

```ruby
def destroy
    @rubric.destroy!
    respond_to do |format|
      format.js { render 'rubrics/refresh', locals: { new: nil, evaluation_exercise: @rubric.evaluation_exercise } }
      format.json { render json: {}, status: :no_content }
    end
  end
```

However, in most places it looks like this (especially if there are also HMTL responses):

```ruby
  def update
    respond_to do |format|
      args = permitted_attributes(@annotation)
      args[:last_updated_by] = current_user
      if @annotation.update(args)
        format.json { render :show, status: :ok, location: @annotation.becomes(Annotation) }
      else
        format.json { render json: @annotation.errors, status: :unprocessable_entity }
      end
    end
  end
```

app/controllers/evaluations_controller.rb
```diff
@@ -59,6 +60,17 @@ def add_users
       [I18n.t('evaluations.add_users.title'), '#']
     ]
     @title = I18n.t('evaluations.add_users.title')
+    @graded = ActiveModel::Type::Boolean.new.cast(params['graded'])
+  end
+
+  def add_rubrics
+    @new_rubrics = true
+    add_or_edit_rubrics I18n.t('evaluations.rubrics.title_add')
+  end
```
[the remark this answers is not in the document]
niknetniko: I moved it to `rubrics#index` and `rubrics#new`.

The split into two actions is similar to the one for the user management page of an evaluation: on the first "add" page, there is some text that is displayed. (see [new.html.erb](https://github.com/dodona-edu/dodona/blob/8a1412c368150213cbf3d9fbef1207cbfac172c9/app/views/rubrics/new.html.erb) and [index.html.erb](https://github.com/dodona-edu/dodona/blob/8a1412c368150213cbf3d9fbef1207cbfac172c9/app/views/rubrics/index.html.erb)). They can be merged however, if that is preferable, since the differences are smaller than with the user page.

app/controllers/evaluations_controller.rb
```diff
@@ -136,6 +148,14 @@ def overview
     @title = I18n.t('evaluations.overview.title')
   end
 
+  def export_scores
```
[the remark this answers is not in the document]
niknetniko: So we don't forget: https://github.com/dodona-edu/dodona/issues/2503

app/controllers/feedbacks_controller.rb
```diff
@@ -34,11 +37,22 @@ def edit
...
+    @score_map = @feedback.scores.index_by(&:rubric_id)
     respond_to do |format|
       format.html { redirect_to evaluation_feedback_path(@feedback.evaluation, @feedback) }
       format.json { render :show, status: :ok, location: @feedback }
-      format.js
+      format.js {}
+    end
+  end
+
+  def refresh
```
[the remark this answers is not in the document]
niknetniko: Moved to `model#show.js`, with an explenation what the `warning` does.

app/models/score.rb
```diff
@@ -0,0 +1,27 @@
...
+class Score < ApplicationRecord
+  attribute :score, :decimal
+  # Used to detect conflicts
+  attribute :expected_score, :decimal
+
+  belongs_to :rubric
+  belongs_to :feedback
+  belongs_to :last_updated_by, class_name: 'User'
+
+  def out_of_bounds?
```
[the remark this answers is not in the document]
niknetniko: I thought we decided in a meeting to allow all scores, but display them in an "error state" (e.g. in red) if out of bounds. (I wrote that down, but I could be mistaken of course).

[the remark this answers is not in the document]
chvp: I don't remember that, but I wasn't (fully) at all meetings, so maybe I just wasn't there for it.

app/controllers/concerns/rescue_json_response.rb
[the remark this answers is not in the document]
niknetniko: As discussed, I'll rewrite this to remove this rescue stuff and use the if-else like in the rest of the code base.

bmesuere:
I'll evaluate the screen separately since I won't have time to do this all at once. Some remarks might be about evaluation in general and not specific to grading but are nonetheless relevant.

image

This is the first screen a user gets when starting an evaluation and this might be optimized a bit. A user has to set a deadline here and indicate if he wants to enable grades. The rest is just extra information (eg, the series overview) but this isn't super clear.

- include something about grades in the info box
- make it clearer that the user has to pick a deadline and can enable grades. Some ideas which may or may not work:
  - slightly change the wording, eg Deadline -> Pick a deadline
  - add a "Settings" header to separate the top part of the page from the bottom part
- Add some text to the top of the series overview that makes it clear that the listing of exercises is just informational 
- Maybe also tweak the card title to make it more clear that we started some sort of wizard, eg "Configuring a series evaluation"

bmesuere:
For user selection, the material design spec favors using checkboxes instead of buttons for selection. Selected rows should also have a slight background color:
image
vs
image

bmesuere:
image

- The form fields seem to have a missing translation
- For naming, I think the entirety of scoring items is called a rubric and not the individual items
- It isn't clear that I have to add score items to each exercise, even if I don't want to use a detailed rubric and only a single score for each exercise.
- The "total" score for each exercise should also be displayed as a separate (immutable) item. Depending on the previous question this should or should not be present if no other items are added. In the case that you have to add at least one item to have a total, the total row should be replaced by a message that you won't be able to grade this exercise unless you add at least one rubric item.
- For me, it isn't super clear if the visibility checkbox only applies to the score of that item or the entire rubric item.
- Design-wise, the page is very busy. Some suggestions:
  - merge the cards for individual exercises
  - add the "add scoring item" button to a dummy row of the table or next to the exercise name, the same for the copy rubric button
- copy rubric is semi-broken. The "total" row is also copied.
- the total score of the evaluation (over all exercises) should also be displayed somewhere
- I would not show the top form automatically, but hide it behind a button. 
- Maybe a modal/dialog is a better UI for showing the add item form?

bmesuere:
image

- The average row should move to the header row because now it looks too similar to a "user row"
- Add the max value for an exercise to the avg row instead of placing it next to the exercise name (this save some h-space)
- can we easily make the t-body scrollable for evaluations with hundreds of users?
- I wouldn't use the color coded "bars" underneath the scores here. In a future version, I can image we'll want to color the cells based on the score.
- The difference between a final and temp value isn't super clear. The help text mentions "black" values, but they appear to be gray. Instead of showing a 0 for unfinished reviews, showing a `-` might be more appropriate? Are these placeholder zero's also taken into account when calculating the averages? They shouldn't be.
- Make it more clear which total scores aren't complete yet.

bmesuere:
- There should be a clearer explanation what will happen if a rubric is changed after the evaluation has already started
- Why don't we allow changing scores after a review as marked as done? Updating a score now requires the user to click the "undo" button, change the score, and then click finish again. Maybe instead add a small warning?
- Not too fond of the `0` and `10` icons since the 10 isn't literally 10. Maybe use thumb up and down instead?
- Add the min max buttons to the individual items as well?
- Do we already have icon buttons without button outline on Dodona? The outlines for min, max and trash are a bit busy for such a small space.
- The "finish" and finish+next buttons are disabled when not all scores are completed. This is not terribly clear, so a tooltip explaining why might be handy in that case. It creates another "friction" issue when you type in the last rubric item. The score is only saved "onblur" which means that you must first click outside of the input area to save the score and then need a second click to proceed. Can't we do something more clever with the buttons instead of disabling them?
- consider step=0.25 instead of 0.01. The help text when creating a rubric item might also mention this.
- I received this error in the console when "saving" a score on an evaluation with a single rubric item (not sure if that is relevant):
  ```
    actions.ts:139 Uncaught (in promise) TypeError: Cannot set property 'innerText' of null
    at FeedbackActions.setTotal (actions.ts:139)
    at eval (eval at _callee$ (score.ts:163), <anonymous>:4:31)
    at eval (<anonymous>)
    at _callee$ (score.ts:163)
    at tryCatch (runtime.js:45)
    at Generator.invoke [as _invoke] (runtime.js:274)
    at Generator.prototype.<computed> [as next] (runtime.js:97)
    at asyncGeneratorStep (actions.ts:225)
    at _next (actions.ts:225)
  ```

niknetniko:
I've applied the suggestions (except for those mentioned below and those I've moved to other issues).

> can we easily make the t-body scrollable for evaluations with hundreds of users?

I experimented a bit, but it was either a sticky header or a sticky first column, but both didn't really seem to work cleanly.

> Add the min max buttons to the individual items as well?

I've added the min button. For the max, you can click the max score text, which doubles as the max button for individual scores.

> Why don't we allow changing scores after a review as marked as done? Updating a score now requires the user to click the "undo" button, change the score, and then click finish again. Maybe instead add a small warning?

This was a deliberate choice, to enforce filling in all scores (and a sort of mental separation: if the feedback is completed, the scores are done). If we remove this restriction (which should be possible, there is no technical reason to disallow this), we probably also want to either:
- Allow marking a review as completed with empty scores
- Only allow changing, but not deleting, scores if the review is completed
- Automatically un-complete a review if a score is removed (I think this one is the most logical then?)

> The "finish" and finish+next buttons are disabled when not all scores are completed. This is not terribly clear, so a tooltip explaining why might be handy in that case. It creates another "friction" issue when you type in the last rubric item. The score is only saved "onblur" which means that you must first click outside of the input area to save the score and then need a second click to proceed. Can't we do something more clever with the buttons instead of disabling them?

I've changed to event to be "change", but that was unrelated  and doesn't fix this issue. I agree it is a friction, but I don't really see a clean solution at the moment.

bmesuere:
> > can we easily make the t-body scrollable for evaluations with hundreds of users?
> 
> I experimented a bit, but it was either a sticky header or a sticky first column, but both didn't really seem to work cleanly.

Have you checked position: sticky: https://css-tricks.com/a-table-with-both-a-sticky-header-and-a-sticky-first-column/

> > Add the min max buttons to the individual items as well?
> 
> I've added the min button. For the max, you can click the max score text, which doubles as the max button for individual scores.

I would also add the max button. The first rule of design is: it things behave similar, they should look similar and if things look similar they should have a similar function. 

> > Why don't we allow changing scores after a review as marked as done? Updating a score now requires the user to click the "undo" button, change the score, and then click finish again. Maybe instead add a small warning?
> 
> This was a deliberate choice, to enforce filling in all scores (and a sort of mental separation: if the feedback is completed, the scores are done). If we remove this restriction (which should be possible, there is no technical reason to disallow this), we probably also want to either:
> 
> * Allow marking a review as completed with empty scores
> * Only allow changing, but not deleting, scores if the review is completed
> * Automatically un-complete a review if a score is removed (I think this one is the most logical then?)

My opinion here is that the "mark review as done" concept was very much needed without the grades. Otherwise, there was no distinction between what has been seen en what hasn't. When working with scores, this has little added value since setting all scores more or less means it's done and if something is missing, it isn't.
 
> > The "finish" and finish+next buttons are disabled when not all scores are completed. This is not terribly clear, so a tooltip explaining why might be handy in that case. It creates another "friction" issue when you type in the last rubric item. The score is only saved "onblur" which means that you must first click outside of the input area to save the score and then need a second click to proceed. Can't we do something more clever with the buttons instead of disabling them?
> 
> I've changed to event to be "change", but that was unrelated and doesn't fix this issue. I agree it is a friction, but I don't really see a clean solution at the moment.

One way is not to actually disable it, or alternatively to update its state (not the scores) on keyup.

I can agree that this gets merged and deployed to gather feedback on the current implementation under these conditions:
- It is only enabled for us
- It will not be enabled for other users until a follow up PR addressing the remaining issues is merged
- Your best estimate is that we will not have to make database changes for the remaining issues that will conflict with the grading items that will be created for the evaluation

The alternative is to wait until it's done but that will mean that it can't be used for the ST evaluation. @chvp what do you think?

niknetniko:
> Have you checked position: sticky: https://css-tricks.com/a-table-with-both-a-sticky-header-and-a-sticky-first-column/

Implemented, but the table has its own scroll bar, as discussed.

One thing that is not done yet is some kind of shadow when scrolling (horizontally/vertically).

> I would also add the max button. The first rule of design is: it things behave similar, they should look similar and if things look similar they should have a similar function.

Added.

> My opinion here is that the "mark review as done" concept was very much needed without the grades. Otherwise, there was no distinction between what has been seen en what hasn't. When working with scores, this has little added value since setting all scores more or less means it's done and if something is missing, it isn't.

You can now always edit the scores, regardless of the completeness of the state.

> One way is not to actually disable it, or alternatively to update its state (not the scores) on keyup.

The button is no longer disabled.

app/assets/javascripts/score_item.ts
```diff
@@ -0,0 +1,78 @@
...
+                headers: {
+                    "Accept": "application/javascript",
+                    "Content-Type": "application/json"
+                },
+                body: JSON.stringify(dataProvider(checkbox.checked))
+            }).then(async response => {
+                if (response.ok) {
+                    eval(await response.text());
+                } else {
+                    // Someone already deleted this question.
```
chvp, line 42: question?

chvp (changes requested):
Only a few minor remarks. Good stuff :+1:

The scores edit page is pretty different from the scores page that is shown when the evaluation is created. Was that intentional? I like the new page more (especially the make all visible/invisible on the edit page don't look very nice).

config/locales/views/score_items/nl.yml
```diff
@@ -0,0 +1,60 @@
...
+        one: "Eén scoreonderdeel in deze evaluatie. Totaalscore: %{score}."
+        other: "%{count} scoreonderdelen in deze evaluatie. Totaalscore: %{score}."
+    new:
+      title: Scoreonderdelen configureren
+      explanation_part1: >
+        Je kan op de oplossingen in deze evaluatie ook punten ingeven op Dodona.
+        Daarvoor moet je wel eerst bepalen waarop je punten zal geven. Binnen Dodona
+        kan je per oefeningen een puntenverdeling opstellen door het toevoegen van scoreonderdelen.
+      explanation_part2: >
+        Hiernaast kan je snel één scoreonderdeel toevoegen aan elke oefeningen in deze evaluatie.
```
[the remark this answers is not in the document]
niknetniko: > "Hiernaast" also sounds weird to me, I would expect "Daarnaast". You're better at Dutch than me though, so I'll defer to your judgment.

No, you're 100% right. The "hiernaast" was from an iteration where the button was displayed next to the text.

db/schema.rb
```diff
@@ -210,6 +210,7 @@
     t.integer "exercise_id"
     t.datetime "created_at", precision: 6, null: false
     t.datetime "updated_at", precision: 6, null: false
+    t.boolean "visible_score", default: true, null: false
```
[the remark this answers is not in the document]
niknetniko: Yes, I had two scenarios in mind where this could be useful:

- The sum of the score item isn't how you want to calculate the sum (e.g. three score items as /10, but counted as 50%, 25%, 25%). Not showing the calculated total should prevent confusion.
- You have invisible score items you don't want to show. E.g. you want to show "correctness", but not "code style". If the total is visible, students could calculate the scores on the hidden score items.

This option has no impact on the visibility of the score items themselves.

niknetniko:
> The scores edit page is pretty different from the scores page that is shown when the evaluation is created. Was that intentional? I like the new page more (especially the make all visible/invisible on the edit page don't look very nice).

The intention was to make something similar to the user selection page (which has two slightly different pages), but there is no technical reason to do this: the new page could be used everywhere.

chvp:
> The intention was to make something similar to the user selection page (which has two slightly different pages), but there is no technical reason to do this: the new page could be used everywhere.

Making them slightly different seems good (we probably need less explanation on the edit page), but the two raw links don't look very nice IMO.

bmesuere:
**Creating score items**
- I'm still struggling with the option to continue with/without grading and if this is clear enough. Would it be an option to hide the bottom card and change the primary button to "continue without grading" and add a button with something like "add rubric" that shows the bottom part and changes the text of the primary button?
- The overflow menu is not correctly placed there. Maybe place the "page subtitle" inside the bottom card and add the overflow menu to the top right corner there? Maybe also add the "add item to all exercises" to the overflow menu instead of a fab (since this is not the primary page action).
- Exercises title seems a bit strange? I'm thinking about something better. Same for a better way to display the total score
- The empty state text in the table is perfect, but if you center it and color it as secondary text, it looks even better
- Consider moving the add and copy buttons to the "total" row of the tables. this way, they align with the edit and delete buttons.
- put total row in bold?
- right align Maximum column title
- "If the score becomes visible to students once the evaluation is released." -> "Make score item visible..."
- SAVE -> ADD

app/controllers/score_items_controller.rb
```diff
@@ -0,0 +1,109 @@
...
+      [I18n.t('evaluations.show.evaluation'), evaluation_path(@evaluation)]
+    ]
+  end
+
+  def preload_eval_exercise(item)
+    # Reload the evaluation, since one of the items changed.
+    @evaluation.score_items.reload
+    # Preload the stuff for one exercise and an existing instance.
+    ActiveRecord::Associations::Preloader.new.preload(item.evaluation_exercise, [score_items: :scores])
+    item.evaluation_exercise
```
chvp, line 107: Not a full review, but I was just wondering how difficult it would be to test these lines as well. IIUC they're only used for the js responses, but it seems like this isn't something we would want to break accidentally.

niknetniko: > Not a full review, but I was just wondering how difficult it would be to test these lines as well. IIUC they're only used for the js responses, but it seems like this isn't something we would want to break accidentally.

Yep that's the plan, it's just that I haven't done it yet

niknetniko:
Recent changes:
- I've applied the suggestions for creating score items; the new page is now using a method similar to the new course page[1]
- I've moved and stickied the total column in the grade table
- Changes to feedback page:
  - I've used input groups
  - The "completed" status is now automatically handled (and hidden) if there are score items for an exercise.

1:
image

bmesuere:
- I'm a big fan of the new stepper component. In the next version (later PR), I would also include the deadline and user selection in the wizard.
- Small issue: "Ja" is already in the "title" of the stepper before picking an option:
  image
- In the *puntenverdeling* card, the "to evaluation" button is now in the top right corner, but I think it would be better if it was in the bottom right in a separate section as we have with many other cards.
- The removal of the "mark as done" is a great improvement. Is it possible to also remove the " + *icon*" from the button? 
  image

This has been a marathon (over 100 commits!) but is looking really good. I will try some UI things in separate commits this afternoon, feel free to undo them.

bmesuere:
A mockup of a potential score UI:
image

niknetniko:
I've implemented a fix for the "mark as done" button, the answer, and applied the suggestion for the score UI:

image

app/models/score.rb
```diff
@@ -0,0 +1,43 @@
...
+  attribute :expected_score, :decimal
+
+  belongs_to :score_item
+  belongs_to :feedback
+  belongs_to :last_updated_by, class_name: 'User'
+
+  after_destroy :uncomplete, unless: :destroyed_by_association
+  after_save :maybe_complete_feedback
+
+  validates :score, presence: true, numericality: { greater_than: -1000, less_than: 1000 }
```
chvp, line 25: The numbers in this validation seem quite arbitrary?

app/views/score_items/new.html.erb
```diff
@@ -0,0 +1,117 @@
+<% content_for :javascripts do %>
```
chvp, line 1: GitHub doesn't seem to highlight this file, any idea why?

chvp (approved):

app/views/score_items/new.html.erb
chvp: Huh, now it does, never mind then.

app/models/score.rb
niknetniko: They are not totally arbitrary, they come from the database:

```ruby
t.decimal "score", precision: 5, scale: 2, null: false
```

Since we use precision 5 and scale 2, we have three digits available before the decimal separator, so the max/min value is +/- 999.99. The precision is rather arbitrary, but it seemed like a reasonable default.

chvp: Makes sense, 5 digits seems like enough.

bmesuere (approved):
I'm really impressed with the quality of the code 👍 
Some minor comments, none of them blocking:
- Update the screenshots in PR start for future reference (can also be done after merging and deploying)
- Just to confirm, who will be able to create a review with grading in this initial phase?
- For the public release, this will need a manual

niknetniko:
As discussed, I've

- moved the CSV export to the model
- limited creating score items (and the buttons/pages) to zeus for the internal test. This should be a nice self-contained commit, so we can easily revert it if we want to make it available globally.

I also updated the screenshots.

bmesuere (approved):

chvp (approved):

chvp:
🎉🎉🎉

Issue #2452, opened 2021-01-05
Reduce support load


bmesuere:
- [x] better default name for new institutions: use domain or smartschool url instead of blank. Add property to filter all institutions without a default name
- [ ] better error handling when signing in with the wrong sign in provider: link to correct provider for the given username/email
- [x] institutions index: add pages and potentially search
- [ ] institution show: edit provider types
- [ ] merge institutions
  - start from main institution
  - select another institution (search?)
  - show overview of courses/users for both institutions
  - provider of secondary institution must be linked to main institution
  - change institutions of users of secondary institution to main institution
  - change institution of potential courses of secondary institution to main institution
  - remove secondary institution
- [ ] merge users using console function

bmesuere:
I split this over several other issues.

PR #2504, opened 2021-02-19
Improvements to institution management

niknetniko:
Part of #2452, implements two of the suggestions:

- Adds pagination & search to institution page
- Better default names for institutions
   - Smartschool: subdomain (e.g. `name` in `[name].smartschool.be`)
   - GSuite: the domain itself (e.g. `example.com`)
   - Office365: the domain of the email of the user
   - Others (LTI, Saml): no change
   
- [x] Tests were added

### DIFF
--- a/app/assets/stylesheets/models.css.less
+++ b/app/assets/stylesheets/models.css.less
@@ -7,3 +7,4 @@
 @import "models/users.css.less";
 @import "models/questions.css.less";
 @import "models/programming_languages.css.less";
+@import "models/institutions.css.less";
--- a/app/assets/stylesheets/models/institutions.css.less
+++ b/app/assets/stylesheets/models/institutions.css.less
@@ -1,0 +1,3 @@
+.generated-name-icon {
+  color: @text-secondary;
+}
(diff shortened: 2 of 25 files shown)


bmesuere (commented):
Do we need an additional boolean to indicate that an automatic name is currently used instead of a manual one? Otherwise it will be hard to find which ones might need an update?

chvp:
> Do we need an additional boolean to indicate that an automatic name is currently used instead of a manual one? Otherwise it will be hard to find which ones might need an update?

Agreed. This should also allow the sort to be done by that column first, allowing the custom SQL to be dropped. We should probably mark all institutions with the current default name with this boolean as well.

niknetniko:
Applied suggestions:

- By default, the `generated_name` column is `true`.
- Existing institutions with the default name are also `true`
- When the name is edited, the column becomes `false`.

I've also added an icon to indicate a generated name:
image

bmesuere (approved):
At least for g suite and smartschool it should be possible to convert the existing n/a's to a more meaningful name.

niknetniko:
> At least for g suite and smartschool it should be possible to convert the existing n/a's to a more meaningful name.

Do you mean this should be added as a migration to this PR? Or is this something for a future PR?

chvp (changes requested):
For sign in, we currently decide whether or not to show institutions based on whether their name is equal to the default name. This should be changed.

I would do the migration @bmesuere mentioned in this PR.

chvp (approved):

bmesuere (approved):
🚢

Issue #2529, opened 2021-03-08
Polling annotations is a slow query


rien:
The last few days we can notice a slight increase in database load. When looking at `show full processlist;` on our database server, the queries selecting unanswered questions pops up a lot:
```sql
SELECT `annotations`.* FROM `annotations`
   INNER JOIN `submissions` 
     ON `annotations`.`submission_id` = `submissions`.`id` 
   WHERE `annotations`.`type` = 'Question' 
     AND `submissions`.`course_id` = 123 
     AND `annotations`.`question_state` = 2 
    ORDER BY `annotations`.`updated_at` DESC
    LIMIT 10
    OFFSET 0;
```

- We should add an index to make this query faster (since this query is performed a lot).
- The `'type' = 'Question'` comparison feels like it is slow because it compares strings.
- Maybe we can think about caching this page?

bmesuere:
- only adding an index will not fix this. The problem is the join with the huge submissions table.
- type = question is not an issue if we create an index
- caching is not a good strategy since we explicitly poll the page for update

The only solution is to denormalize a bit and add the course_id to the annotations table. In combination with an index (and maybe slightly less aggressive polling?) this should improve response times.

bmesuere:
Regarding to polling, we now check every 2 seconds even if the page is hidden. If you have an tab in the background, this keeps the page active. As an alternative, we could implement a backoff mechanism:

- when scheduling the next event, use the set delay (2s) if the page is [visible](https://developer.mozilla.org/en-US/docs/Web/API/Document/hidden)
- when the page is not visible, add 1 second (?) to the delay
- if the page becomes visible (visibilitychange), cancel the scheduled callback, instantly execute the callback, schedule a new one

PR #2539, opened 2021-03-09
Implement back-off in InactiveTimout

niknetniko:
Implement the backoff "algorithm" described in #2529 when the page is inactive.

I've tested in Chrome, Firefox & Edge on Windows.

(For now I kept a `console.log` on line 160, since it makes checking if it actually works in the browser a lot easier. Will need to be removed before merging though).


chvp (commented):
I would propose adding a maximum to the delay.

chvp (approved):

bmesuere (approved):

PR #2541, opened 2021-03-10
Denormalize course_id into annotations

### LINKED ISSUE #2529: Polling annotations is a slow query

chvp:
Also add index to annotations for common queries.

Closes #2529.

### DIFF
--- a/app/models/annotation.rb
+++ b/app/models/annotation.rb
@@ -6,17 +6,19 @@
 #  line_nr            :integer
 #  submission_id      :integer
 #  user_id            :integer
-#  annotation_text    :text(65535)
+#  annotation_text    :text(16777215)
 #  created_at         :datetime         not null
 #  updated_at         :datetime         not null
 #  evaluation_id      :bigint
 #  type               :string(255)      default("Annotation"), not null
 #  question_state     :integer
 #  last_updated_by_id :integer          not null
+#  course_id          :integer          not null
 #
 class Annotation < ApplicationRecord
   include ApplicationHelper
 
+  belongs_to :course
   belongs_to :submission
   belongs_to :user
   belongs_to :evaluation, optional: true
@@ -36,6 +38,7 @@ class Annotation < ApplicationRecord
   scope :by_exercise_name, ->(name) { where(submission: Submission.by_exercise_name(name)) }
 
   before_validation :set_last_updated_by, on: :create
+  before_create :set_course_id
   after_destroy :destroy_notification
   after_save :create_notification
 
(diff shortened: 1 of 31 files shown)


bmesuere (approved):

app/models/course.rb
```diff
@@ -126,27 +126,28 @@ class Course < ApplicationRecord
...
+           inverse_of: :course
 
   has_many :answered_questions,
            lambda {
              where question_state: :answered
            },
-           through: :submissions,
-           source: :questions
+           class_name: "Question",
+           inverse_of: :course
```
chvp, line 150: These changes make sure the field is used for the query.

db/migrate/20210308102217_add_index_to_annotation.rb
```diff
@@ -0,0 +1,10 @@
+class AddIndexToAnnotation < ActiveRecord::Migration[6.1]
+  def change
+    add_column :annotations, :course_id, :integer
```
niknetniko, line 3: Is there a downside to also making this a foreign key?

app/models/annotation.rb
```diff
@@ -6,17 +6,19 @@
...
 #  evaluation_id      :bigint
 #  type               :string(255)      default("Annotation"), not null
 #  question_state     :integer
 #  last_updated_by_id :integer          not null
+#  course_id          :integer          not null
 #
 class Annotation < ApplicationRecord
   include ApplicationHelper
 
+  belongs_to :course
```
niknetniko, line 21: Should this be `belongs_to :course, optional: true` if the course is deleted?

db/migrate/20210308102217_add_index_to_annotation.rb
chvp: No, forgot about that, my bad.

bmesuere (approved):

niknetniko (approved):

PR #2590, opened 2021-04-06
include judge in internal error notification

ninewise:
Can be used to filter internal error mails about judges you're not
involved with.

- [x] Tests were added
- [ ] Documentation update can be found at dodona-edu/dodona-edu.github.io# (not relevant)


bmesuere (changes requested):
Great idea, but tests seem to fail :(

bmesuere (approved):

PR #2594, opened 2021-04-06
Add docker configuration

thepieterdc:
This pull request adds a docker configuration. This differs from #1425, because this version uses a docker-compose file, which makes it much more flexible to use.

The main goal of this configuration is to use it in integration tests in the plugins.

Additionally, a GitHub Action is added to publish this image after every PR merge.

image


rien (approved):
Looking good :+1: 

It should be mentioned that is not possible with this configuration to evaluate submissions (there is nothing to run delayed jobs).

niknetniko (approved):

thepieterdc:
@bmesuere Before merging, make sure the DockerHub username and access token are configured in the repository settings (same as what was done in https://github.com/dodona-edu/docker-images)

```
        username: ${{ secrets.DOCKER_USERNAME }}
        password: ${{ secrets.DOCKER_PASSWORD }}
```

bmesuere (approved):
Ok for now, but I think this would be better off in its own repo. We can always move it later.

bmesuere:
@ninewise can you check/review this PR?

ninewise (approved):
Don't have experience with docker compose, though.

thepieterdc:
> Don't have experience with docker compose, though.

`docker-compose` allows you to start both the database and the dodona container with one command: `docker-compose up`.

PR #2623, opened 2021-04-14
Fix more caching issues

chvp:
The cache could still be wrong if a user had not submitted to a series yet (or had not submitted after an exercise was removed), since the submissions to the series were used to determine the users for which the cache had to be cleared.

I've opted to change the caching mechanism so that it uses the timestamp a series was updated at when building the cache string. This could clear the cache a bit more than necessary (title and description changes will also trigger a cache clear), but I don't foresee this happening on a regular basis.

Note that this could also still fail: MySQL uses second precision for timestamps, so if a teacher changes a series, a cache store is triggered and the series is changed again all in the same second, a wrong value could be stored in the cache. This case seems exceedingly unlikely to me though.

- [x] Tests were added

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -57,6 +57,9 @@ Rails/HelperInstanceVariable:
   Enabled: false
 Rails/OutputSafety:
   Enabled: false
+Rails/SkipsModelValidations:
+  AllowedMethods:
+    - touch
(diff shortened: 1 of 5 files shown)


niknetniko (approved):

bmesuere (approved):
👍 seems like an elegant solution

Issue #2612, opened 2021-04-12
Set up a self hosted runner on mestra


bmesuere:
For enhanced security, we might want to set up a self hosted runner for certain tasks. A first step is to get the runner up and running (hah), next we can look at which additional workflows we might want to add. Some ideas:
- restart workers
- update images on all workers
- deploy naos
- deploy dodona
- "reset" naos every night

bmesuere:
Related:
- #347
- #1977

bmesuere:
I've looked into this, and there is some good news and some bad news.
The good news: Self hosted runners are extremely easy to set up. You just have to run a few commands and everything just works. The runner polls github, so no need to open ports and it updates itself.

The bad news is, they are basically unusable in public repositories. You use a registered self-hosted runner by adding "self-hosted" to the runs on key. What this means is that for a public repo with a configured self hosted runner:
- a user can fork your repo
- add a workflow that sets the self-hosted flag and runs on pull request
- create a pull request

This will cause the workflow, which can run arbitrary code, to run on your self hosted runner which is of course a huge security problem.

Currently, there is no way to restrict who can use your self hosted runners, so you can basically only use them in private repositories.

bmesuere:
After a discussion, the way forward seems this:
- use github runners as we do now
- lock down what the user account (with the github ssh key) can do on mestra to a set of scripts
- create a separate repository for those scripts

A first step would be to let mestra deploy itself instead of running capistrano on github.

I will close this issue.

PR #2345, opened 2020-10-16
Skip evaluation if submission is already queued or running

rien:
This pull request

- [ ] Tests were added
- [ ] Documentation update can be found at dodona-edu/dodona-edu.github.io#

Closes # .

PR #2634, opened 2021-04-21
Let mestra deploy itself

bmesuere:
This pull request runs the capistrano deploy on mestra instead of the github runner.

This is related with #2612 and https://github.com/dodona-edu/dodona-ansible/pull/115.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -23,7 +23,7 @@ jobs:
       uses: shimataro/ssh-key-action@v2
       with:
         key: ${{ secrets.MESTRA_KEY }}
-        known_hosts: "-"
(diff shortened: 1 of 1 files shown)


bmesuere:
The basic script seems to work. This still needs to be tested with additional ssh key restrictions on mestra before it can be merged.

rien:
I have changed the workflow to send the git ref over stdin and confirmed this works. Together with #2345 this should fix deployments completely.

I have also tested out the ssh-key restrictions and I was not able to execute anything other than the specified command.

bmesuere:
I will update the script on mestra with an intermediate check so that we can't deploy (a random branch) to production by accident.

chvp (approved):

PR #2637, opened 2021-04-23
Add recaptcha.net to frame_src on contact page

chvp:
Seems like google changed the frame origin of recaptcha.


bmesuere (approved):

niknetniko (approved):

rien:
We might want to consider moving to [hcaptcha](https://www.hcaptcha.com/) in the long term. It is a [drop-in replacement](https://docs.hcaptcha.com/switch) for recaptcha but more privacy-friendly. Cloudflare is already using this instead of recaptcha.

PR #2639, opened 2021-04-24
Add a "support us" page

### LINKED ISSUE #2613: Add "support us" page
Add a "support us" page with some information about donations and check with fondsenwerving before publishing. 
Also add a short section to the github readme with the same information and check if we can use GitHub Sponsors.

> Giften kunnen vanaf nu worden overgemaakt op rekeningnummer van het Universiteitsfonds
> 390-0965803-29 of IBAN: BE26 3900 9658 0329 (BIC: BBRUBEBB)
> Met als mededeling: Dodona (of mag ook het dossiernummer zijn WE_1027).

bmesuere:
This pull request adds a "support us" page.

- [x] Create page in Dutch
- [x] Dynamic numbers
- [ ] Get feedback from fondsenwerving
- [x] Translate in English
- [x] Update about page with tweaked text and add link to support us page
- [x] Add link to support us page in the footer
- [x] Add link to support us page on github through [funding.yml](https://docs.github.com/en/github/administering-a-repository/displaying-a-sponsor-button-in-your-repository)

Closes #2613


bmesuere:
This is a first draft of the page in Dutch. @pdawyndt do you have any feedback or suggestions? A next step would be to send this to fondsenwerving.

image

db/schema.rb
```diff
@@ -34,7 +34,7 @@
     t.index ["key"], name: "index_active_storage_blobs_on_key", unique: true
```
chvp, line 34: Changes to this file seem unrelated.

pdawyndt:
I've gone through the text. Seems very good. No further comments;

db/schema.rb
bmesuere: Probably because is ran a migration. I'll remove it!

bmesuere:
I suggest we deploy this with the next release and ask for feedback from DCOM by sending them the link.

@chvp I think I used the fastest way to get the metrics. If it would be too slow, we can look into caching the fragment.

chvp (approved):
I probably would have gone for full translated pages/partials instead of the long strings in the translation files, but I don't have a strong preference.

Apparently there is no test that checks whether the about page renders, so maybe that could be added?

PR #2646, opened 2021-04-26
Tweak Support Us url

bmesuere:
The previous url pointed to /support which looked a bit like it was a page where they could get user support from us. It now points to /support-us.


chvp (changes requested):
Test still needs to be updated

bmesuere:
> Test still needs to be updated

Done, I search for support_path and missed the support_url.

chvp (approved):

PR #2649, opened 2021-04-26
fix new metrics section

bmesuere:
This pull request fixes the number of schools that are shown and a typo in the text.


bmesuere:
Fixed by #2648

PR #2671, opened 2021-04-29
Make sure series#new 404s when course does not exist

chvp:
Note that the test tests for an exception. In production, this exception results in a 404.

I also removed the one remaining use of `Maybe` and removed the gem from our dependencies. IMO, Ruby is not a monadic language, so trying to force a Maybe monad in there seems unnecessary to me, especially since the introduction of the `&.` syntax. I can remove the second commit if people disagree. (Ping @rien, since you introduced both uses of `Maybe` in the codebase.)

- [x] Tests were added

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -121,9 +121,6 @@ gem 'rubyzip', '~> 2.3.0'
 # add request server timings to the devtools
 gem 'rails_server_timings', '~> 1.0.8'
 
-# Maybe in Ruby
-gem 'possibly', '~> 1.0.1'
-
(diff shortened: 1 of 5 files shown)


bmesuere (approved):
I agree with removing the gem

niknetniko (approved):

rien (approved):
I have no idea why I introduced the `Maybe` other than I did not now of `&`. `.tap` and `.yield_self` at the time. So I am definitely OK with removing that.

Issue #1421, opened 2019-10-20
Printing exercise descriptions gets screwed up


pdawyndt:
Since the introduction of iframes in exercise desciptions, printing those pages gets screwed up.

pdawyndt:
Same happens when printing the overview containing all exercise descriptions of a series. Actually, it seems as if the page layout in the printed version is too narrow.

charvp:
Can you post a screenshot/explain more what you think is wrong? Things look fine when I print to a PDF.

pdawyndt:
![image](https://user-images.githubusercontent.com/5736113/67157428-4435d500-f32c-11e9-8513-bd964e208a1f.png)

rien:
I have fixed the width issue with cc9b5761. However, I have bumped on another problem: the iframe-resizer library we use [doesn't support resizing PDF's](https://github.com/davidjbradshaw/iframe-resizer/blob/master/docs/troubleshooting.md#pdf-and-opendocument-files).

bmesuere:
@rien I think the unsupported thing is resizing iframes (only) containing a PDF, which is not something we do as far as I know.

bmesuere:
This seems to be fixed?

pdawyndt:
Not on my machine.

rien:
I have done some further investigation and printing seems to be broken in different ways in firefox and chromium. I guess while using iframe's we've entered the land of obscure render engine bugs which may not be easy to solve.

However, printing seems to be good when going to the iframe's location (e.g. https://medusa.ugent.be/nl/exercises/1281456553/description/adleTCAeEJlDEEj8/).

I suggest closing this issue because the effort is probably not worth the reward.

pdawyndt:
Is there a way we could automatically redirect the printing to medusa from dodona? I guess we don't want users to delve into the iframe content themselves?

niknetniko:
> Is there a way we could automatically redirect the printing to medusa from dodona?

This isn't possible; the best we can do is include our own print button on the page and/or intercept the Ctrl-P shortcut. The print menu option from the browser is not modifiable, so when a user clicks the menu, it will always print the current page (this is how e.g. Google Sheets works).

We can apply the CSS fix for the width, but Chromium & Safari will still have semi-broken results, and Firefox will be completely broken (the [relevant bug](https://bugzilla.mozilla.org/show_bug.cgi?id=113217) was opened 19 years ago, so I wouldn't count on it being fixed any time soon)

[...]

PR #2673, opened 2021-04-30
Use the sandbox domain for the tutor iframe

### LINKED ISSUE #2601: Move the tutor iframe to a sandboxed domain
We don't have a lot of faith in the JS code of the python tutor. Since it's shown in an iframe already, it would be good to host it from a separate domain, similar to exercise descriptions.

bmesuere:
This pull request adds the sandbox host to the global `dodona` JS object and uses that value to create the tutor iframe. No changes were made to the routes because the tutor is a bunch of html pages in the public map.

This should reduce the impact of xss bugs in the tutor code.

I tested this on the 3 pages where we use the tutor:
- the bottom of an exercise page
- a submission page
- an evaluation page

Downside: the tutorviz file uses a manual copy of the iframe lib. This will not be updated automatically by dependabot.

Closes #2601


bmesuere:
2 issues remaining:
- 1 not related with this PR: the sandbox url on mestra is set to mestra.ugent.be and we have no idea why
- loading the tutor from the sandbox domain fails because we call the load of the iframe from the parent page and this is not allowed. The question remains if we can manually allow this.

chvp:
The sandbox url is configured in the apache VHost using passenger_env. It was set to mestra.ugent.be because we did not yet have a sandbox domain at that time.

bmesuere:
This is ready for review and deployed on naos for testing. For now, I enabled the logging to console, but this should be removed before merging. I manually tested using chrome on the 3 pages where we display submissions.

There are a number of issues/strange things I encountered when developing this:
- The mestra sandbox domain (mestra-sandbox.dodona.be) was not added to the https certificate (@rien)
- To get this working, the iframe resize lib needs to be loaded on the parent page. After trying several things, the code to get this working was adding `require("iframe-resizer");` to the pythia submission pack, similar to how this was added to the exercise pack. However, the existing code contains 2 other ways of adding this library which didn't work for me. They may not work in the current code either be be leftovers of previous experiments:
  - The description pack contains `import { iframeResizerContentWindow } from "iframe-resizer";` and `window.iframeResizerContentWindow = iframeResizerContentWindow;` however we never call this function anywhere.
  - The series overview pack contains `import { iframeResizer } from "iframe-resizer";`. Like I said, this syntax didn't work for me.

chvp:
The descriptions on the series overview page are indeed broken, so that import should probably be changed as well. (Shows how much that page is used, if we didn't know this was broken for a year.)

The `iframeResizerContentWindow` is required by iframe-resizer to be able to resize the iframe. See https://github.com/davidjbradshaw/iframe-resizer/blob/master/docs/troubleshooting.md#iframe-not-resizing. Maybe a comment could be added? I'm not sure if the `window` assignation is required or just the import.

chvp (approved):

pdawyndt:
@chvp One of the use-cases for the series overview page was printing all exercises in a series at once, but that feature itself has been broken since the introduction of the description iframes (#1421). Chain reaction.

niknetniko (approved):

public/tutorviz/javascript/iframe.min.js
```diff
@@ -0,0 +1,9 @@
+/*! iFrame Resizer (iframeSizer.contentWindow.min.js) - v4.3.2 - 2021-04-26
```
chvp, line 1: Actually, if this file has to be vendored (and I don't see a better solution immediately), maybe a comment should be added to package.json to update this file if the iframe-resizer dependency is updated.

bmesuere: JSON doesn't support comments :(

chvp: Right, that's unfortunate.

PR #2677, opened 2021-05-01
Add additional deploy scripts

### LINKED ISSUE #347: update docker images from website
_[Original issue](https://github.ugent.be/dodona/dodona/issues/347) by @ninewise on Fri Dec 23 2016 at 14:42._

### LINKED ISSUE #1977: Add functionality to restart workers without deploy
- Fix the UFW rule: There was an issue with a UFW rule related to docker. This caused the UFW setup to fail which in turn caused all outgoing network traffic to be blocked. Booting was then stuck on trying to mount the dodona nfs
- Check if we can set a timeout for mounting while booting. If dodona.ugent.be would be down, we can't reboot the workers (even if UFW would work)
- Next summer when we update the servers, install docker on the webserver. This way, we could start 1 or 2 workers there to evaluate submissions in case of emergency.
- Check if we can restart the workers without having to do a deploy

bmesuere:
This pull request adds the possibility to 
- deploy naos from github, using a manual dispatch (with approval from core member)
- restart workers from github, using a manual dispatch (with approval from core member)
- redeploy images from github, using a manual dispatch (with approval from core member)

Related: https://github.com/dodona-edu/dodona-ansible/pull/119

Closes #1977, closes #347


bmesuere:
- The updated mestra script was tested using github
- The naos script was tested by running it directly on mestra (instead of from github) because the manual triggers aren't available till this is merged in develop.
- The images and worker scripts were not tested on naos, waiting for review (and someone else being standby) for this.

.github/workflows/deploy-mestra.yml
```diff
@@ -1,12 +1,12 @@
...
 
 on:
   pull_request:
     types:
       - labeled
 
 jobs:
-  mestra:
+  deploy:
     if: ${{ github.event.action == 'review_requested' || (github.event.action == 'labeled' && github.event.label.name == 'deploy') }}
```
chvp, line 10:
```suggestion
    if: ${{ github.event.action == 'labeled' && github.event.label.name == 'deploy' }}
```

chvp (approved):

PR #2691, opened 2021-05-05
Move checkbox to the left in the search bar options

beardhatcode:
This makes it consistent with the actions

before:
image

after:
image

- [x] Tests were added -> not needed
- [x] Documentation update can be found at dodona-edu/dodona-edu.github.io# -> not needd


chvp:
I'm not sure I agree. The checkbox is on the right to differentiate it from an icon, because it's not an icon, it's a checkbox.

beardhatcode:
The goal was to align the checkboxes, when there were multiple
image

bmesuere:
Unfortunately, checkbox menu items are still not supported in bootstrap 5, so we'll keep them at the right for now.

PR #2692, opened 2021-05-05
Explore Lit

bmesuere:
This pull request serves as exploration of [Lit](https://lit.dev/), a light-weight web component framework by Google and successor of the Polymer project.

This could help us to
- create reusable UI components which sometimes now are rails partials. Intuitively, they are easier to use than partials and because they live in JS land, they can also be created client side which is something we're missing now.
- more clearly separate complex components from the rest of the application. I'm thinking for example about the scoring UI where we have an input group with 3 buttons and input area. Simply bringing together the JS, html and css for this would simplify development.
- reuse existing components developed by others for example the [Material Design components](https://github.com/material-components/material-components-web-components) by Google or an [emoji picker](https://github.com/nolanlawson/emoji-picker-element#readme). Because they are simply html elements, it is very easy to incrementally add these. Just swap out existing buttons for a material button for example.

I created a PR to be able to gather feedback and more easily deploy this to production (to test in older browsers).

Below is everything I learned while experimenting.

## Adding it to the project
1. Add `lit` and `@babel/plugin-proposal-decorators` with yarn and enable the plugin.
2. Write a component and import it in the application pack to make it globally available.
3. Use it in any html page.

## Development
- Components are a single file containing the html, css and js. There are plugins like `eslint-plugin-lit` and `ts-lit-plugin` to help with type checking.
- I haven't found an easy way to develop a component without adding it to some page. https://storybook.js.org/ was mentioned on many websites but I haven't checked it out yet.
- How do we document the available properties and actions of a component?
- There is no good way to avoid collisions of element names. It is recommended to prefix your own elements.

## Testing
- Still unclear. It is typescript code so it should fit in with our jest tests, but we probably also want to do (automated) visual checks.
- Can Storybook also help with this?

## Deployment
- The code is written in typescript and thus parsed with our webpack/babel setup. Everything should work.

## Conclusions

- Lit seems a good way to create reusable components. There are some issues with shadow dom scoping, but using the shadow dom is not mandatory.
- The material design components by google are probably not good enough. The development pace is very slow and they warn against using it in production.
- We will need some kind of component library/framework

### Vuetify
- [Link](https://vuetifyjs.com/en/)
- pro: includes a lot of components
- pro: includes styles for typography and grid system
- con: not very lightweight
- con: very hard to incrementally add
- con: still waiting for the next major release

### Material Web Components
- [Link](https://github.com/material-components/material-components-web-components)
- pro: fits our style perfectly
- pro: easy to incrementally add
- con: very slow development and incomplete (e.g. no card component)
- con: doesn't include general typography and grid system

### Bootstrap 5
- [Link](https://getbootstrap.com/docs/5.0/getting-started/introduction/)
- pro: includes general typography (in rem!) and grid system
- pro: a lot has changed since v3, but at the same time, many utilities will keep working
- con: no webcomponents
- con: still migrating to css variables so still needs sass/less compilation

### ionic framework
- [link](https://ionicframework.com/docs/components)
- pro: lots of components, with active development
- pro: material style
- con: seems focussed on mobile apps
- con: no typography/grid

### Others
- [Vaadin](https://vaadin.com/docs/latest/ds/overview): many components, not really material
- [Shoelace](https://shoelace.style/): single developer
- [Adobe](https://opensource.adobe.com/spectrum-web-components/#:~:text=LitElement): many components, not material


bmesuere:
Minimal additions are:
- Add Lit package with yarn
- Add extra babel plugin to support decorators
- Require the component file in the application pack and simply use it somewhere in the html

bmesuere:
We migrated to bootstrap 5.

PR #2700, opened 2021-05-10
Migrate to Bootstrap 5

bmesuere:
This pull request migrates our UI framework to Bootstrap 5. This has a number of benefits:
- We get rid of bootstrap 3 including *less* and its security vulnerabilities
- By using sass, we can enable css compression
- We can add it as a node package (and no longer vendored) which makes it easier to update in the future
- A modern grid system based on flexbox, including additional tiers
- New components (toggle button, skeleton in 5.1)
- More utility classes

Downsides:
- Exercises and descriptions using bootstrap specific css classes might break. Where possible, backward compatible aliases were added.

A previous analysis of why we made this change can be found in #2692 

---

- [x] Remove vendored Bootstrap 3
- [x] Migrate from less to sass for the remaining files and compile assets without errors
- [x] Add Bootstrap 5 as a dependency and load the code
- [x] Tweak obvious var name changes and reenable mixins
- [x] Check if we want to keep the new bootstrap scaling
- [x] Follow the migration guide for 3 -> 4 and update views
- [x] Follow the migration guide from 4 -> 5 and update views
- [x] Check and update all remaining views
- [x] Fix light and dark styles
- [x] Fix .box-shadow mixin
- [x] Fix &:extend(.btn-icon) in btn.css
- [x] Fix .transition in card.css
- [x] Fix ~calc in table.css, navbar.css, course.css, layout.css, series.css
- [x] fix push and pull classes (migration4)
- [x] Responsive navbar behaviors are now applied to the .navbar class via the required .navbar-expand-{breakpoint} where you choose where to collapse the navbar. Previously this was a Less variable modification and required recompiling.
- [x] Panel is removed an used in the stepper
- [x] hidden class was removed
- [x] Fix carets for dropdowns
- [x] Navbar
- [x] Migrate "hidden-*" classes
- [x] typeahead is broken on sign in page
- [x] Dropdown in filter bar is still broken
- [x] Tabs are broken
- [x] scrollspy is still broken
- [x] label styling is missing
- [x] Modal for label edit is broken
- [x] Default buttons have the wrong color
- [x] Feedback table layout needs some tweaks
- [x] Alerts in dark mode should only have a border and no background
- [x] Tables seem so have a thin border at the bottom/right
- [x] Pagination links are broken
- [x] Font color on series drop down is gray


bmesuere:
This is ready for review with 2 issues remaining:
- scrollspy is broken on the course page: I will make a last attempt, but switching to a small dedicated library might also be an option
- dark mode is broken: The trick of using string interpolation to import the right file doesn't work with sass. I'll think about a better solution, but worst case we will have to copy the "base" file into both the dark and light theme file. I also didn't test the dark mode colors yet.

bmesuere:
Dark mode was fixed and a solution for the scrollspy issue can be found in #2718.

niknetniko (commented):
Things I noticed (most of which are very minor and non blocking) when checking the pages (I haven't looked at any code yet):

- [ ] Deprecation warnings in the dark theme (repeated a few times):

```
DEPRECATION WARNING on line 18 of /home/niko/projects/dodona/app/assets/stylesheets/theme/theme-dark.css.scss:
!global assignments won't be able to declare new variables in future versions.
Consider adding `$brand-primary: null` at the top level.
```
- [x] There is no indication that a field is a dropdown in Chrome (e.g. the course edit page):
image
- [ ] The question checkbox is no longer aligned (course edit page)
image
- [ ] Buttons are no longer aligned with edit field (course edit page, but also e.g. profile page, scores are fine though)
image
- [ ] Edit labels modal is messy:
image
- [ ] Images in descriptions are different (not aligned, caption is always shown), e.g. the curling exercise (compare with https://dodona.ugent.be/nl/activities/2051796188/):
image
- [ ] Code blocks have no more background (same exercise):
image
- [ ] Evaluation status table is weird
image
- [x] Clicking grade table tab does not remove underline from evaluation status tab (this does work in production and in other places):
image
- [ ] The sidebar is not wide enough:
image
- [ ] Programming language icon is not aligned correctly on programming language edit page:
image
- [ ] This is ugly in dark mode (course edit page):
image
- [ ] accordions seem broken, the styling and functionality (can't click them), (e.g. in the course description, but also in exercise descriptions):
image
- [ ] Repository page has wrong alignments:
image

chvp:
Only thing I found that Niko didn't:
screenshot_2021-05-17-115653

chvp (commented):
Forgot to do an actual review that can be re-requested, sorry for the spam.

bmesuere:
- deprecation warning can be ignored because these variables are all defined at a higher level. The !global must be added to escape scoping.
- button alignment is due to default browser font size and bootstrap 5 using rem. We probably want this set to 14px in a future PR.
- sidebar: I didn't change the width, was 240px previously as well. Maybe this also has something to do with your font size? In chrome the overflow ... is applied correctly
- accordions: I fixed the styling (an accordion must have an accordion id), but to make them work, the data attributes need to be changed from data-toggle and data-parent to data-bs-toggle and data-bs-parent. No idea how to proceed here. Run a regex on all course, series and exercise descriptions?

app/assets/stylesheets/application-dark.css.scss.erb
```diff
@@ -0,0 +1,8 @@
+<% if Rails.env.development? || Rails.env.test? || Rails.env.staging? %>
+  //$development: true;
+  $development: false;
```
niknetniko, line 3: Should this be `false`?

app/assets/stylesheets/application.css.scss.erb
```diff
@@ -0,0 +1,8 @@
+<% if Rails.env.development? || Rails.env.test? || Rails.env.staging? %>
+  //$development: true;
+  $development: false;
```
niknetniko, line 3: Same here, should this be `false`?

app/assets/stylesheets/components/link.css.scss
```diff
@@ -2,5 +2,7 @@ a.anchor {
   display: block;
   position: relative;
   top: -90px;
+  //width: 1px;
+  //height: 1px;
```
niknetniko, line 6: Are these needed?

niknetniko (changes requested):
For non-Python exercises, there is some padding between the status color bar and the other stuff:

image

app/assets/stylesheets/application-dark.css.scss.erb
bmesuere: This was set to false for the moment (and maybe the next few weeks) to be able to estimate the effects of the css changes on the "production colors" on naos, mestra and localhost.

niknetniko: Makes sense

bmesuere:
> For non-Python exercises, there is some padding between the status color bar and the other stuff:
> 
> image

This is now fixed and deployed on naos.

niknetniko (changes requested):
While checking on naos, I've noticed two more things:

- The next/previous on reading exercises are now all the way to the left and to the right:
image
- The editor (or mark as read) is now always below the page visibility, so you always need to scroll, even if the exercise description is not longer than one page (example https://naos.ugent.be/nl/courses/61/series/645/activities/126856420/ and https://naos.ugent.be/nl/courses/61/series/645/activities/615906596/):
image
(screenshot is a bit difficult to make)
This seems to be a consequence of the `height` being set on the `.activity-sidebar .nav`.

niknetniko (changes requested):
The fix for the sidebar height breaks the stickyness of the sidebar for me.

Some poking around in Chrome suggests this could be fixed by adding something like this to `.activity-sidebar, .series-sidebar` (unless another fix is beter of course):

```scss
position: sticky;
top: $navbar-height + 20px;
```

niknetniko (approved):

PR #2718, opened 2021-05-16
Add custom scrollspy implementation

bmesuere:
This pull request adds a custom ScrollSpy implementation. The one included in bootstrap 5 doesn't work because of our anchor trick. This implementation is based on one I found on github (credits included in the file) which I adapted to work on our page. More specifically, I check if the parent of the parent is in the viewport instead of the anchor itself.

I made a few tweaks to improve performance which is now acceptable. The worst part is fetching the sections on each scroll event instead of doing this just once. The reason for this is that we modify the page with our dynamic loading. This could be optimized further by making the series load code manually call an update function, but for now this doesn't seem to be worth it (in fact, the code that is call onScroll to check if we need to load a new series is a low slower).

Finally, I converted the code to typescript.

### DIFF
--- a/app/assets/javascripts/scrollspy.ts
+++ b/app/assets/javascripts/scrollspy.ts
@@ -3,21 +3,18 @@
  * under the MIT license
  **/
 export class ScrollSpy {
-    /**
-    * Initialises a new {@code ScrollSpy} instance.
-    * Supported options are:
-    * {
-    *  sectionSelector: string;
-    *  targetSelector: string;
-    *  offset: number;
-    *  hrefAttribute: string;
-    *  activeClass: string;
-    * }
-    *
-    * @param {string|HTMLElement} menu - Selector to nav menu.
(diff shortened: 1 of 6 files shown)


app/assets/javascripts/scrollspy.ts
```diff
@@ -0,0 +1,142 @@
+/**
+ * This file was adapted from https://github.com/sidsbrmnn/scrollspy
```
niknetniko, line 2: I would mention the version (or date or commit) of the upstream library we've used. That way we don't need to guess in the future should the need ever arise to backport things from upstream (although this is very unlikely in this case).

bmesuere:
```suggestion
 * This file was adapted from https://github.com/sidsbrmnn/scrollspy version 1.0.4
```

niknetniko (approved):

PR #2719, opened 2021-05-16
Unvendor flatpickr

bmesuere:
This pull request removes flatpickr from the vendor directory and adds it using yarn.
This also properly imports it instead of making it available globally.


app/assets/stylesheets/libraries/flatpickr.css.scss
```diff
@@ -1,3 +1,5 @@
+@import "../../../../node_modules/flatpickr/dist/themes/light";
```
chvp, line 1: I presume that if there is a light theme that there is also a dark theme? Can we/do we want to make the import dependent on the Dodona theme?

app/assets/javascripts/series.js
```diff
@@ -115,29 +116,11 @@ function initSeriesEdit() {
...
-                rangeSeparator: " tot ",
-                scrollTitle: "Scroll voor volgende / vorige",
-                toggleTitle: "Klik om te wisselen",
-                ordinal: function ordinal(nth) {
-                    if (nth === 1 || nth === 8 || nth >= 20) return "ste";
-                    return "de";
-                },
-            };
-            flatpickr.localize(Dutch);
+            options.locale = Dutch;
```
chvp, line 121: Does this have the same translations? I agree with not defining this here, but using the same translations in visualisations/date pickers/etc. also seems important to me.

app/assets/stylesheets/libraries/flatpickr.css.scss
bmesuere: There is a dark theme, but since sass doesn't have conditional imports (or string interpolation in paths), there is no easy way to pick the right file.

For our own dark mode, I used a work around where all styles in a theme stylesheet were wrapped with a mixin so I could conditionally use the mixin (and thus the styles). We can't do that here since this would require editing the theme file.

chvp: OK, seems fine then.

app/assets/javascripts/series.js
bmesuere: The localization that was used before in this file was a manual copy of the import, so it's at least an improvement. It doesn't seem to contain strange translations?

chvp: No, the translations definitely seem fine, but this made me think we might want to unify these type of translations.

niknetniko (approved):
With safari 14.1 (released 26/04/21) finally supporting date input pickers, we might be able to get rid of this completely in the future and use the native date pickers. But unvendoring 👍

app/assets/javascripts/series.js
bmesuere: yes, true, I'll look into that with the new visualisation PR from @tibdhond

PR #2721, opened 2021-05-17
Bump ruby to 3.0.3

chvp:


bmesuere (commented):
I probably also asked with the previous update, but doesn't this also require an ansible update?

chvp:
Ah yes, that also needs to happen.

chvp:
https://github.com/dodona-edu/dodona-ansible/pull/121

bmesuere:
@jorg-vr this PR should be updated to be able to merge this next week. A deploy must then happen in concert with a server-side ruby upgrade.

bmesuere (approved):
There are a number of gems that are bumped. Was this by accident or part of the Ruby bump?

bmesuere:
Pinging @chvp for approval since review doesn't work.

jorg-vr:
@bmesuere  they seem to be caused by my last commit. I ran `bundle install` and `bundle update`, should I avoid bundle update?

chvp:
`bundle update` is generally not necessary, since dependabot updates our dependencies for us. Changes look good to me.

Issue #2695, opened 2021-05-06
Small grading bugs


chvp:
NOTE: Only use this issue for bugs. Feature requests should get a separate issue.

chvp:
Total average is calculated incorrectly (it is the average of exercise averages instead of the sum).

![E0sCuuOXsAI6mxd](https://user-images.githubusercontent.com/42220376/117261501-819b7e00-ae50-11eb-9f5c-ef4a6709d83e.jpg)

bmesuere:
Typo in Dutch of the score edit page:
![image](https://user-images.githubusercontent.com/481872/117261699-be677500-ae50-11eb-94a0-b1633b93ffc3.png)

(better UI also welcome)

bmesuere:
The score overview looks very cluttered compared to the other overview. This is due to the uneven column widths:
![image](https://user-images.githubusercontent.com/481872/117261863-eb1b8c80-ae50-11eb-9a0c-bd00b6ee550c.png)
vs
![image](https://user-images.githubusercontent.com/481872/117261914-0090b680-ae51-11eb-9912-84a6a73868c6.png)

pdawyndt:
- [ ] missing tooltip for thumbs up/down button; only "total" thumbs up/down buttons have a tooltip, but since the meaning of these buttons might not be 100% clear upon first use, it would be logical to also give the buttons for the individual score items a tooltip

pdawyndt:
@rien: What about a (subtle) icon per score item that is either grey (empty value), orange (modified value that is being saved), green (valid value, saved) or red (invalid value, saved)?

chvp:
Exercises linked in student view of evaluation don't link into course scope, resulting in an unauthorized error for students.

chvp:
Seems like we get a race condition when submitting scores sometimes:
```
  Mysql2::Error: Duplicate entry '23-210354922' for key 'scores.index_scores_on_score_item_id_and_feedback_id'
  app/controllers/scores_controller.rb:10:in `block in create'
```

PR #2744, opened 2021-05-21
Fix some grading bugs

niknetniko:
This pull request fixes two items from #2695.

- Adds tooltips to min/max buttons
- Refresh the user feedback row when adding grades (so the row is updated as completed when all scores are filled in)
  I've extracted this row to a new partial, and added the partial to the js update code.

### DIFF
--- a/config/locales/views/feedbacks/en.yml
+++ b/config/locales/views/feedbacks/en.yml
@@ -37,7 +37,8 @@ en:
       give_max_all: "Assign the maximal score to all score items"
       give_zero_all: "Assign 0 to all score items"
     score_actions:
-      give_one: "Assign %{score} to this score item"
+      give_max_one: "Assign %{score} to this score item"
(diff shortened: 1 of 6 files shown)


config/locales/views/feedbacks/en.yml
```diff
@@ -37,7 +37,8 @@ en:
       give_max_all: "Assign the maximal score to all score items"
       give_zero_all: "Assign 0 to all score items"
     score_actions:
-      give_one: "Assign %{score} to this score item"
+      give_max_one: "Assign %{score} to this score item"
+      give_zero_one: "Geef 0 aan dit scoreonderdeel"
```
bmesuere, line 41: Should be English here

bmesuere (changes requested):

chvp (approved):

bmesuere (approved):

Issue #1957, opened 2020-04-30
Change "annotations" in "comments"


pdawyndt:
The term "annotation" might seem OK in English, its counterpart "annotatie" in Dutch sounds very weird. Therefore I suggest to change the terminology in the UI into "comment/commentaar". Also results in more natural wording when talking about the annotations in documentation and other descriptive text.

pdawyndt:
Another option would be "opmerking" in Dutch.

PR #2776, opened 2021-05-25
Rework how grading results are shown to students

### LINKED ISSUE #2743: Improve discoverability of grading details for students
As this was mentioned a few times by students.
Perhaps we can add an icon next to it or something like that.

niknetniko:
This pull request adds a chevron to indicate there are score details:

image

It changes when open:
image

Closes #2743.

### DIFF
--- a/config/locales/views/evaluations/en.yml
+++ b/config/locales/views/evaluations/en.yml
@@ -80,7 +80,7 @@ en:
         one: "One annotation"
         other: "%{count} annotations"
       no_submission: "You had no timely submission for this exercise"
-      invisible_total: "The total score is not visible"
+      invisible_total: "The total score was not released"
       no_grading: "This exercise is not graded"
     members_table:
       remove_user: Remove student from evaluation
--- a/config/locales/views/evaluations/en.yml
+++ b/config/locales/views/evaluations/en.yml
@@ -73,12 +73,15 @@ en:
       released: Feedback was added to your code.
       exercise: Exercise
       no_annotations: "# annotations"
+      feedback: "Feedback"
       grade: Grade
       annotation_count:
-        zero: "No feedback was given"
+        zero: "No annotations"
         one: "One annotation"
         other: "%{count} annotations"
       no_submission: "You had no timely submission for this exercise"
+      no_grades: "None"
+      invisible_total: "The total score is not visible"
     members_table:
       remove_user: Remove student from evaluation
       remove_user_consequences: "If this student already received feedback, it will be deleted. Are you sure you want to remove this student?"
--- a/config/locales/views/evaluations/en.yml
+++ b/config/locales/views/evaluations/en.yml
@@ -81,7 +81,7 @@ en:
         other: "%{count} annotations"
       no_submission: "You had no timely submission for this exercise"
       invisible_total: "The total score was not released"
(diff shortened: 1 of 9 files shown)


bmesuere (approved):
Is there a reason to work with a collapse here? If I'm correct, this is only shown to students on a page that is dedicated for this, so why not show everything by default?

pdawyndt:
@bmesuere The scores are also shown to the students in the header of the submission and there it seems OK to collapse the scores. But I agree that collapsing seems usesless on a dedicated page. Also, the first row of the table shows the total score but this repeated information (it is already next to the chevron that was clicked) and might be removed.

Is the option to expand also provide in the special case (but an option that seems a reasonable choice) where only the total score is visible for the student?

niknetniko:
As discussed, the evaluation overview for students has been modified to always show the grades if the evaluation is graded:
image

For the total score row:
- When the total score is not visible, you see a "-" with the shown tooltip.
- When the evaluator didn't grade your submission for some reason, you'll get "- / [max, e.g. 10]"
- When the exercise isn't graded, you get "Geen"
- The subtext is the old text about the annotations / no valid submission.

If the evaluation is not graded or all grades are private, there is no change.

Another suggestion was applying #1957, but I think that's out of scope for this PR; we probably should change this all at once.

bmesuere (changes requested):
I played around with this on mestra and I think it's a great improvement. Some smaller remarks.

- The order of the exercises in the overview does not match the order in the evaluation.
- If one of multiple items is hidden, the sum of the sub scores doesn't match with the total score. Not sure if we want to fix it. I would guess most teacher will show all or nothing. Potentially we could add "niet alle deelscores werden vrijgegeven" to the total row if this is the case?
- If an exercises was not a graded exercise, I would not show "geen", but also a dash. In that case, maybe also put something else instead of "totaalscore"?
- Since we have enough space, convert the "no annotations" into a longer sentence?
- Is there a reason why you didn't put "score" in the table header for that column?
- "De totaalscore werd niet vrijgegeven" is maybe a little clearer than "De totaalscore is niet zichtbaar"?
- I wasn't sure about the align-top of the score and chevron. At first, I thought that middle looked better, but then I realized that they are aligned with the name of the score item (when using top) which probably looks better if there's a multiline description. That reason isn't valid for the chevron, since it applies to the entire row, but then is wouldn't align with the top-aligned score. So the current solution probably is the best.
- I would emphasize the total row even more. On the left of the table, this is clear already, but when looking at the right, the totals aren't standing out. One way to do this is by making them bold, eg:
  image

config/locales/views/evaluations/en.yml
```diff
@@ -73,12 +73,15 @@ en:
...
+      feedback: "Feedback"
       grade: Grade
       annotation_count:
-        zero: "No feedback was given"
+        zero: "No annotations were added to your solution"
         one: "One annotation"
         other: "%{count} annotations"
       no_submission: "You had no timely submission for this exercise"
+      invisible_total: "The total score was not released"
+      no_grading: "This exercise is not graded"
```
bmesuere, line 84:
```suggestion
      no_grading: "This exercise was not graded"
```

bmesuere (approved):

chvp (approved):

Issue #2597, opened 2021-04-08
Personal Dodona accounts


bmesuere:
This is an analysis of what needs to be done to allow Dodona accounts with personal O365 or gmail addresses (and not a commitment to actually go forward with this).

Technical:
- Lift the current block on signing in with generic addresses
- Create an identity provider for both Office and Gmail and add an extra boolean indicating that this is a non-school provider (maybe add this flag to accounts as well?)
- Check if the different (hotmail/microsoft/outlook) accounts use the same provider
- Update the sign in page with dedicated buttons for personal accounts
- Show the terms of service on first sign in (actually, it wouldn't be a bad idea to do this for other new accounts as well)
- Add a setting to courses to allow registration with personal accounts, off by default
- Remove the generic providers from the list of schools in places when a user selects his/her school

Other:
- Will we explicitly promote this? Or launch silently to begin with? The all courses listing should probably be updated with a separate section that's visible for such accounts where we highlight "open" courses: programmeursleerling, youtube, r/data science, ...
- I would never allow a teacher status for personal accounts

jorg-vr:
> Lift the current block on signing in with generic addresses

I assume this setting is in office 365 and gmail (And thus not in our code base)? Can someone set this up for naos so I can build and test the other features?

>  Remove the generic providers from the list of schools in places when a user selects his/her school

I do not understand this requirement? `in places when a user selects his/her school` Does this mean the sign in page? or which places are we talking about? Do we want to remove the search bar? or just the name of the generic provider?

bmesuere:
@jorg-vr I think I meant the course settings. We don't want a course for "gmail" users.

PR #2779, opened 2021-05-27
A better "all courses" experience

### LINKED ISSUE #1118: search by institute label finds too many courses
According to the institute listing, HoGent has two courses. However, clicking the link yields for hits if searching is done through the institute label. Perhaps courses wirhout a linked institute are always included in the search results.

image

_[Original issue](https://github.ugent.be/dodona/dodona/issues/1118) by @pdawyndt on Mon Jul 01 2019 at 17:54._

bmesuere:
This pull request improves the "all courses" experience.

Right now, the "all courses" experience isn't great. Especially, because this is one of the first pages a new user encounters.
The cards are visually pleasing, but make it hard to scan if you're looking for something specific. In addition, the filter options can be confusing because filtering for an institution also lists courses for outside that institution.

To make it easier to scan the page, the course listing was reverted to a standard table instead of using the cards. To avoid confusion with the filter, 3 tabs are shown: "institution courses", "all courses" and "my courses". By default, the first tab is shown.

In addition, it also makes the courses page available for users who are not signed in.

Follow up issue: Featured courses (#2780)

**Before**
image

**After**
image

Closes #1118

### DIFF
--- a/app/views/courses/index.html.erb
+++ b/app/views/courses/index.html.erb
@@ -15,8 +15,9 @@
     </div>
     <div class="card course-search">
       <%= render partial: 'layouts/searchbar', locals: {institutions: Institution.all, eager: !filter_institutions} %>
-    </div>
-    <div id="grouped-course-cards-wrapper">
+      <div class="card-supporting-text">
+        <div id="courses-table-wrapper"></div>
+      </div>
     </div>
   </div>
 </div>
--- a/app/views/courses/index.html.erb
+++ b/app/views/courses/index.html.erb
@@ -19,9 +19,13 @@
       <div class="card-supporting-text">
         <div class="card-tab">
           <ul id="course-tabs" class="nav nav-tabs" data-baseurl="<%= courses_path %>" role="tablist">
-            <li role="presentation"><a href="#institution">Institution courses</a></li>
+            <% if @show_institution_courses %>
+              <li role="presentation"><a href="#institution">Institution courses</a></li>
+            <% end %>
             <li role="presentation"><a href="#all">All courses</a></li>
-            <li role="presentation"><a href="#my">My courses</a></li>
+            <% if @show_my_courses %>
+              <li role="presentation"><a href="#my">My courses</a></li>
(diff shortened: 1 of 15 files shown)


config/locales/views/courses/nl.yml
```diff
@@ -51,6 +51,10 @@ nl:
     index:
       title: Alle cursussen
       users: "Gebruikers"
+      exercises: "Oefeningen"
+      institution_courses: "%{institution} cursussen"
```
niknetniko, line 55: Technically, things like "UGent cursussen" should be one word, e.g. "UGent-cursussen".

app/views/courses/index.html.erb
```diff
@@ -14,20 +16,26 @@
...
     </div>
     <div class="card course-search">
-      <%= render partial: 'layouts/searchbar', locals: {institutions: Institution.all, eager: !filter_institutions} %>
-    </div>
-    <div id="grouped-course-cards-wrapper">
+      <div class="card-supporting-text">
+        <div class="card-tab">
+          <ul id="course-tabs" class="nav nav-tabs" data-baseurl="<%= courses_path %>" role="tablist">
+            <% if @show_institution_courses %>
+              <li role="presentation"><a href="#institution"><%= t '.institution_courses', institution: (current_user.institution&.short_name || current_user.institution&.name) %></a></li>
```
niknetniko, line 23: When there is no institution, this results in a relatively weird first tab "cursussen", which is confusing because the difference with the "Alle cursussen" tab is not clear. Perhaps not worth fixing now, but this will need to be reconsidered if personal accounts (#2597) and if we don't assign those an institution.

niknetniko (approved):
I think it's a big improvement usability-wise 👍

app/views/courses/index.html.erb
bmesuere: I fixed this by not showing the first tab if the user has no institution

niknetniko (approved):

chvp (approved):

PR #2794, opened 2021-06-03
Check for jquery usage with eslint

bmesuere:
This pull request adds the use of an eslint plugin to check for the usage of jquery functions. All rules are enabled as a warning so linting will not fail but your IDE will probably highlight problematic lines.

In addition, the autofix rule was used to fix some deprecated function calls.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,10 +1,18 @@
 {
-  "plugins": ["jest"],
+  "plugins": [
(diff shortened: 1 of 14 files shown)


niknetniko (approved):
Works in RubyMine:

image

chvp (approved):

PR #2795, opened 2021-06-03
Fix code tab not automatically loading during feedback

### LINKED ISSUE #2784: JavaScript to automatically switch to code tab on feedback page is broken
When the code tab is linked (e.g. when clicking the annotations icon in the submissions listing), it still works.

bmesuere:
This pull request fixes the tab-switching-to-code during feedback. The issue was caused because the call to switch the tab was made before the page was ready.

Closes #2784

### DIFF
--- a/app/views/feedbacks/show.html.erb
+++ b/app/views/feedbacks/show.html.erb
@@ -83,9 +83,11 @@
 </div>
 <% if @feedback.submission.present? %>
   <script>
-    window.MathJax.startup.promise.then(() => {
-      window.dodona.codeListing.setEvaluation(<%= @feedback.evaluation.id %>)
+      $(function () {
+        window.MathJax.startup.promise.then(() => {
+          window.dodona.codeListing.setEvaluation(<%= @feedback.evaluation.id %>);
+        });
         $(".nav.nav-tabs li:last-child a").tab("show");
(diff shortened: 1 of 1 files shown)


bmesuere:
@chvp you removed the jquery wrapping on this page on a previous commit (and replaced it with the mathjax stuff). Do you remember why?

niknetniko (approved):
git blame reveals the commit where this was moved outside the jquery initialisation: https://github.com/dodona-edu/dodona/commit/41186108f1d6edce7afda98a8308043db99b9c50#diff-3ba34bac78253446477e2912305423349b70955647532e99e0903383fe727f3bR76

chvp (commented):
The mathjax startup promise should only be resolved after the page has already fully loaded, so I don't really get how this fixes things.

bmesuere:
> The mathjax startup promise should only be resolved after the page has already fully loaded, so I don't really get how this fixes things.

The initial code didn't work because bootstrap 5 somehow delayed its jquery hooks which caused the `tab` call to fail. The original code failed consistently, while I haven't been able to crash the new code. I played with slowing down my connection speed in dev tools to simulate out of order loading etc.

chvp (approved):

PR #2843, opened 2021-06-10
Upgrade checkboxes and radios to bootstrap 5

bmesuere:
This pull request introduces bootstrap 5 styling for checkboxes and radio buttons. This results in a uniform experience on all platforms. In addition, it enables the use of toggle switches where checkboxes were previously used to enable things.

**Before**
image
image

**After**
image
image
image
image


app/views/courses/_form.html.erb
```diff
@@ -88,12 +90,10 @@
...
     <div class="field form-group row">
       <%= f.label :questions, t(".questions.title"), class: 'col-sm-3 col-form-label' %>
-      <div class="col-sm-6">
-        <div class="checkbox">
-          <%= label_tag do %>
-            <%= f.check_box :enabled_questions %>
-            <span class="checkbox-label"><%= t ".questions.toggle-label" %></span>
-          <% end %>
+      <div class="col-sm-6 pt-2">
+        <div class="form-check <%= "form-switch" unless course.new_record? %>">
```
niknetniko, line 94: Should this be a switch? The reason I'm not 100% sure here is that in most places, the switch is immediate, while here the form has to be saved (and I'm not entirely sure why it's only a switch when it's not a new course).

The md spec [seems to support this](https://material.io/components/switches#behavior):

> When a user toggles a switch, its corresponding action takes effect immediately.

niknetniko (approved):

app/views/courses/_form.html.erb
bmesuere: It is a checkbox when creating the course (since it is a form there), but a switch when editing since it is a "mode" that gets enabled. But you are right that this is technically not instant and needs to be saved first.

niknetniko: That does make sense 👍

chvp (changes requested):
https://mestra.ugent.be/nl/courses/5/submissions/ -> automatic refresh should be a toggle like dark mode
https://mestra.ugent.be/nl/courses/5/series/50/activities/65117396/submissions/ -> most recent correct per user should also be a toggle

bmesuere:
I wasn't sure about those. They feel more like forms and not a mode, especially the most recent solution one. Therefore, I opted to keep them as checkboxes to not mix the icons (and add a separate setting for them). This is maybe something we can revisit if we implement a better filtering solution?

chvp (approved):

PR #2844, opened 2021-06-10
html-encode the text of machine annotations

### LINKED ISSUE #2050: Improper encoding of linter messages
On [this feedback page](https://dodona.ugent.be/nl/feedbacks/64494972/) the linter message should read as `if test < low` but has been stripped to `if test`. We should check if this stripping happens by Dodona or by the judge.

image

bmesuere:
This pull request fixes an issue where machine annotations containing a `<` were not properly rendered. The cause was that the text of the annotations was injected as html into the page. This PR add html encoding to the text of machine annotations and adds a test for this.

This might break the annotations of linters outputting html instead of text.

Closes #2050

### DIFF
--- a/app/assets/javascripts/code_listing/machine_annotation.ts
+++ b/app/assets/javascripts/code_listing/machine_annotation.ts
@@ -9,9 +9,13 @@ export interface MachineAnnotationData {
 export class MachineAnnotation extends Annotation {
     constructor(data: MachineAnnotationData) {
         // Filter out lines only containing dashes.
-        const text = data.text.split("\n")
+        let text = data.text.split("\n")
             .filter(s => !s.match("^--*$"))
             .join("\n");
+        // use the dom engine to encode the text to html
+        const node = document.createElement("span");
+        node.textContent = text;
+        text = node.innerHTML;
         super(data.row + 1, text, data.type);
     }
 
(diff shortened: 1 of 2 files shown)


bmesuere:
Not that I know of. 

Annotations were previously handled by the editor (in read only mode) which probably converted the text to valid html.

chvp:
I think we also listed them above the editor.

niknetniko (approved):

bsels:
> Do we know of linters outputting HTML? How were the messages included before this was reworked last year?

The TESTed judge produce HTML to link to external sources.

bmesuere:
> > Do we know of linters outputting HTML? How were the messages included before this was reworked last year?
> 
> The TESTed judge produce HTML to link to external sources.

I think those will be broken then or all other judges should switch encode their text 😞

bsels:
@niknetniko has an idea to in corporate a single link in these message.
See https://github.com/dodona-edu/universal-judge/issues/210#issuecomment-823868678.

niknetniko:
> @niknetniko has an idea to in corporate a single link in these message.
> See [dodona-edu/universal-judge#210 (comment)](https://github.com/dodona-edu/universal-judge/issues/210#issuecomment-823868678).

Basically we could add a `link` or `external` or `info` (the name is debatable) to the annotation data, which would contain a URL, and leave the rendering up to Dodona

PR #2849, opened 2021-06-12
Add featured courses

### LINKED ISSUE #2780: Featured courses
Add the concept of "featured courses" to Dodona which can be used to highlight specific courses such as De Programmeursleerling or the Youtube course and exercises of @deepeebee. In addition this would help avoid confusion caused by copies of those courses.

Technically, a featured course would be a course where an additional boolean was set by zeus-level users. These featured courses could then get a separate tab on the all courses page (see #2779) and would be listed first in the course copy wizard. In addition, we could consider using these to improve the empty state on the home page of new users.

To discuss: how do we decide which course becomes a featured course

bmesuere:
This pull request adds the concept of "featured courses" to Dodona which can be used to highlight specific courses such as De Programmeursleerling or the Youtube course and exercises of @deepeebee. In addition this would help avoid confusion caused by copies of those courses.

`featured` is a course flag that can only be set by zeus users. Featured courses are listed in a separated tab on the all courses page and are listed first in the copy course listing.

image
image
image
image
image

Closes #2780

### DIFF
--- a/app/policies/course_policy.rb
+++ b/app/policies/course_policy.rb
@@ -151,9 +151,9 @@ class CoursePolicy < ApplicationPolicy
   def permitted_attributes
     # record is the Course class on create
     if zeus?
-      %i[name year description visibility registration color teacher institution_id moderated enabled_questions featured]
+      %i[name year description visibility registration teacher institution_id moderated enabled_questions color featured]
     elsif course_admin? || (record == Course && user&.admin?)
-      %i[name year description visibility registration color teacher institution_id moderated enabled_questions]
+      %i[name year description visibility registration teacher institution_id moderated enabled_questions]
     else
       []
     end
--- a/app/policies/course_policy.rb
+++ b/app/policies/course_policy.rb
@@ -150,7 +150,9 @@ class CoursePolicy < ApplicationPolicy
 
   def permitted_attributes
     # record is the Course class on create
-    if course_admin? || (record == Course && user&.admin?)
+    if zeus?
+      %i[name year description visibility registration color teacher institution_id moderated enabled_questions featured]
+    elsif course_admin? || (record == Course && user&.admin?)
       %i[name year description visibility registration color teacher institution_id moderated enabled_questions]
     else
       []
--- a/app/views/courses/_form.html.erb
+++ b/app/views/courses/_form.html.erb
@@ -75,6 +75,16 @@
       <span class="help-block offset-sm-3 col-sm-6"><%= t ".description-help" %></span>
     </div>
 
+    <div class="field form-group row">
+      <%= f.label :questions, t(".questions.title"), class: 'col-sm-3 col-form-label' %>
+      <div class="col-sm-6 pt-2">
+        <div class="form-check <%= "form-switch" unless course.new_record? %>">
+          <%= f.check_box :enabled_questions, class: 'form-check-input' %>
+          <%= f.label :enabled_questions, t(".questions.toggle-label"), class: 'form-check-label' %>
+        </div>
+      </div>
+    </div>
+
     <% if current_user.zeus? %>
       <div class="field form-group row">
         <%= f.label :color, class: 'col-sm-3 col-form-label' %>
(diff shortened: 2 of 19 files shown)


app/policies/course_policy.rb
```diff
@@ -150,7 +150,9 @@ def export?
 
   def permitted_attributes
     # record is the Course class on create
-    if course_admin? || (record == Course && user&.admin?)
+    if zeus?
+      %i[name year description visibility registration color teacher institution_id moderated enabled_questions featured]
+    elsif course_admin? || (record == Course && user&.admin?)
       %i[name year description visibility registration color teacher institution_id moderated enabled_questions]
```
bmesuere, line 156: @chvp `color` was in the list of permitted attributes here, but I think it shouldn't be?

chvp: Indeed, it shouldn't be. This has been a bug since the color feature was introduced: https://github.com/dodona-edu/dodona/commit/ab5081547dbbcfee7ca5b17c82a2424b69268b4a
I can't seem to find the related PR though, to see if any reasoning was given.

bmesuere: Ok, I'll move it to the newly added zeus list.

chvp: PR is https://github.com/dodona-edu/dodona/pull/415

app/views/courses/_form.html.erb
```diff
@@ -86,17 +96,16 @@
...
       </div>
-    <% end %>
-
-    <div class="field form-group row">
-      <%= f.label :questions, t(".questions.title"), class: 'col-sm-3 col-form-label' %>
-      <div class="col-sm-6 pt-2">
-        <div class="form-check <%= "form-switch" unless course.new_record? %>">
-          <%= f.check_box :enabled_questions, class: 'form-check-input' %>
-          <%= f.label :enabled_questions, t(".questions.toggle-label"), class: 'form-check-label' %>
+      <div class="field form-group row">
```
niknetniko, line 99: What's the reasoning behind making this a checkbox, while the questions toggle is a switch? This won't be a real problem since only zeus can see this, but the difference isn't entirely clear to me at first glance:

image

niknetniko (approved):

app/views/courses/_form.html.erb
bmesuere: In determining if something should be a toggle or a switch, I ask myself "is this something that can be enabled/disabled" en would it make sense to reformulate the option as such.

My feeling is that the featured flag does not meet that criterium and that it is more like a property. ("enable featured course" doesn't sound right, "enable questions" does).

(The series options are still checkboxes and should probably be toggles, but for that they need to be reformulated in a positive way (show progress instead of hide progress) and this required controller code changes)

niknetniko: Makes a lot of sense, thanks for the explanation 👍

app/views/courses/_form.html.erb
```diff
@@ -75,6 +75,16 @@
...
+      <%= f.label :questions, t(".questions.title"), class: 'col-sm-3 col-form-label' %>
+      <div class="col-sm-6 pt-2">
+        <div class="form-check <%= "form-switch" unless course.new_record? %>">
+          <%= f.check_box :enabled_questions, class: 'form-check-input' %>
+          <%= f.label :enabled_questions, t(".questions.toggle-label"), class: 'form-check-label' %>
+        </div>
+      </div>
+    </div>
+
     <% if current_user.zeus? %>
```
chvp, line 88: I think it would be better to use our permission helper instead of a user level check. (Also for the featured checkbox, obviously.) See for example https://github.com/dodona-edu/dodona/blob/develop/app/views/users/_form.html.erb#L21

chvp (changes requested):

app/views/courses/_form.html.erb
bmesuere: :+1: I didn't know this existed, will change.

chvp (approved):

PR #2865, opened 2021-06-16
Debounce grade clicks

### LINKED ISSUE #2821: Debounce score increment/decrement to make sure input is not lost
From https://github.com/dodona-edu/dodona/issues/2695#issuecomment-841622499

niknetniko:
This pull request adds a debounce to saving the grade values, to fix #2821.

We'll probably need to play with the delay a bit, since it is a compromise between not waiting too long after typing a value, and waiting long enough that the debounce is useful when using the increment buttons.

Todo:

- [x] Fix the system tests; they don't seem to understand the debounce.
- [x] See if @chvp's suggestion to show the "saving" animation without delay (if the delay is so long that users think nothing happens) is needed

In the end I did opt to include the debounce from lodash:
- it's just the debounce, so it doesn't that large
- seemed unnecessary to write it ourselves
- it's already included by other dependencies, so there is no actually added code

This did require the `allowSyntheticDefaultImports` to be enabled, but I couldn't find any downsides. (and typescript in my ide kept complaining)

### DIFF
--- a/app/assets/javascripts/feedback/score.ts
+++ b/app/assets/javascripts/feedback/score.ts
@@ -1,4 +1,4 @@
-import { fetch } from "util.js";
+import { fetch, delay } from "util.js";
 import FeedbackActions from "feedback/actions";
 
 /**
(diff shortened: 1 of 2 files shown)


bmesuere:
I think we already use a similar function: https://github.com/dodona-edu/dodona/blob/develop/app/assets/javascripts/util.js#L9

niknetniko:
I've updated the code to use the existing utility function instead.

app/assets/javascripts/feedback/score.ts
```diff
@@ -59,18 +59,26 @@ export default class ScoreForm {
...
+                    return;
+                }
+                if (!this.input.reportValidity()) {
+                    return;
+                }
+                if (updating) {
+                    return;
+                }
+                updating = true;
+                console.log("execute");
```
chvp, line 79: Remove debugging.
```suggestion
```

chvp (approved):

bmesuere (approved):

PR #2871, opened 2021-06-19
Add courses link to signed out navbar

bmesuere:
This pull request adds a link to the courses listing to the signed out navbar. This page was only recently enabled for signed out users.


chvp (commented):
There's quite a lot of items in the navbar now. This isn't necessarily a problem, but this is how it would look on a Samsung S9:
Screen Shot 2021-06-19 at 19 58 48

niknetniko (approved):

bmesuere:
Yes, I noticed as well, that's why I put the most important first (+ to get them indexed by Google). The alternative was to hide them in an overflow menu but I don't think it's worth the effort.

chvp (approved):

PR #2883, opened 2021-06-24
Fix some dutch strings in the english translation files for grading

### LINKED ISSUE #2803: Translate some words in configure score item to English
image
Nee -> No
image
Stel een puntenverdeling op -> Create a scoring scheme

chvp:
Closes #2803.

### DIFF
--- a/config/locales/views/score_items/en.yml
+++ b/config/locales/views/score_items/en.yml
@@ -31,7 +31,7 @@ en:
       no_answer: "No"
       no_explanation: "Continue to the evaluation"
       second_step: "Create a scoring scheme"
-      second_explanation: You can create a grading scheme below for each exercise. You can always change these later.
+      second_explanation: You can create a scoring scheme below for each exercise. You can always change these later.
       summary_html:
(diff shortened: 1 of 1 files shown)


niknetniko (approved):

config/locales/views/score_items/en.yml
```diff
@@ -28,16 +28,16 @@ en:
       choose: Do you want to grade submissions in Dodona?
       yes_answer: "Yes"
       yes_explanation: "Create a scoring scheme below"
-      no_answer: "Nee"
+      no_answer: "No"
       no_explanation: "Continue to the evaluation"
-      second_step: "Stel een puntenverdeling op"
+      second_step: "Create a scoring scheme"
```
bmesuere, line 33: The line below mentions "grading scheme" instead of scoring scheme, line 30 uses scoring scheme as well. We probably should pick one and use it for all strings.

chvp: I opted for scoring scheme.

bmesuere (approved):

PR #2904, opened 2021-07-01
Add dark mode sponsor logos

### LINKED ISSUE #2645: Logos on about and sponsor page are ugly in dark mode
image

BTWS2:
This pull request adds alternative dark mode logos on the home page and support us page.

image

image

Closes #2645.


bmesuere (commented):
I'm not sure if handling this using css is the best way since we then rely on the browser to be smart enough to not load the image that's not used.

@chvp is darkmode always set using a server roundtrip or is this done in javascript? If it is the first, it might be better to insert the correct image using rails.

@BTWS2 Thank you for your contribution! Can you include a screenshot in the pull request of future contributions? This makes reviewing a lot quicker.

BTWS2:
@bmesuere indeed, I found the class in [the markdown documentation](https://docs.dodona.be/en/references/exercise-description/#dark-mode-visibility).

Yes, I will include screenshots in the future. I didn't have time yet to get Dodona up and running locally.

chvp (changes requested):
The markdown method has to do it that way because it can't serve dynamic content. We can. The dark mode is always set by the server, so inserting the correct image on the rails side definitely seems better to me.

chvp:
So, now that I remembered that the browser can set dark mode without the server knowing about it I think the way it was done here is probably the best way to do this. We should test that the images are only loaded when needed though.

BTWS2:
I tested it with a small example and [it works with `loading="lazy"`](https://selleo.com/til/posts/rrfqojakbh-dont-hide-images-with-display-none).

image

bmesuere (commented):
image

This doesn't work because dark-only and light-only only work in exercise descriptions. I will update the PR to make these classes available globally.

chvp (approved):
It might be nice to use the `light-only` and `dark-only` classes in the navbar for the switch there. There's some ugly javascript in `application.html.erb` for that right now. But that should probably be done in a separate pull request.

bmesuere:
@BTWS2 Thank you for your contribution! As you can see, I updated your initial post and the title of the pull request. I did this so it's easier to see what was changed when checking this PR at a later time. In addition, the title of the PR is used to automatically generate the release notes.

BTWS2:
You're welcome! Thank you for the assistance.

PR #2917, opened 2021-07-06
Fix spinning score items

### LINKED ISSUE #2891: Score items keep spinning after being cleared
First fill a score item and then clear it again: score items keeps spinning. You can use this to create some cool effects though:

https://user-images.githubusercontent.com/5736113/123450563-2af32a80-d5dd-11eb-8529-0c45738a2714.mp4

niknetniko:
This pull request changes two related things:

- Don't start the spinning animation when the input is invalid; this gives the impression the value is being saved, but it's not.
- Allow to clear the input field, and interpret it as clearing the grade. E.g. pressing backspace or delete until the field is empty is now equivalent to pressing the delete button.

- [x] Tests were added

Closes #2891

### DIFF
--- a/app/assets/javascripts/feedback/score.ts
+++ b/app/assets/javascripts/feedback/score.ts
@@ -130,7 +130,14 @@ export default class ScoreForm {
         }
     }
 
-    private sendUpdate(newFocus: HTMLElement | null = null): void {
+    private sendUpdate(newFocus?: HTMLElement): void {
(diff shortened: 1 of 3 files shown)


bmesuere (approved):
Looks good, but if we ever want to convert something to a lit element, this seems like a good candidate :)

chvp (approved):

PR #2927, opened 2021-07-15
Show feedback status when grading

### LINKED ISSUE #2745: Visual clue if evaluation is completed or not in grading mode
From #2695:

>  page where submissions are graded no longer contain a visual clue whether or not the evaluation has been completed for this submission

One possible solution would be re-adding the section where the complete button normally is, but without button:
Not completed             |  Completed
:-------------------------:|:-------------------------:
  |  

Possible upsides:
- Same place as when using non-graded version
- Directly above the button, so not completed is noticeable

Possible downsides:
- Perhaps confusing, as there are no buttons (you can't manually control the completed state from the UI when using grades), while there are buttons in the non-graded version
- Is a lot of vertical space, so perhaps a smaller indicator is better

If we add such an indicator, we can consider if this is still necessary:

>  no warning is given when next evaluation is visited with an incomplete grading; is this by design or should there at least come a warning? I missed to fill in one of the score items and wasn't aware of it until I continued evaluating from the starting page of the evaluation (ending up at an incomplete evaluation)

niknetniko:
This pull request shows the feedback status when grading. The table below shows the four different possible statusses:

  / |Not completed             |  Completed
:----:|:-------------------------:|:-------------------------:
Graded | image | image
Not graded | image | image

- The completed status when not grading has been centered for consistency.
- The uncompleted status when grading has a tooltip (as suggested in the issue).

Closes #2745.


bmesuere:
Sorry, I must have missed the creation of the issue for this. Is there a reason why we're adding this back other than "it isn't there"?

chvp:
> Sorry, I must have missed the creation of the issue for this. Is there a reason why we're adding this back other than "it isn't there"?

When correcting when grading, I did miss it. It's easier to see quickly if you're finished there than examining all score items tot see if they have been filled.

config/locales/views/feedbacks/nl.yml
```diff
@@ -26,16 +26,18 @@ nl:
...
       exercise_progress: "%{feedback_count} / %{feedback_total} evaluaties afgewerkt"
       total: "Totaal"
       give_max_all: "Geef de maximumscore aan alle scoreonderdelen"
       give_zero_all: "Geef 0 aan alle scoreonderdelen"
+    feedback_status:
+      evaluation_done: Evaluatie afgewerkt
+      evaluation_not_done_html: Evaluatie <strong>niet</strong> afgewerkt
+      mark_as_done: Evaluatie afwerken
+      undo: Ongedaan maken
+      done_when_graded: De evalutie wordt automatisch afgewerkt als elk scoreonderdeel punten gekregen heeft.
```
chvp, line 40:
```suggestion
      done_when_graded: De evaluatie wordt automatisch afgewerkt als elk scoreonderdeel punten gekregen heeft.
```

chvp (approved):

chvp:
After some discussion with @bmesuere this morning, I reworked this to make the total more like a form input and use the icon to the left for showing the status when grading. I used the `pencil` and `check` icons, since that's we use for the status icon when not grading.
screenshot_2021-08-03-142026
screenshot_2021-08-03-142034
screenshot_2021-08-03-142045
screenshot_2021-08-03-142053

pdawyndt:
Penil won’t give the impression that the total score can be edited by clicking the “button”?

chvp:
I replaced the pencil with circle-slice-3.

screenshot_2021-08-03-155022
screenshot_2021-08-03-155032

bmesuere:
This looks good 👍 Do you think we additionally need to highlight the fields with missing data/mute the complete ones?

chvp:
I don't think so. It's already visible through the disabled status of the clear button, and more visual focus would result in a lot of visual focus when there are a lot of score items (which is the case where not finding the missing data is most common).

bmesuere (approved):

PR #2951, opened 2021-08-03
Use alternative background color for line highlighting in dark mode

### LINKED ISSUE #2412: Use alternative background color for line highlighting in dark mode
The yellow background color that is used for line highlighting makes the actual code hard to read in dark mode:

image

This is not the case in light mode:

image

Example: [submission 6623077](https://dodona.ugent.be/en/submissions/6623077/)

TimonDB:
This pull request adapts the color for highlighting code lines in dark mode in case of a runtime error. The code line should now be more readable.

Before:
Screenshot from 2021-08-03 12-07-59

After:
Screenshot from 2021-08-03 14-41-54

Closes #2412 .

### DIFF
--- a/app/assets/stylesheets/theme/theme-dark.css.scss
+++ b/app/assets/stylesheets/theme/theme-dark.css.scss
@@ -23,7 +23,7 @@
   $danger-800: $red-200 !global;
   $danger-500: $red-300 !global;
   $info-500: $light-blue-300 !global;
-  $warning-500: $orange-400 !global;
+  $warning-500: $amber-500 !global;
 
   $body-bg: $blue-grey-800 !global;
   $content-bg: $blue-grey-900 !global;
@@ -53,6 +53,7 @@
 
   $code-bg: $body-bg !global;
   $code-light-bg: $body-bg !global;
+  $code-bg-warning: rgba(80, 52, 18, 0.4) !global;
 
   $alert-warning-bg: inherit !global;
   $alert-info-bg: inherit !global;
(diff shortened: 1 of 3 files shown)


chvp (commented):
@bmesuere probably has a better opinion on this, but it still doesn't seem very readable to me. Going (a lot) darker might be better.

bmesuere:
> @bmesuere probably has a better opinion on this, but it still doesn't seem very readable to me. Going (a lot) darker might be better.

Indeed, I would also create a separate variable for this since the warning color can be used in other places and this change can thus have other unintended side effects.

chvp (approved):

niknetniko (approved):

app/assets/stylesheets/theme/theme-dark.css.scss
```diff
@@ -23,7 +23,7 @@
   $danger-800: $red-200 !global;
   $danger-500: $red-300 !global;
   $info-500: $light-blue-300 !global;
-  $warning-500: $amber-300 !global;
```
chvp, line 26: Ah, actually, this probably shouldn't be changed. AFAIK this is used in a number of other places, where we don't necessarily want to change the color. If this messes up the left border you can create an extra variable for it.

TimonDB: Yes, I just saw it too. I will undo it

TimonDB: It has minimal effect on the left border when I change it back to `amber-300`, so creating an extra variable for the left border is not necessary I think 
Screenshot from 2021-08-03 14-54-00

TimonDB:
As a result, in light theme there is also a left border visible, can this stay this way?
Screenshot from 2021-08-03 14-58-50

chvp (approved):

bmesuere (approved):

PR #2952, opened 2021-08-03
Make sure progress is updated when entering the last score while grading

chvp:
This pull request makes sure the progress indicator at the bottom of the feedback actions is updated appropiately while grading.


niknetniko (approved):

bmesuere (approved):
was there an open issue for this?

PR #2954, opened 2021-08-03
Re-implement util functions using URL API

### LINKED ISSUE #2614: Use URL API in util.js
The functions `updateURLParameter`, `updateArrayURLParameter`, `getURLParameter` and `getArrayURLParameter` in `util.js` can probably be re-implemented using the [URL](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL) and [URL.searchParams](https://developer.mozilla.org/en-US/docs/Web/API/URL/searchParams) APIs. These are supported in all modern browsers: [URL](https://caniuse.com/url) and [URLSearchParams](https://caniuse.com/urlsearchparams).

TimonDB:
This pull request re-implements the helper URL functions by using the URL API. Also, a test file for these functions is added.

Closes #2614  .

### DIFF
--- a/app/assets/javascripts/exercise.js
+++ b/app/assets/javascripts/exercise.js
@@ -142,14 +142,10 @@ function initExerciseDescription() {
     centerImagesAndTables();
 }
 
-function initExerciseShow(exerciseId, programmingLanguage, loggedIn, editorShown, courseId, _deadline, _baseSubmissionUrl) {
+function initExerciseShow(exerciseId, programmingLanguage, loggedIn, editorShown, courseId, _deadline, baseSubmissionUrl) {
     let editor;
     let lastSubmission;
     let lastTimeout;
-
-    // remove the submission id to get the base submission url
-    const to = (_baseSubmissionUrl.substr(0, _baseSubmissionUrl.length - 1)).lastIndexOf("/");
-    const baseSubmissionUrl = _baseSubmissionUrl.substring(0, to + 1);
 
     function init() {
         if (editorShown) {
--- a/app/assets/javascripts/exercise.js
+++ b/app/assets/javascripts/exercise.js
@@ -142,10 +142,14 @@ function initExerciseDescription() {
     centerImagesAndTables();
 }
 
-function initExerciseShow(exerciseId, programmingLanguage, loggedIn, editorShown, courseId, _deadline) {
+function initExerciseShow(exerciseId, programmingLanguage, loggedIn, editorShown, courseId, _deadline, _baseSubmissionUrl) {
     let editor;
     let lastSubmission;
     let lastTimeout;
+
+    //remove the submission id to get the base submission url
+    const to = (_baseSubmissionUrl.substr(0, _baseSubmissionUrl.length - 1)).lastIndexOf("/");
+    let baseSubmissionUrl = _baseSubmissionUrl.substring(0, to + 1);
 
     function init() {
         if (editorShown) {
@@ -270,7 +274,7 @@ function initExerciseShow(exerciseId, programmingLanguage, loggedIn, editorShown
                 return;
             }
             event.preventDefault();
-            loadFeedback($(this).attr("href"), $(this).data("submission_id"));
+            loadFeedback(baseSubmissionUrl + $(this).data("submission_id"), $(this).data("submission_id"));
         });
     }
 
@@ -296,7 +300,7 @@ function initExerciseShow(exerciseId, programmingLanguage, loggedIn, editorShown
                     $submissionRow.find(".load-submission").get(0).click();
                 } else if ($("#activity-feedback-link").hasClass("active") &&
                     $("#activity-feedback-link").data("submission_id") === lastSubmission) {
-                    loadFeedback(`/submissions/${lastSubmission}`, lastSubmission);
+                    loadFeedback(baseSubmissionUrl + lastSubmission, lastSubmission);
                 }
                 showFABStatus(status);
                 setTimeout(enableSubmitButton, 100);
--- a/app/assets/javascripts/exercise.js
+++ b/app/assets/javascripts/exercise.js
@@ -142,7 +142,7 @@ function initExerciseDescription() {
     centerImagesAndTables();
 }
 
-function initExerciseShow(exerciseId, programmingLanguage, loggedIn, editorShown, courseId, _deadline, baseSubmissionUrl) {
+function initExerciseShow(exerciseId, programmingLanguage, loggedIn, editorShown, courseId, _deadline, baseSubmissionsUrl) {
     let editor;
     let lastSubmission;
     let lastTimeout;
@@ -270,7 +270,7 @@ function initExerciseShow(exerciseId, programmingLanguage, loggedIn, editorShown
                 return;
             }
             event.preventDefault();
-            loadFeedback(baseSubmissionUrl + $(this).data("submission_id"), $(this).data("submission_id"));
+            loadFeedback(baseSubmissionsUrl + $(this).data("submission_id"), $(this).data("submission_id"));
         });
     }
 
@@ -296,7 +296,7 @@ function initExerciseShow(exerciseId, programmingLanguage, loggedIn, editorShown
                     $submissionRow.find(".load-submission").get(0).click();
                 } else if ($("#activity-feedback-link").hasClass("active") &&
                     $("#activity-feedback-link").data("submission_id") === lastSubmission) {
-                    loadFeedback(baseSubmissionUrl + lastSubmission, lastSubmission);
+                    loadFeedback(baseSubmissionsUrl + lastSubmission, lastSubmission);
                 }
                 showFABStatus(status);
                 setTimeout(enableSubmitButton, 100);
--- a/app/assets/javascripts/exercise.js
+++ b/app/assets/javascripts/exercise.js
@@ -147,9 +147,9 @@ function initExerciseShow(exerciseId, programmingLanguage, loggedIn, editorShown
     let lastSubmission;
     let lastTimeout;
 
-    //remove the submission id to get the base submission url
+    // remove the submission id to get the base submission url
     const to = (_baseSubmissionUrl.substr(0, _baseSubmissionUrl.length - 1)).lastIndexOf("/");
-    let baseSubmissionUrl = _baseSubmissionUrl.substring(0, to + 1);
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/util.js
```diff
@@ -100,24 +100,16 @@ function updateArrayURLParameter(url, param, _paramVals) {
 }
 
 function getURLParameter(_name, _url) {
```
bmesuere, line 102: Since `name` is no longer overwritten, we can use `name` here instead of `_name`.

app/assets/javascripts/util.js
```diff
@@ -100,24 +100,16 @@ function updateArrayURLParameter(url, param, _paramVals) {
...
-    const url = _url || window.location.href;
-    const name = _name.replace(/[[\]]/g, "\\$&");
-    const regex = new RegExp("[?&]" + name + "(=([^&#]*)|&|#|$)");
-    const results = regex.exec(url);
-    if (!results) return null;
-    if (!results[2]) return "";
-    return decodeURIComponent(results[2].replace(/\+/g, " "));
+    const url = new URL(_url || window.location.href);
```
bmesuere, line 103:
```suggestion
    const url = new URL(_url ?? window.location.href);
```

@chvp I think this is never called with an empty string so we can used nullish coalescing here instead of the logical or?

If so, also change in the function below.

app/assets/javascripts/util.js
```diff
@@ -100,24 +100,16 @@ function updateArrayURLParameter(url, param, _paramVals) {
...
+    const result = url.searchParams.get(_name);
```
bmesuere, line 104: According to the [docs](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/get), `get` returns `null` if the parameter isn't found. You can thus return the value directly without the checks in the lines below.

app/assets/javascripts/util.js
```diff
@@ -100,24 +100,16 @@ function updateArrayURLParameter(url, param, _paramVals) {
...
-    const url = _url || window.location.href;
-    const result = [];
-    for (const part of url.split(/[?&]/)) {
-        const regResults = new RegExp(`${name}%5B%5D=([^#]+)`).exec(part);
-        if (regResults && regResults[1]) {
-            result.push(decodeURIComponent(regResults[1]));
-        }
-    }
+    const url = new URL(_url || window.location.href);
+    const result = url.searchParams.getAll(name);
```
bmesuere, line 111: Same as with the other function. getAll will return an empty array when the parameter is not present.

bmesuere (changes requested):

app/assets/javascripts/util.js
TimonDB: ah so there is only an underscore before the parameter in case it is overwritten in the function? I was already wondering what the logic behind the underscore was

app/assets/javascripts/util.js
chvp: Sound reasonable, if it was called with the empty string it was probably a bug that the `href` was used in that case.

app/assets/javascripts/util.js
chvp: If we redefine the variable as `const` we can't use the same variable name (and mutating arguments is generally seen as bad style).

chvp (changes requested):
The issue also mentios the `updateURL...` functions. Is there a reason these weren't changed.

TimonDB:
> The issue also mentios the `updateURL...` functions. Is there a reason these weren't changed?

@chvp The `updateURL...` functions sometimes have only a pathname as URL argument (e.g. `nl/submissions/53` was one I encountered), this resulted in an error (`Error: Invalid URL`) when trying to create the URL `const url = new URL(_url)`

niknetniko:
> The `updateURL...` functions sometimes have only a pathname as URL argument (e.g. `nl/submissions/53` was one I encountered), this resulted in an error (`Error: Invalid URL`) when trying to create the URL `const url = new URL(_url)`

Is this used this way in a lot of places? Because if not, it might be worth to update the caller to provide the full URL, allowing us to also update the `updateURL...` functions, which would be nice.

app/assets/javascripts/util.js
```diff
@@ -99,25 +99,15 @@ function updateArrayURLParameter(url, param, _paramVals) {
...
-    const name = _name.replace(/[[\]]/g, "\\$&");
-    const regex = new RegExp("[?&]" + name + "(=([^&#]*)|&|#|$)");
-    const results = regex.exec(url);
-    if (!results) return null;
-    if (!results[2]) return "";
-    return decodeURIComponent(results[2].replace(/\+/g, " "));
+function getURLParameter(name, _url) {
+    const url = new URL(_url ?? window.location.href);
+    const result = url.searchParams.get(name);
+    return result;
```
bmesuere, line 105:
```suggestion
    return url.searchParams.get(name);
```

app/assets/javascripts/util.js
```diff
@@ -99,25 +99,15 @@ function updateArrayURLParameter(url, param, _paramVals) {
...
-    const result = [];
-    for (const part of url.split(/[?&]/)) {
-        const regResults = new RegExp(`${name}%5B%5D=([^#]+)`).exec(part);
-        if (regResults && regResults[1]) {
-            result.push(decodeURIComponent(regResults[1]));
-        }
-    }
+    const url = new URL(_url ?? window.location.href);
+    const result = url.searchParams.getAll(name);
     return result;
```
bmesuere, line 111:
```suggestion
    return url.searchParams.getAll(name);
```

bmesuere (changes requested):
I also agree with @niknetniko's comment

chvp:
The easiest way to fix that there would probably be to add a `base_submissions_url` data attribute in the HTML. This can then be read in the setup function and used in place of the `/submissions/` strings in the javascript file.

bmesuere (changes requested):
It would be reassuring if some jest tests were added covering those new functions. In addition, it would make sense to "port" these functions to typescript.

@chvp do we already have a file where we can place such helper functions in ts?

chvp:
> @chvp do we already have a file where we can place such helper functions in ts?

We don't, no.

app/assets/javascripts/exercise.js
```diff
@@ -142,11 +142,15 @@ function initExerciseDescription() {
...
 }
 
-function initExerciseShow(exerciseId, programmingLanguage, loggedIn, editorShown, courseId, _deadline) {
+function initExerciseShow(exerciseId, programmingLanguage, loggedIn, editorShown, courseId, _deadline, _baseSubmissionUrl) {
     let editor;
     let lastSubmission;
     let lastTimeout;
 
+    // remove the submission id to get the base submission url
+    const to = (_baseSubmissionUrl.substr(0, _baseSubmissionUrl.length - 1)).lastIndexOf("/");
```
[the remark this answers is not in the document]
TimonDB: @bmesuere How do I get the general submission/index path? I don't see where I can find it in the codebase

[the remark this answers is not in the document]
niknetniko: In this case it will probably be `submissions_url`. You can find all routes by issuing the command `rails routes` (see the excellent rails guide to see all possible ways to view all existing routes: https://guides.rubyonrails.org/routing.html#listing-existing-routes)

test/javascript/util.test.js
```diff
@@ -0,0 +1,63 @@
...
+let noParameterURL;
+let oneParameterURL;
+let twoParameterURL;
+let multipleValueUrl;
+
+beforeEach(() => {
+    noParameterURL = "https://www.dodona.ugent.be/test_functions";
+    oneParameterURL = "https://www.dodona.ugent.be/test_functions?param1=paramVal1";
+    twoParameterURL = "https://www.dodona.ugent.be/test_functions?param1=paramVal1&param2=paramVal2";
+    multipleValueUrl = "https://www.dodona.ugent.be/test_functions?param=paramVal1&param=paramVal2&param=paramVal3";
```
niknetniko, line 12: In tests and examples, the convention is to use one of the reserved domains for testing, such as `example.com`. That way the tests can never accidentally use a real service or domain name.

niknetniko (changes requested):

test/javascript/util.test.js
TimonDB: I didn't know that, but sounds indeed reasonable. I will adapt it, thanks for the feedback!

app/assets/javascripts/exercise.js
```diff
@@ -142,7 +142,7 @@ function initExerciseDescription() {
     centerImagesAndTables();
 }
 
-function initExerciseShow(exerciseId, programmingLanguage, loggedIn, editorShown, courseId, _deadline) {
+function initExerciseShow(exerciseId, programmingLanguage, loggedIn, editorShown, courseId, _deadline, baseSubmissionUrl) {
```
chvp, line 145:
```suggestion
function initExerciseShow(exerciseId, programmingLanguage, loggedIn, editorShown, courseId, _deadline, baseSubmissionsUrl) {
```

app/assets/javascripts/exercise.js
```diff
@@ -270,7 +270,7 @@ function initExerciseShow(exerciseId, programmingLanguage, loggedIn, editorShown
                 return;
             }
             event.preventDefault();
-            loadFeedback($(this).attr("href"), $(this).data("submission_id"));
+            loadFeedback(baseSubmissionUrl + $(this).data("submission_id"), $(this).data("submission_id"));
```
chvp, line 273:
```suggestion
            loadFeedback(baseSubmissionsUrl + $(this).data("submission_id"), $(this).data("submission_id"));
```

app/assets/javascripts/exercise.js
```diff
@@ -296,7 +296,7 @@ function initExerciseShow(exerciseId, programmingLanguage, loggedIn, editorShown
                     $submissionRow.find(".load-submission").get(0).click();
                 } else if ($("#activity-feedback-link").hasClass("active") &&
                     $("#activity-feedback-link").data("submission_id") === lastSubmission) {
-                    loadFeedback(`/submissions/${lastSubmission}`, lastSubmission);
+                    loadFeedback(baseSubmissionUrl + lastSubmission, lastSubmission);
```
chvp, line 299:
```suggestion
                    loadFeedback(baseSubmissionsUrl + lastSubmission, lastSubmission);
```

chvp (approved):

niknetniko (approved):

PR #2960, opened 2021-08-04
Fix scrolling of iFrames on mobile

### LINKED ISSUE #2869: Scrolling exercise descriptions on mobile browsers doesn't work well
The iframe seems to absorb the scroll events, so the page itself is hard to scroll.

niknetniko:
The culprit seems to be the `heightCalculationMethod: 'bodyScroll'`.
When using the default calculation method (`bodyOffset`), the scrolling
works fine. A bit of history (from https://github.com/dodona-edu/dodona/pull/1234):
- The `bodyScroll` value has been used since the creation of this helper in 04cad8f35d19f7a82eaf3f3bbacd5accb14de23c (no reason given).
- Before that, the value was `max`, added in 5ac1b0be098e401f33a58e8f3f9d013106f3b29f.

That last commit says "Fix iframe styling and resizing", but in my tests, using the default value doesn't break anything. The docs say:

> In cases where CSS styles causes the content to flow outside the body you may need to change this setting to one of the following options.

So it seems like the default value should work for our use.

Closes #2869 .

- [x] Needs to be verified, since I couldn't be bothered to link my phone to my pc.

### DIFF
--- a/app/helpers/activity_helper.rb
+++ b/app/helpers/activity_helper.rb
@@ -69,7 +69,6 @@ module ActivityHelper
                                    dark: dark).html_safe
     resizeframe = %{
       window.iFrameResize({
-          heightCalculationMethod: 'bodyScroll',
           onResized: dodona.afterResize,
           onMessage: dodona.onFrameMessage,
           onScroll: dodona.onFrameScroll,


chvp:
One thing I still want to check is if this still works when the iframe content changes size (e.g. by opening a collapse).

niknetniko:
> One thing I still want to check is if this still works when the iframe content changes size (e.g. by opening a collapse).

I tested it with an exercise to which I added a collapse on mestra and it seems to work (although mestra has been redeployed, so I have no link).

PR #2961, opened 2021-08-04
Align rows in scoresheet

### LINKED ISSUE #2772: Rows in scoresheet don't align
image

TimonDB:
This pull request fixes the misalignment in the scoresheets. By removing the height component of `.user-name`, the Grade overview table seems to be fixed. This solution however broke the Evaluation detail table. The new problem seems to be caused by some properties in the `status-table-wrapper`, which I think were used to keep the username column at a fixed width of `200px`. I now tried to avoid the new problem by using another way to keep the username column width fixed.

Before:
Screenshot from 2021-08-04 16-03-28

After:
Screenshot from 2021-08-04 16-05-09

Closes #2772.

### DIFF
--- a/app/assets/stylesheets/components/table.css.scss
+++ b/app/assets/stylesheets/components/table.css.scss
@@ -8,8 +8,13 @@
   background-attachment: local, local, scroll, scroll;
 }
 
+.user-name {
+  height: 42px;
+}
+
 .status-table-wrapper {
   overflow-x: auto;
+  margin-left: 200px;
 
   .table {
     border-collapse: separate;
@@ -27,12 +32,17 @@
       overflow: hidden;
     }
     .user-name {
+      position: absolute;
+      height: 35px;
       width: 200px;
-      min-width: 200px;
+      left: 18px;
       top: auto;
     }
+    .user-name:not(.sticky-column) {
+      height: 42px;
+    }
     thead .user-name {
-      height: auto;
+      height: auto !important;
     }
   }
 }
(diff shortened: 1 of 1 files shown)


chvp (changes requested):
So yeah, as mentioned on Slack, grading is now broken:
screenshot_2021-08-04-162420

niknetniko (changes requested):
This breaks the sticky first column (the name) in the evaluation details.
Instead of removing the absolute position of the `.user-name`, it might be fixable by overriding the height of the `user-name` column for either the evaluation details or the grade overview (since both need a different height).

TimonDB:
> This breaks the sticky first column (the name) in the evaluation details.
> Instead of removing the absolute position of the `.user-name`, it might be fixable by overriding the height of the `user-name` column for either the evaluation details or the grade overview (since both need a different height).

I will look into it. Out of curiosity, what is the purpose of a `sticky-column`?

app/assets/stylesheets/components/table.css.scss
```diff
@@ -108,3 +111,7 @@ tr.gu-mirror {
     color: $text-secondary;
   }
 }
+
+#scoresheet-table-wrapper .user-name{
```
chvp, line 115:
```suggestion
#scoresheet-table-wrapper .user-name {
```

chvp (approved):

niknetniko (approved):

bmesuere (approved):

PR #2964, opened 2021-08-05
Don't allow browsers to store cached versions of evaluation feedback pages

### LINKED ISSUE #2813: Going back doesn't show scores in evaluation
Clicking the ⬅ button in the browser (Chrome), goes to the previous student, but doesn't show the entered scores. If you try to enter the scores again you get the warning `This score was already changed by another user` and the previously entered scores load.

Is it possible to load the scores immediately after clicking ⬅?

chvp:
https://bugs.chromium.org/p/chromium/issues/detail?id=516846 mentions that a `Vary` header should fix this as well, but I don't think we have a request header that we can piggy-back off of to make sure the browser revalidates. See the comment in the change for trade-offs and rationale.

Closes #2813.

Remember to add `?pp=disable` when testing, otherwise our profiler overwrites the `Cache-Control` header anyway.

### DIFF
--- a/app/controllers/feedbacks_controller.rb
+++ b/app/controllers/feedbacks_controller.rb
@@ -29,6 +29,17 @@ class FeedbacksController < ApplicationController
     # the score_item ID in the `warning` param, it will be rendered with
     # the bootstrap warning classes.
     @warning = params[:warning]
+    # Don't allow browsers to store this page. This prevents the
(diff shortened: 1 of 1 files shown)


bmesuere (approved):
This seems to work for me 👍

niknetniko (approved):

PR #2967, opened 2021-08-05
Set current user's name as default value when creating a course

### LINKED ISSUE #2660: Set better default teacher name when creating a course
By default this field is left blank or the name of the copied course is taken. It would be better to set the name of the current user as a default value.

TimonDB:
This pull request uses the current user's name as default value for the teacher when creating a new course. When copying an existing course, the current user's name and institution are used as default values.

Before:
Screenshot from 2021-08-05 17-15-18

After:
Screenshot from 2021-08-05 17-15-41

Closes #2660.

### DIFF
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -62,10 +62,10 @@ class CoursesController < ApplicationController
       @course = Course.new(
         name: @copy_options[:base].name,
         description: @copy_options[:base].description,
-        institution: @copy_options[:base].institution,
+        institution: current_user.institution,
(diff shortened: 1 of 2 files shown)


chvp (commented):
I think the original issue also asked for the default to be changed when copying a course. Most copying is either done by the same author, where there would be no difference, or from a template course, where we definitely want to overwrite the teacher.

chvp (approved):

niknetniko (approved):

bmesuere (approved):

PR #2976, opened 2021-08-10
Add feature `Add score item to all exercises` when editing score items

### LINKED ISSUE #2802: Add `Add score item to all exercises` feature when editing score items later
In the wizard you have the option to add a score item to all exercises. If you want to do that later, you don't have this option anymore. Was this a design choice?

image
image

TimonDB:
This pull request adds the option to add a score item to all exercises when editing the score items. Before, this option was only possible when creating a new evaluation.

Before:
Screenshot from 2021-08-10 11-04-07

After:
Screenshot from 2021-08-10 11-03-37

Closes #2802.

### DIFF
--- a/app/assets/javascripts/pythia_submission.js
+++ b/app/assets/javascripts/pythia_submission.js
@@ -130,6 +130,7 @@ function initPythiaSubmissionShow(submissionCode, activityPath) {
             $tutor.removeClass("fullscreen");
             $("#tutorviz").height($("#tutorviz").data("standardheight"));
         } else {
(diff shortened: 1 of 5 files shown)


chvp (approved):
I'm not a huge fan of the extra hidden field, but I don't see any better option.

niknetniko:
> I'm not a huge fan of the extra hidden field, but I don't see any better option.

I think we could use `redirect_back(fallback_location: ...)`, as we already do in some places.

chvp (approved):

niknetniko (approved):

bmesuere (approved):

PR #2979, opened 2021-08-11
Add iCal link for syncing deadlines with external calendar

### LINKED ISSUE #2596: iCal link for syncing deadlines with external calendar

TimonDB:
This pull request adds an iCal link for syncing the deadlines of a course with an external calendar.

Closes #2596.


Gemfile
```diff
@@ -181,3 +181,5 @@ gem 'tzinfo-data', platforms: %i[mingw mswin x64_mingw jruby]
 
 # interfacing with docker
 gem 'docker-api', '~> 2.2.0'
+
+gem "icalendar", "~> 2.7"
```
bmesuere, line 185: This seems like a good gem with plenty of users 👍 
We have the habit of placing a little comment in the gemfile mentioning where the specific gem is used.

app/controllers/courses_controller.rb
```diff
@@ -377,6 +377,28 @@ def reorder_series
     end
   end
 
+  def ical
+    @series = Series.where(course_id: @course.id)
```
bmesuere, line 381: We only want public series to show up so this needs to be filtered.

app/controllers/courses_controller.rb
```diff
@@ -377,6 +377,28 @@ def reorder_series
...
+
+    respond_to do |format|
+      format.html
+      format.ics do
+        cal = Icalendar::Calendar.new
+        cal.x_wr_calname = 'Course deadlines'
```
bmesuere, line 387: Use the actual course name.

app/controllers/courses_controller.rb
```diff
@@ -377,6 +377,28 @@ def reorder_series
     end
   end
 
+  def ical
+    @series = Series.where(course_id: @course.id)
+
+    respond_to do |format|
```
bmesuere, line 383: @chvp do we need a format block here if we always want to respond with the same response?

app/controllers/courses_controller.rb
```diff
@@ -377,6 +377,28 @@ def reorder_series
...
+      format.html
+      format.ics do
+        cal = Icalendar::Calendar.new
+        cal.x_wr_calname = 'Course deadlines'
+        series.each do |serie|
+          cal.event do |e|
+            e.dtstart = serie.deadline
+            e.dtend = serie.deadline
+            e.summary = serie.name
+            e.description = serie.description
```
bmesuere, line 393: I would not set the description (since this might expose semi-hidden data). Instead, the url field can be used to link to the specific series.

app/controllers/courses_controller.rb
```diff
@@ -377,6 +377,28 @@ def reorder_series
...
+    @series = Series.where(course_id: @course.id)
+
+    respond_to do |format|
+      format.html
+      format.ics do
+        cal = Icalendar::Calendar.new
+        cal.x_wr_calname = 'Course deadlines'
+        series.each do |serie|
+          cal.event do |e|
+            e.dtstart = serie.deadline
```
bmesuere, line 390: Are the timezones ok if you set the date like this? The icalendar docs use Icalendar::Values::DateTime.

app/controllers/courses_controller.rb
```diff
@@ -377,6 +377,28 @@ def reorder_series
...
+            e.dtend = serie.deadline
+            e.summary = serie.name
+            e.description = serie.description
+          end
+        end
+        cal.publish
+        render plain: cal.to_ical
```
bmesuere, line 397: You will need to look into the ical spec, but it might be needed to use `send_data` instead in combination with a specific type, the attachment flag and a filename ending in ics.

app/policies/course_policy.rb
```diff
@@ -142,6 +142,10 @@ def questions?
     course_admin?
   end
 
+  def ical?
+    course_admin?
```
bmesuere, line 146: The ical link will be used unauthenticated, so should always work.

app/views/courses/ical.html.erb
```diff
@@ -0,0 +1,26 @@
+<%= render partial: 'navbar_links' %>
```
bmesuere, line 1: A view is not needed because you specifically call render (or send_data).

bmesuere (changes requested):

app/controllers/courses_controller.rb
TimonDB: @bmesuere is a serie public when its `activities_visible` property is set to True?

app/policies/course_policy.rb
TimonDB: Does this mean that the function `def ical?`  is unnecessary?

bmesuere: no, for now simply put `true` here. We can later discuss if we want to allow ical links for completely hidden courses.

app/controllers/courses_controller.rb
chvp: Should not be necessary, no.

app/controllers/courses_controller.rb
```diff
@@ -377,6 +377,23 @@ def reorder_series
     end
   end
 
+  def ical
+    @series = Series.where(course_id: @course.id, visibility: :open)
```
chvp, line 381: There is no need for the `@` in `@series`, since there is no view.

app/controllers/courses_controller.rb
```diff
@@ -377,6 +377,23 @@ def reorder_series
...
+
+    cal = Icalendar::Calendar.new
+    cal.x_wr_calname = @course.name
+    @series.each do |serie|
+      cal.event do |e|
+        e.dtstart = serie.deadline
+        e.dtend = serie.deadline
+        e.summary = serie.name
+        e.description = series_url(serie)
```
chvp, line 390: There is a `url` field in the ICS spec, so it might be better to use that one. A series also has a description, but I'm not sure we want to add that here (since it can contain HTML/markdown).

app/views/courses/show.html.erb
```diff
@@ -91,6 +91,11 @@
                   <% end %>
                 </li>
               <% end %>
+              <li>
+                <%= link_to ical_course_path(@course), class: "dropdown-item" do %>
+                  <i class="mdi mdi-calendar-clock mdi-18"></i>Export course deadlines
```
TimonDB, line 96: This has to be cleaned up (proper translation), but to demonstrate, I just hardcoded the text

app/controllers/courses_controller.rb
```diff
@@ -377,6 +377,23 @@ def reorder_series
...
 
+  def ical
+    @series = Series.where(course_id: @course.id, visibility: :open)
+
+    cal = Icalendar::Calendar.new
+    cal.x_wr_calname = @course.name
+    @series.each do |serie|
+      cal.event do |e|
+        e.dtstart = serie.deadline
+        e.dtend = serie.deadline
```
chvp, line 388: [The spec](https://datatracker.ietf.org/doc/html/rfc5545#section-3.8.2.2) says the following: 
```
      The value type
      of this property MUST be the same as the "DTSTART" property, and
      its value MUST be later in time than the value of the "DTSTART"
      property.  Furthermore, this property MUST be specified as a date
      with local time if and only if the "DTSTART" property is also
      specified as a date with local time.
```
So this will produce an invalid file. Substracting a second or minute from the start seems like the best option to avoid this, I think.

chvp: The spec also mentions a [duration](https://datatracker.ietf.org/doc/html/rfc5545#section-3.8.2.5) which can be used in place of the `end`. Durations must be positive, but I can't immediately find whether they consider 0 a positive number or not.

app/controllers/courses_controller.rb
TimonDB: ah yes, I will adapt it. Btw, what is the best option to use in this case: `Series.where(...)` or `@course.series.where(...)`?

chvp: I have a slight preference for the latter, but it's not a strong one. It reads more natural to me.

app/controllers/courses_controller.rb
TimonDB: Since most calendars only show the hour and minutes of the event, would it not be better to add a second the end? This way, the start time and end time will both be truncated to the correct time

TimonDB: For example: it seems like my calendar app users `floor()` to display the start time, resulting in 11:59 instead of 12:00
Screenshot from 2021-08-12 12-20-32

chvp: Yeah, makes sense.

app/controllers/courses_controller.rb
TimonDB: Maybe we can just add a short message like `"Deadline for #{serie.name} of course #{@course.name}"` instead of the serie's description?

chvp: Sounds good to me.

app/controllers/courses_controller.rb
TimonDB: I have set the start and end time of the event in their respective UTC time, which is indicated in the .ics file by a `'Z'` at the end of `dtstart` and `dtend`. This way, all calendar apps should in theory show these start and end times correctly according to the user's time zone

config/locales/views/courses/en.yml
```diff
@@ -122,6 +122,7 @@ en:
         other: "%{count} pending users are declined."
       copy: "Copy this course"
       manage_series: "Manage series"
+      download_deadlines: "Export course deadlines"
```
chvp, line 125: This somehow needs to mention a calendar, I think. Right now I don't think users will really know what this means.

chvp (changes requested):
This looks good :+1: 
This still needs some tests, and we need to figure out what we do for closed courses. There's no real secret information in the exported ICS files, but I can imagine that some teachers might be hesitant in allowing this without any form of authentication.

config/locales/views/courses/en.yml
TimonDB: Fair point. Maybe "Export course deadlines for external calendar" is more clarifying?

bmesuere:
I think this needs another UI solution. If you click the link right now, a file is downloaded which can be imported in a calendar application. However, this is not what we want because this will cause a 1-time import of the deadlines. We want to expose actual link to the user so he can add the feed instead. (Sorry I previously pushed for the download stuff to be added in the controller as this doesn't seem to be the way to go after all)

TimonDB:
> I think this needs another UI solution. If you click the link right now, a file is downloaded which can be imported in a calendar application. However, this is not what we want because this will cause a 1-time import of the deadlines. We want to expose actual link to the user so he can add the feed instead. (Sorry I previously pushed for the download stuff to be added in the controller as this doesn't seem to be the way to go after all)

So if I understand correctly, we want to create a [webcal link](https://toasterlovin.com/rails-webcal-links-with-icalendar-gem/) that a user can click to subscribe to the calendar of a specific course? And the user's calendar app will then periodically fetch new events?

chvp:
A webcal link seems like a good mechanism :+1:

TimonDB:
When adding the course calendar to my Google Calendar using the URL (e.g. http://dodona.localhost:3000/nl/courses/3/ical), the deadlines don't seem to be imported. The [website](https://toasterlovin.com/rails-webcal-links-with-icalendar-gem/) I used as a guideline states the following:

> Your controller needs to be unauthenticated, or else the user's calendar app won't be able to access the calendar. So, if your `ApplicationController` is authenticated, your `CalendarsController` will probably need to inherit directly from `ActionController::Base` to avoid authentication.

Could this in our case also be the reason why the deadlines don't show up on Google Calendar?

bmesuere:
The policy is set to unauthenticated access. You can easily test this by loading the url in a separate browser or in private mode. Just to be sure, did you try to add the localhost url to google calendar? Because that will not work since it's only a local web server...

TimonDB:
> The policy is set to unauthenticated access. You can easily test this by loading the url in a separate browser or in private mode. Just to be sure, did you try to add the localhost url to google calendar? Because that will not work since it's only a local web server...

Ah yes, I indeed tried to add the localhost url, not my smartest move. That explains why it doesn't work

config/locales/views/courses/en.yml
chvp: "Add course to calendar", maybe?

chvp (commented):
Looks good, but translation strings still need to be updated.

bmesuere:
- Is it possible to add an extra (bootstrap) tooltip to this menu item that gives a little more information? Something like: Add this link to your calendar application to show the course deadlines in your calendar
- For now, I would limit the menu item to only be visible to course admins because we will probably need to check if the updates work

chvp (approved):

bmesuere:
It is now 24+ hours later and the deadline was still not updated in my google calendar.

bmesuere:
It was now updated so this seems to work.

bmesuere (approved):

PR #2993, opened 2021-08-17
Runtime error link switches to correct tab

### LINKED ISSUE #2992: Switch to Code tab when clicking Runtime error link
The [Runtime error](https://dodona.ugent.be/en/submissions?status=runtime%20error&page=1) link in the feedback of an exercise points to https://dodona.ugent.be/en/submissions/???/#, instead of https://dodona.ugent.be/en/submissions/???/#tab-code-?? (e.g. https://dodona.ugent.be/en/submissions/8013739/).
image
Because of this, clicking the Runtime error link does nothing.

TimonDB:
This pull request fixes the Runtime error tab link. Explanation of the fix: The nav-tab with `href="#code"` was searched. However, the two tabs had `href="#tab-corectheid-0"` and `href="#tab-code-1"`, so the search delivered an empty result. 

Closes #2992.

### DIFF
--- a/app/assets/javascripts/submission.js
+++ b/app/assets/javascripts/submission.js
@@ -49,7 +49,7 @@ function initSubmissionShow(parentClass, mediaPath, token) {
 
             $(".tab-link-marker").removeClass("tab-link-marker");
             $(".feedback-table .nav-tabs > li a").filter(function () {
-                return $(this).attr("href").indexOf(tab) > 0;
+                return $(this).attr("href").indexOf(`#tab-${tab}`) === 0;
             }).tab("show");
             if (element !== undefined) {
                 $("#element").addClass("tab-link-marker");
--- a/app/assets/javascripts/submission.js
+++ b/app/assets/javascripts/submission.js
@@ -49,7 +49,7 @@ function initSubmissionShow(parentClass, mediaPath, token) {
 
             $(".tab-link-marker").removeClass("tab-link-marker");
             $(".feedback-table .nav-tabs > li a").filter(function () {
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/submission.js
```diff
@@ -49,7 +49,7 @@ function initSubmissionShow(parentClass, mediaPath, token) {
 
             $(".tab-link-marker").removeClass("tab-link-marker");
             $(".feedback-table .nav-tabs > li a").filter(function () {
-                return $(this).attr("href").indexOf("#" + tab) === 0;
+                return $(this).attr("href").indexOf(tab) > 0;
```
chvp, line 52:
```suggestion
                return $(this).attr("href").indexOf(`#tab-${tab}`) === 0;
```
I think this makes more sense, since the word `code` will probably show up in other tab id's, which is the most common use of this code.

chvp (approved):

app/assets/javascripts/submission.js
```diff
@@ -49,7 +49,7 @@ function initSubmissionShow(parentClass, mediaPath, token) {
 
             $(".tab-link-marker").removeClass("tab-link-marker");
             $(".feedback-table .nav-tabs > li a").filter(function () {
-                return $(this).attr("href").indexOf("#" + tab) === 0;
+                return $(this).attr("href").indexOf(`#tab-${tab}`) === 0;
```
bmesuere, line 52:
```suggestion
                return $(this).attr("href").startsWith(`#tab-${tab}`);
```

bmesuere (changes requested):

bmesuere (approved):

PR #2995, opened 2021-08-17
Fix error when trying to create an URL if only path is given 

TimonDB:
This pull request provides a baseURL in the URL creation. This is needed because it is possible only a path is provided instead of a full URL. `window.location.href` is used as baseURL, because `window.location.host` did not provide a valid baseURL ( `dodona.localhost:3000` while `https://dodona.localhost:3000` was needed)

### DIFF
--- a/app/assets/javascripts/util.js
+++ b/app/assets/javascripts/util.js
@@ -15,7 +15,7 @@ const delay = (function () {
 })();
 
 function updateURLParameter(_url, param, paramVal) {
-    const url = new URL(_url);
+    const url = new URL(_url, window.location.href);
     if (paramVal) {
         url.searchParams.set(param, paramVal);
     } else {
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/util.js
```diff
@@ -15,7 +15,7 @@ const delay = (function () {
 })();
 
 function updateURLParameter(_url, param, paramVal) {
-    const url = new URL(_url);
+    const url = new URL(_url, window.location.href);
```
niknetniko, line 18: I think we probably want to use [`window.location.origin`](https://developer.mozilla.org/en-US/docs/Web/API/Location/origin), since `window.location.href` is the full URL, which will give problems if you pass a relative URL to the constructor (same for the other usages).

niknetniko (changes requested):

niknetniko (changes requested):
Code looks good to me 👍 , but I think it's a good idea to add one or more test cases where a partial URL is used.

TimonDB:
> Code looks good to me  , but I think it's a good idea to add one or more test cases where a partial URL is used.

Ah yes, fair point. I will add them

chvp (approved):
Agree with adding some tests, looks fine otherwise.

niknetniko (approved):

bmesuere (approved):

PR #2996, opened 2021-08-17
Use GLightbox instead of Strip as lightbox

### LINKED ISSUE #2720: Migrate away from Strip
Our current lightbox implementation is bloated and depends on jquery.
Investigate if we can migrate to a lighter (and more modern) implementation, for example https://dbrekalo.github.io/simpleLightbox/

### LINKED ISSUE #3000: Keep position when opening Lightbox
When an image is opened in a Lightbox, the description scrolls to the top of the page. Better UX if the description stays where it is.

Example: https://dodona.ugent.be/activities/1883509077/

TimonDB:
This pull request migrates away from Strip and uses [GLightbox](https://biati-digital.github.io/glightbox/#) instead.

Closes #2720 and closes #3000.


niknetniko (commented):
I haven't run it locally yet, but:

- I think we should unvendor this and include it as a yarn dependency instead of adding the files to our repo. The [docs](https://dbrekalo.github.io/simpleLightbox/) seem to indicate this is possible (and if not possible, we perhaps need to look at a different lightbox implementation).
- We'll need to localize the next/previous buttons (see `closeBtnCaption`, etc.)
- Check if MathJax works in the captions out of the box. If not, we probably need to hook it in, like we did with Strip. The callback `beforeSetContent` seems like a good place to start.

TimonDB:
@niknetniko What do you mean by "We'll need to localize the next/previous buttons"?

niknetniko:
> @niknetniko What do you mean by "We'll need to localize the next/previous buttons"?

The [docs](https://dbrekalo.github.io/simpleLightbox/) mention these strings, which I assume are only available in English by default:

```
// customize / localize controls captions
closeBtnCaption: 'Close',
nextBtnCaption: 'Next',
prevBtnCaption: 'Previous',
loadingCaption: 'Loading...',
```

These are displayed on hover:
image

(Unless we don't display the controls, in which case this isn't really needed).

TimonDB:
When trying to hook in MathJax to check for math in the captions using the `beforeContentSet` callback, this didn't seem to work as this is probably executed before the content is created. So instead I searched for another lightbox implementation that does have a `onShow` callback to hook MathJax. With [GLightbox](https://biati-digital.github.io/glightbox/#), it all seems to work fine, so if OK for everyone, I think it is better to use this implementation

app/assets/javascripts/exercise.js
```diff
@@ -77,14 +80,14 @@ function onFrameMessage(event) {
...
+    let index = 0;
     const images = [];
     $(".activity-description img, a.dodona-lightbox").each(function () {
         const imagesrc = $(this).data("large") || $(this).attr("src") || $(this).attr("href");
         const altText = $(this).data("caption") || $(this).attr("alt") || imagesrc.split("/").pop();
         const imageObject = {
-            url: imagesrc,
-            caption: altText,
+            href: imagesrc,
+            title: altText,
```
niknetniko, line 90: Using title here results in a bit of unnecessary padding:
image

- I think we could either use `description` (and increase the description's font size a bit) or remove the padding with css.
- Additionally, in dark mode, we want to make this caption background black, with white text. Or, if that's difficult to do for only dark mode, always make the background black and text white, like the current lightbox implementation.

niknetniko (changes requested):
Glightbox looks good, and is MIT licensed 👍

The code looks good, just two minor UI things.

niknetniko (approved):

chvp (approved):

app/assets/javascripts/exercise.js
```diff
@@ -56,14 +57,16 @@ function initLabelsEdit(labels, undeletableLabels) {
...
-        onShow: function () {
-            // There might have been math in the image captions, so ask
-            // MathJax to search for new math (but only in the captions).
-            window.MathJax.typeset([".strp-caption"]);
-        }
-    }, content.index);
+    const lightbox = new GLightbox({
+        elements: content.elements,
+        startAt: content.startAt,
+    });
```
bmesuere, line 63: Is there a reason why you don't pass `content` directly to the constructor?

app/assets/stylesheets/libraries/glightbox.css.scss
```diff
@@ -0,0 +1,10 @@
+@import "../../../../node_modules/glightbox/dist/css/glightbox";
+
+.glightbox-clean .gslide-description {
+    background: none;
+    font-size: 1.4em;
+}
+
+.gslide-desc {
+    color: white;
+}
```
bmesuere, line 10:
```suggestion
}

```

vendor/assets/javascripts/vendor.js
```diff
@@ -1,2 +0,0 @@
-// strip
-//=require strip/strip.pkgd.min.js
```
bmesuere, line 2: This was the last remaining vendored file. The file is now empty so this, the parent directory and the require in app/assets/javascripts/application.js could be removed.

@chvp What do you think? We will probably never add new vendored files here because they use sprockets instead of webpacker.

bmesuere (changes requested):

vendor/assets/javascripts/vendor.js
chvp: Agreed.

app/assets/javascripts/exercise.js
TimonDB: Not really, I didn't know you could directly pass `content`

bmesuere (approved):

PR #3014, opened 2021-08-24
Add keyboard shortcut to go to next student in evaluation

### LINKED ISSUE #2817: Add shortcut to go to next student in evaluation
With tab key teachers can enter the scores easily. Adding a shortcut (e.g. `Ctrl+Enter` similar to `Shift+Enter` for annotations) to go the next student would improve the UX.

TimonDB:
This pull request implements a keyboard shortcut `ctrl+enter` to go to the next student in the evaluation.

A tooltip to the *next*-button is added to increase discoverabiltiy.
image

Closes #2817.


chvp (changes requested):
If we implement this, this needs to be discoverable somehow, IMO. The shortcut in the annotation form is documented by some hint text beneath the text area. Not sure where we could put this here.

BramDevlaminck:
Another idea might be to add a tooltip that shows that CTRL+ENTER achieves the same as clicking on the button?

image

BTWS2:
> Another idea might be to add a tooltip that shows that CTRL+ENTER achieves the same as clicking on the button?
> 
> image

That's also ok for me.

chvp (commented):
Not sure whether I like Ctrl+Enter as a shortcut. I often press this when annotating because it's used as submit in a lot of contexts (e.g. GitHub).

PR #3015, opened 2021-08-24
Provide correct user URL in submissions.json

### LINKED ISSUE #2732: API: no access to /users/....json from submissions.json
I'm trying to get the name of users by looking at a json like https://dodona.ugent.be/nl/courses/439/submissions.json. 

Each user has a url in that json, but accessing that url gives a `403: "Forbidden"` (https://dodona.ugent.be/nl/users/5235.json)
Accessing trough the course (https://dodona.ugent.be/nl/courses/439/members/5235.json)
gives a `406: "Not Acceptable"`

TimonDB:
This pull request provides the correct user URL in the submissions json. The `user/:id` URL was provided, however, a course admin can only view a user profile via the `courses/:id/members/:id` URL.

Closes #2732.

### DIFF
--- a/app/controllers/course_members_controller.rb
+++ b/app/controllers/course_members_controller.rb
@@ -33,6 +33,10 @@ class CourseMembersController < ApplicationController
     @course_labels = CourseLabel.where(course: @course)
     @series = policy_scope(@course.series)
     @series_loaded = 5
+    @users_lables = @course.course_memberships
+                           .includes(:course_labels, :user)
+                           .map { |m| [m.user, m.course_labels] }
+                           .to_h
   end
 
   def edit
--- a/app/controllers/course_members_controller.rb
+++ b/app/controllers/course_members_controller.rb
@@ -33,7 +33,7 @@ class CourseMembersController < ApplicationController
     @course_labels = CourseLabel.where(course: @course)
     @series = policy_scope(@course.series)
     @series_loaded = 5
-    @users_lables = @course.course_memberships
+    @users_labels = @course.course_memberships
                            .includes(:course_labels, :user)
                            .map { |m| [m.user, m.course_labels] }
                            .to_h
--- a/app/controllers/course_members_controller.rb
+++ b/app/controllers/course_members_controller.rb
@@ -33,10 +33,6 @@ class CourseMembersController < ApplicationController
     @course_labels = CourseLabel.where(course: @course)
     @series = policy_scope(@course.series)
     @series_loaded = 5
-    @users_labels = @course.course_memberships
-                           .includes(:course_labels, :user)
-                           .map { |m| [m.user, m.course_labels] }
-                           .to_h
   end
 
   def edit
--- a/app/views/course_members/index.json.jbuilder
+++ b/app/views/course_members/index.json.jbuilder
@@ -1,5 +1,3 @@
 json.array! @course_memberships do |cm|
-  json.extract! cm.user, :id, :username, :first_name, :last_name, :email
-  json.status cm.status
-  json.url course_member_url(cm.course, cm.user, format: :json)
+  json.partial! 'course_member_data', locals: { user: cm.user, course_membership: cm }
 end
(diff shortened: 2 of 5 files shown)


app/controllers/course_members_controller.rb
```diff
@@ -33,6 +33,10 @@ def show
     @course_labels = CourseLabel.where(course: @course)
     @series = policy_scope(@course.series)
     @series_loaded = 5
+    @users_lables = @course.course_memberships
```
niknetniko, line 36:
```suggestion
    @users_labels = @course.course_memberships
```

app/views/course_members/show.json.jbuilder
```diff
@@ -0,0 +1,8 @@
+json.extract! @user,
+              :id,
+              :username,
+              :first_name,
+              :last_name,
+              :email
+json.status @course_membership.status
+json.labels @users_lables[@user], :name
```
niknetniko, line 8:
```suggestion
json.labels @users_labels[@user], :name
```

app/views/submissions/_submission_basic.json.jbuilder
```diff
@@ -1,6 +1,6 @@
 json.extract! submission, :created_at, :status, :summary, :accepted, :id
 json.url submission_url(submission, format: :json)
-json.user user_url(submission.user, format: :json)
+json.user course_member_url(course_id: submission.course, id: submission.user.id, format: :json) if submission.course
```
niknetniko, line 3: Should we provide the `user_url` if there is no course attached to the submission? This might be useful if you are getting your own submissions, but not sure if that happens.

niknetniko (changes requested):

app/views/submissions/_submission_basic.json.jbuilder
chvp: That is how it is done in the HTML, so that seems like the best option in the JSON as well.

app/views/course_members/show.json.jbuilder
chvp: Should probably be tested, if this wasn't caught by the tests already.

chvp: Ah, never mind, didn't see the comment above yet.

niknetniko (approved):

bmesuere (approved):

app/controllers/course_members_controller.rb
```diff
@@ -33,6 +33,10 @@ def show
     @course_labels = CourseLabel.where(course: @course)
     @series = policy_scope(@course.series)
     @series_loaded = 5
+    @users_labels = @course.course_memberships
+                           .includes(:course_labels, :user)
+                           .map { |m| [m.user, m.course_labels] }
+                           .to_h
```
chvp, line 39: It feels a bit wasteful to load all user -> course label mappings to show a single user. The HTML version of this page doesn't need this, so you can look there how the course labels are loaded for a single user. Probably move that here, since it can be shared between the json and html views.

chvp (changes requested):

app/views/course_members/index.json.jbuilder
```diff
@@ -1,4 +1,5 @@
 json.array! @course_memberships do |cm|
-  json.extract! cm.user, :id, :username, :permission, :first_name, :last_name, :email
+  json.extract! cm.user, :id, :username, :first_name, :last_name, :email
+  json.status cm.status
   json.url course_member_url(cm.course, cm.user, format: :json)
```
chvp, line 4: We usually extract the common items between the index and the show to a partial.

chvp (changes requested):

app/views/course_members/index.json.jbuilder
chvp: There are slight differences between the data per user in `index` and in `show`, but I think those differences are not necessary, and can be removed.

chvp (approved):

PR #3021, opened 2021-08-25
Add series navigation on course member page

### LINKED ISSUE #2436: Add series navigation to course user page
The course user page lacks the series navigation in the right margin, that the user itself would find in its own course page. Since the number of series might be equally large on this page, the series navigation might serve the same purpose here.

image

TimonDB:
This pull request adds the series navigation in the right margin of the course member page. Before, this series navigation was only shown on the course page and not on the course member page.

Screenshot from 2021-08-25 11-49-42

Closes #2436.

### DIFF
--- a/app/assets/javascripts/course.js
+++ b/app/assets/javascripts/course.js
@@ -133,7 +133,7 @@ function initCourseShow() {
         const nav = document.getElementById("scrollspy-nav");
         if (nav) {
             new ScrollSpy(nav, {
-                sectionSelector: ".series .anchor",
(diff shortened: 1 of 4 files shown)


niknetniko (commented):
It feels a bit weird that the statistieken header is also underneath the Oefeningenreeks title, but perhaps that's just me

bmesuere:
I would leave out statistics since we also don't add the "course description" item on the course page.

bmesuere (approved):

PR #3032, opened 2021-08-26
Add button to clear all scores when giving feedback

### LINKED ISSUE #2716: Clear grades
The thumbs up/down buttons at the "total"-level fill up all scores, but there is no way to also clear all scores if this actions was mistakenly done. Adding this option has even become more important for evaluations with scores, since filling up all scores has replaced manually marking the evaluation of a solution as completed.

I only felt a need to add a clear button for the top-level actions, but for better UI-alignment similar buttons might also be added to clear the individual score items. EDIT: I only discovered afterwards that the individual score items already have the clear option (paste-bin on the left; would have expected something like a cross on the right); still think it's equally (even more) useful top-level

TimonDB:
This pull request adds a button to the "total"-level to clear all currently given scores.
Also the thumbs up/down buttons at the "total"-level now only fills up empty scores and leaves manually assigned scores untouched.

Screenshot from 2021-08-26 16-57-23

Closes #2716.


bmesuere:
Superseded by #3918

PR #3049, opened 2021-08-30
List own courses first when copying a course

### LINKED ISSUE #1661: Copy course page does not list own courses first
I was going to copy the 2018-2019 course page from last year when creating a new course.

When I wanted to select the course to copy from, the list did not contain courses I was admin for on the first page. I feel like it would make sense to put these courses first initially.

https://dodona.ugent.be/nl/courses/new/

TimonDB:
This pull request lists the user's own courses (course admin) first in the course table when copying a course. To accomplish this, I split up the courses into 2 sets, one where the user is course admin of the courses and one where the user is not course admin.

Closes #1661.

### DIFF
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -29,7 +29,10 @@ class CoursesController < ApplicationController
     elsif params[:tab] == 'featured'
       @courses = @courses.where(featured: true)
     elsif params[:copy_courses]
-      @courses = @courses.reorder(featured: :desc, year: :desc, name: :asc)
+      @own_courses = @courses.joins("INNER JOIN course_memberships ON courses.id = course_memberships.course_id AND course_memberships.user_id = #{@current_user.id} AND course_memberships.status = 1")
+                             .reorder(featured: :desc, year: :desc, name: :asc)
+      @other_courses = @courses.joins("INNER JOIN course_memberships ON courses.id = course_memberships.course_id AND course_memberships.user_id = #{@current_user.id} AND course_memberships.status != 1")
+                               .reorder(featured: :desc, year: :desc, name: :asc)
     end
 
     @courses = apply_scopes(@courses)
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -29,10 +29,13 @@ class CoursesController < ApplicationController
     elsif params[:tab] == 'featured'
       @courses = @courses.where(featured: true)
     elsif params[:copy_courses]
-      @own_courses = @courses.joins("INNER JOIN course_memberships ON courses.id = course_memberships.course_id AND course_memberships.user_id = #{@current_user.id} AND course_memberships.status = 1")
-                             .reorder(featured: :desc, year: :desc, name: :asc)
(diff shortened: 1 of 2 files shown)


app/controllers/courses_controller.rb
```diff
@@ -29,7 +29,10 @@ def index
     elsif params[:tab] == 'featured'
       @courses = @courses.where(featured: true)
     elsif params[:copy_courses]
-      @courses = @courses.reorder(featured: :desc, year: :desc, name: :asc)
+      @own_courses = @courses.joins("INNER JOIN course_memberships ON courses.id = course_memberships.course_id AND course_memberships.user_id = #{@current_user.id} AND course_memberships.status = 1")
+                             .reorder(featured: :desc, year: :desc, name: :asc)
+      @other_courses = @courses.joins("INNER JOIN course_memberships ON courses.id = course_memberships.course_id AND course_memberships.user_id = #{@current_user.id} AND course_memberships.status != 1")
+                               .reorder(featured: :desc, year: :desc, name: :asc)
```
chvp, line 35: This removes all courses from the list where the user is not registered. I think a secondary sort in ruby code would be better. You can then just sort on the result of `current_user.course_admin?(course)`.

chvp (changes requested):

app/controllers/courses_controller.rb
bmesuere: Agreed, I think we always want featured courses first, next own courses, next other. Within each category sort by year (desc) and then name (asc).

TimonDB: OK, I'll adapt it

app/controllers/courses_controller.rb
```diff
@@ -29,7 +29,13 @@ def index
...
       @courses = @courses.where(featured: true)
     elsif params[:copy_courses]
-      @courses = @courses.reorder(featured: :desc, year: :desc, name: :asc)
+      @featured_courses = @courses.where(featured: true)
+      @own_courses = @courses.where(featured: false)
+                             .reorder(year: :desc, name: :asc)
+                             .select { |course| current_user.course_admin?(course) }
+      @other_courses = @courses.where(featured: false)
+                               .reorder(year: :desc, name: :asc)
+                               .reject { |course| current_user.course_admin?(course) }
```
chvp, line 38:
```suggestion
      featured_courses = @courses.where(featured: true)
      own_courses = @courses.where(featured: false)
                             .reorder(year: :desc, name: :asc)
                             .select { |course| current_user.course_admin?(course) }
      other_courses = @courses.where(featured: false)
                               .reorder(year: :desc, name: :asc)
                               .reject { |course| current_user.course_admin?(course) }
      @courses = featured_courses + own_courses + other_courses
```
This way, no changes to the HTML are necessary. (Make sure to check this locally, I'm not entirely sure if the `+` works if the actual data hasn't been loaded yet.)

chvp (changes requested):

app/controllers/courses_controller.rb
TimonDB: This seems to result in an error: 
Screenshot from 2021-08-30 16-50-11

It would be nice if that was possible because now there is a lot of redundant code in the HTML

chvp: You can look at the server log to see the actual error. Probably an explicit load is necessary for the `featured_courses`.

PR #3051, opened 2021-08-30
Improve CI ruby test speed

### LINKED ISSUE #3016: Improve CI ruby test speed
The ruby tests are slow using GitHub Actions. While rails supports running tests in parallel, this is disabled in our config. The reason for this was that codecov reported incorrect results when running tests in parallel (only the results of 1 of the threads were uploaded).

This was a few years ago and is probably already fixed at the codecov side.

TimonDB:
This pull request enables rails parallel testing to improve the CI ruby test speed.

Closes #3016.


bmesuere:
Turns out de tests were already parallel 😢

PR #3056, opened 2021-08-31
Fix searching users using labels

### LINKED ISSUE #2524: Filter course scoresheet with label
Filtering series with labels works as expected, e.g. [https://dodona.ugent.be/en/series/4715/scoresheet/?course_labels%5B%5D=gg10&page=1](https://dodona.ugent.be/en/series/4715/scoresheet/?course_labels%5B%5D=gg10&page=1):
image

Filtering the course scoresheet with a label doesn't, e.g. [https://dodona.ugent.be/en/courses/440/scoresheet/?course_labels%5B%5D=gg10&page=1](https://dodona.ugent.be/en/courses/440/scoresheet/?course_labels%5B%5D=gg10&page=1):
image

TimonDB:
This pull request enables searching users using the course labels, which was broken on every page. This was due to the new implementation of the URL API, where array query parameters were not followed by a `[]`, which is necessary when using Rails. Also fixes the original issue where searching users via course labels was only broken on the scoresheet page.

Closes #2524.

### DIFF
--- a/app/assets/javascripts/util.js
+++ b/app/assets/javascripts/util.js
@@ -26,9 +26,9 @@ function updateURLParameter(_url, param, paramVal) {
 
 function updateArrayURLParameter(_url, param, _paramVals) {
     const paramVals = new Set(_paramVals); // remove duplicate items
-    let url = new URL(_url, window.location.origin);
-    // decode "%5B%5D" back to "[]" in URL so the parameter can be deleted
-    url = new URL(decodeURI(url));
+    
+    // use decodeURI to convert "%5B%5D" back to "[]"
+    const url = new URL(decodeURI(_url), window.location.origin);
     url.searchParams.delete(`${param}[]`);
     paramVals.forEach(paramVal => {
         url.searchParams.append(`${param}[]`, paramVal);
--- a/app/assets/javascripts/util.js
+++ b/app/assets/javascripts/util.js
@@ -29,7 +29,7 @@ function updateArrayURLParameter(_url, param, _paramVals) {
     const url = new URL(_url, window.location.origin);
     url.searchParams.delete(param);
     paramVals.forEach(paramVal => {
-        url.searchParams.append(param, paramVal);
+        url.searchParams.append(`${param}[]`, paramVal);
     });
     return url.toString();
 }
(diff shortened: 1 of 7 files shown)


app/assets/javascripts/util.js
```diff
@@ -26,10 +26,12 @@ function updateURLParameter(_url, param, paramVal) {
 
 function updateArrayURLParameter(_url, param, _paramVals) {
     const paramVals = new Set(_paramVals); // remove duplicate items
-    const url = new URL(_url, window.location.origin);
-    url.searchParams.delete(param);
+
+    // use decodeURI to convert "%5B%5D" back to "[]"
+    const url = new URL(decodeURI(_url), window.location.origin);
```
chvp, line 31: I'm not sure if a full decodeURI is the way to go. What happens if e.g. a label contains a character that is usually percent-encoded?

TimonDB: hmmm fair point, didn't think of that... But maybe we can use `_url.replace("%5B%5D", "[]")` to get the brackets back into the url string? That might be a simple solution for this issue, I only think of that now

chvp: A replace sounds fine to me.

app/assets/javascripts/util.js
```diff
@@ -26,10 +26,11 @@ function updateURLParameter(_url, param, paramVal) {
 
 function updateArrayURLParameter(_url, param, _paramVals) {
     const paramVals = new Set(_paramVals); // remove duplicate items
-    const url = new URL(_url, window.location.origin);
-    url.searchParams.delete(param);
+    // convert "%5B%5D" back to "[]"
+    const url = new URL(_url.split("%5B%5D").join("[]"), window.location.origin);
```
bmesuere, line 30: Use replaceAll here

bmesuere (changes requested):

app/assets/javascripts/util.js
TimonDB: I did try to use replaceAll, but that gives an error: 
Screenshot from 2021-09-02 09-15-02

chvp: Your checker is probably just confused about the type. This is not typescript, so it shouldn't give any linting errors. The function [does exist](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replaceAll), so it should work.

chvp (changes requested):
This pull request also seems to contain the changes for #3058. Can you remove those?

chvp (approved):

bmesuere (approved):

PR #3058, opened 2021-09-01
Enable user search in evaluations

### LINKED ISSUE #2824: Allow user search via course label in evaluations
Both creating and viewing.

From https://github.com/dodona-edu/dodona/issues/1978#issuecomment-638678942

TimonDB:
This pull request enables the searchbar in the evaluations.  
Note: it needs #3056 to work correctly 

Screenshot from 2021-09-01 14-16-06

Closes #2824.

### DIFF
--- a/app/views/evaluations/show.html.erb
+++ b/app/views/evaluations/show.html.erb
@@ -40,6 +40,7 @@
     </div>
     <div class="card">
       <div class="card-supporting-text">
+      <%= render partial: 'layouts/searchbar', locals: { course_labels: @course_labels } %>
         <div class="card-tab">
           <ul id="user-tabs" class="nav nav-tabs" role="tablist">
             <li role="presentation">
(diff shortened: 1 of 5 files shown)


app/views/evaluations/show.html.erb
```diff
@@ -67,8 +70,9 @@
                     </div>
                   </div>
                 </div>
-                <p class="help-block"></p>
```
bmesuere, line 70: This help paragraph was removed. @niknetniko can you remember if we used this for something?

app/views/evaluations/show.js.erb
```diff
@@ -0,0 +1,2 @@
+$("#evaluation-table-wrapper").html("<%= escape_javascript(render 'evaluation_table', feedbacks: @feedbacks, evaluation: @evaluation, users: @users) %>")
```
bmesuere, line 1: This could be optimized to only update the visible table, but it would leave it like this for now.

app/views/evaluations/show.html.erb
niknetniko: I think this was a mistake, and can be safely removed.

bmesuere (approved):

chvp (approved):

PR #3061, opened 2021-09-02
Support multiple callout message formats

inteon:
Also allow "callout-info", "callout-warning" and "callout-danger" as message formats.

### DIFF
--- a/app/helpers/renderers/feedback_table_renderer.rb
+++ b/app/helpers/renderers/feedback_table_renderer.rb
@@ -295,6 +295,11 @@ class FeedbackTableRenderer
         # `markdown` is always safe
         @builder << markdown(m[:description])
       end
+    elsif m[:format].in?(%w[callout-info callout-warning callout-danger])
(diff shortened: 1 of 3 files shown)


chvp (commented):
This needs changes to the JSON schema I think. However, I'm not entirely sure this is something we want to introduce? I don't think we want to facilitate an overuse of callouts? But I'll let @bmesuere be the final judge on that.

bmesuere (changes requested):
I'm also not super fond of callouts in the feedback table, but I can see why people would use them. Since we already support the default callout, it's reasonable to add the other classes. If the alternative is achieving the same result using html messages, I very much prefer using a message type since we then have more control over it.

The message types are not listed explicitly in the schema (so the output would validate as is), but the options are listed in the description. Can you add them to [https://github.com/dodona-edu/dodona/tree/develop/public/schemas](https://github.com/dodona-edu/dodona/tree/develop/public/schemas)?

bmesuere (approved):

chvp (approved):

PR #3065, opened 2021-09-02
Displaying & diffing judge CSV output

inteon:
This pull requests add support for diffing and displaying CSV output.

When a judge outputs its expected and generated results as csv, the default line based diff doesn't give a great overview of the errors in the output. This is resolved with this pull request by adding an optional `format` field to tests in the judge output. When the format is set to csv, a separate differ is used which displays the differences using a table.

There are fallbacks for large outputs to limit the feedback table generation time.

If a lot of columns are returned in the csv, a horizontal scrollbar will be shown.

diff identical:
image
diff row missing:
image
diff too many rows:
image
diff too many rows (unified):
image
diff wrong column:
image
diff wrong column (unified):
image
diff wrong column name:
image
horizontal scrolling (lots of columns):
image

### DIFF
--- a/app/assets/stylesheets/models/submissions.css.scss
+++ b/app/assets/stylesheets/models/submissions.css.scss
@@ -158,16 +158,10 @@
       white-space: pre;
     }
   }
-  .diffs .unified-diff {
+  .diffs .unified-diff, .diffs .split-diff {
     display: none;
   }
-  .diffs .split-diff {
-    display: none;
-  }
-  .diffs.show-unified .unified-diff {
-    display: table;
-  }
-  .diffs.show-split .split-diff {
+  .diffs.show-unified .unified-diff, .diffs.show-split .split-diff {
     display: table;
   }
   .diffs.show-split .split-diff.csv-diff {
@@ -175,7 +169,7 @@
     width: 50%;
     vertical-align: top;
   }
-  .diffs.show-unified .unified-diff.csv-diff {
+  .diffs.show-unified .unified-diff.csv-diff, .test-accepted .diffs .diff {
     display: inline-block;
   }
 
--- a/app/assets/stylesheets/models/submissions.css.scss
+++ b/app/assets/stylesheets/models/submissions.css.scss
@@ -134,7 +134,7 @@
       color: $feedback-diff-table-header-fg;
       padding-left: 10px;
       padding-right: 10px;
-      width: 0px; /* minimize the width */
+      width: 0; /* minimize the width */
     }
     .line-nr:empty::before {
       content: " "; /* make sure the line-height is respected */
--- a/app/assets/stylesheets/models/submissions.css.scss
+++ b/app/assets/stylesheets/models/submissions.css.scss
@@ -134,6 +134,10 @@
       color: $feedback-diff-table-header-fg;
       padding-left: 10px;
       padding-right: 10px;
+      width: 0px; /* minimize the width */
+    }
+    .line-nr:empty::before {
+      content: " "; /* make sure the line-height is respected */
     }
     .output {
       width: 100%;
@@ -149,6 +153,16 @@
       line-height: 1.0;
     }
   }
+  table.diff.csv-diff {
+    td, .del, .ins, .unchanged {
+      padding: 2px 6px;
+    }
+  }
+  table.diff.csv-diff:not(.simplified) {
+    td, .del, .ins, .unchanged {
+      white-space: nowrap;
+    }
+  }
   .diffs .unified-diff {
     display: none;
   }
@@ -160,6 +174,14 @@
   }
   .diffs.show-split .split-diff {
     display: table;
+  }
+  .diffs.show-split .split-diff.csv-diff {
+    display: inline-block;
+    width: 50%;
+    vertical-align: top;
+  }
+  .diffs.show-unified .unified-diff.csv-diff {
+    display: inline-block;
   }
 
   .feedback-table-options {
--- a/app/assets/stylesheets/models/submissions.css.scss
+++ b/app/assets/stylesheets/models/submissions.css.scss
@@ -157,6 +157,13 @@
       padding: 2px 6px;
       white-space: pre;
     }
+    td.line-nr, th.line-nr {
+      position: -webkit-sticky;
+      position: sticky;
+      left: 0;
+      vertical-align: top;
+      text-align: right;
+    }
   }
   .diffs .unified-diff, .diffs .split-diff {
     display: none;
--- a/app/assets/stylesheets/models/submissions.css.scss
+++ b/app/assets/stylesheets/models/submissions.css.scss
@@ -160,7 +160,7 @@
   }
   table.diff.csv-diff:not(.simplified) {
     td, .del, .ins, .unchanged {
-      white-space: nowrap;
+      white-space: pre;
     }
   }
   .diffs .unified-diff {
--- a/app/assets/stylesheets/models/submissions.css.scss
+++ b/app/assets/stylesheets/models/submissions.css.scss
@@ -134,7 +134,6 @@
       color: $feedback-diff-table-header-fg;
       padding-left: 10px;
       padding-right: 10px;
-      width: 0; /* minimize the width */
     }
     .line-nr:empty::before {
       content: " "; /* make sure the line-height is respected */
--- a/app/assets/stylesheets/models/submissions.css.scss
+++ b/app/assets/stylesheets/models/submissions.css.scss
@@ -155,10 +155,6 @@
   table.diff.csv-diff {
     td, .del, .ins, .unchanged {
       padding: 2px 6px;
-    }
-  }
-  table.diff.csv-diff:not(.simplified) {
-    td, .del, .ins, .unchanged {
       white-space: pre;
     }
   }
--- a/app/helpers/renderers/csv_differ.rb
+++ b/app/helpers/renderers/csv_differ.rb
@@ -46,9 +46,7 @@ class CsvDiffer
     @exp_headers, *@expected = @expected
     @exp_headers ||= []
 
-    @generated_rowcount = @generated.length
-    @expected_rowcount = @expected.length
-    @simplified_table = @generated_rowcount > 100 || @expected_rowcount > 100
+    @simplified_table = @generated.length > 100 || @expected.length > 100
 
     @gen_header_indices, @exp_header_indices, @gen_headers, @exp_headers, @combined_headers = diff_header_indices(@gen_headers, @exp_headers)
 
@@ -164,9 +162,9 @@ class CsvDiffer
       builder.tbody do
         if @simplified_table
           if is_old
-            simple_old_row(builder)
+            simple_row(builder, @generated, 'del')
           else
-            simple_new_row(builder)
+            simple_row(builder, @expected, 'ins')
           end
         else
           @diff.each do |chunk|
@@ -191,7 +189,7 @@ class CsvDiffer
     gen_cols = @generated.transpose.map { |col| col.join("\n") }
     builder.tr do
       builder.td(class: 'line-nr') do
-        builder << (1..@generated_rowcount).to_a.join("\n")
+        builder << (1..@generated.length).to_a.join("\n")
       end
       builder.td(class: 'line-nr')
 
@@ -208,7 +206,7 @@ class CsvDiffer
     builder.tr do
       builder.td(class: 'line-nr')
       builder.td(class: 'line-nr') do
-        builder << (1..@expected_rowcount).to_a.join("\n")
+        builder << (1..@expected.length).to_a.join("\n")
       end
 
       builder << Array.new(@combined_headers.length) { |i| @exp_header_indices.index(i) }.map do |idx|
@@ -221,30 +219,15 @@ class CsvDiffer
     end
   end
 
-  def simple_old_row(builder)
-    gen_cols = @generated.transpose.map { |col| col.join("\n") }
+  def simple_row(builder, data, cls)
+    gen_cols = data.transpose.map { |col| col.join("\n") }
 
     builder.tr do
       builder.td(class: 'line-nr') do
-        builder << (1..@generated_rowcount).to_a.join("\n")
+        builder << (1..data.length).to_a.join("\n")
       end
       gen_cols.each do |col|
-        builder.td(class: 'del') do
-          builder << CGI.escape_html(col)
-        end
-      end
-    end
-  end
-
-  def simple_new_row(builder)
-    exp_cols = @expected.transpose.map { |col| col.join("\n") }
-
-    builder.tr do
-      builder.td(class: 'line-nr') do
-        builder << (1..@expected_rowcount).to_a.join("\n")
-      end
-      exp_cols.each do |col|
-        builder.td(class: 'ins') do
+        builder.td(class: cls) do
           builder << CGI.escape_html(col)
         end
       end
--- a/app/helpers/renderers/csv_differ.rb
+++ b/app/helpers/renderers/csv_differ.rb
@@ -134,14 +134,14 @@ class CsvDiffer
 
   private
 
-  def split_build_table(builder, headers, is_old)
+  def split_build_table(builder, headers, is_generated_output)
     builder.table(class: 'split-diff diff csv-diff') do
       builder.colgroup do
         builder.col(class: 'line-nr')
         builder.col(span: headers.length)
       end
       builder.thead do
-        if is_old
+        if is_generated_output
           icon_cls = 'mdi-file-account'
           title = I18n.t('submissions.show.your_output')
         else
@@ -161,7 +161,7 @@ class CsvDiffer
       end
       builder.tbody do
         if @simplified_table
-          if is_old
+          if is_generated_output
             simple_row(builder, @generated, 'del')
           else
             simple_row(builder, @expected, 'ins')
@@ -169,7 +169,7 @@ class CsvDiffer
         else
           @diff.each do |chunk|
             builder.tr do
-              if is_old
+              if is_generated_output
                 is_empty, row = old_row(chunk)
                 position = chunk.old_position
               else
--- a/app/helpers/renderers/csv_differ.rb
+++ b/app/helpers/renderers/csv_differ.rb
@@ -1,4 +1,4 @@
-class DiffCsv
+class CsvDiffer
   require 'builder'
 
   def self.render_accepted(builder, generated)
@@ -7,23 +7,23 @@ class DiffCsv
 
     return if gen_headers.blank?
 
-    builder.div(class: 'diffs show-unified') do
-      builder.table(class: 'unified-diff diff csv-diff') do
+    builder.div(class: 'diffs') do
+      builder.table(class: 'diff csv-diff') do
         builder.colgroup do
           builder.col(class: 'line-nr')
-          builder.col(class: 'del-output-csv', span: gen_headers.length)
+          builder.col(span: gen_headers.length)
         end
         builder.thead do
           builder.tr do
-            builder << "<th class='line-nr'></th>"
+            builder.th(class: 'line-nr')
             builder << gen_headers.map { |el| %(<th>#{CGI.escape_html el}</th>) }.join
           end
         end
         builder.tbody do
           generated.each.with_index do |line, idx|
             builder.tr do
-              builder << %(<td class="line-nr">#{idx + 1}</td>)
-              builder << line.map { |el| %(<td>#{CGI.escape_html el || ''}</td>) }.join
+              builder.th(idx + 1, class: 'line-nr')
+              builder << line.map { |el| %(<td>#{CGI.escape_html el}</td>) }.join
             end
           end
         end
@@ -67,17 +67,21 @@ class DiffCsv
       builder.colgroup do
         builder.col(class: 'line-nr')
         builder.col(class: 'line-nr')
-        builder.col(class: 'output-csv', span: @combined_headers.length)
+        builder.col(span: @combined_headers.length)
       end
       builder.thead do
         builder.tr do
-          builder << "<th class='line-nr' title='#{I18n.t('submissions.show.your_output')}'><i class='mdi mdi-18 mdi-file-account'/></th>"
-          builder << "<th class='line-nr' title='#{I18n.t('submissions.show.expected')}'><i class='mdi mdi-18 mdi-file-check'/></th>"
-          builder << "<th colspan='#{@combined_headers.length}'>#{I18n.t('submissions.show.your_output')}</th>"
+          builder.th(class: 'line-nr', title: I18n.t('submissions.show.your_output')) do
+            builder.i(class: 'mdi mdi-18 mdi-file-account')
+          end
+          builder.th(class: 'line-nr', title: I18n.t('submissions.show.expected')) do
+            builder.i(class: 'mdi mdi-18 mdi-file-check')
+          end
+          builder.th(colspan: @combined_headers.length)
         end
         builder.tr do
-          builder << "<th class='line-nr'></th>"
-          builder << "<th class='line-nr'></th>"
+          builder.th(class: 'line-nr')
+          builder.th(class: 'line-nr')
           builder << @combined_headers.join
         end
       end
@@ -128,27 +132,26 @@ class DiffCsv
 
   def split_build_table(builder, headers, is_old)
     builder.table(class: 'split-diff diff csv-diff') do
-      if is_old
-        cls = 'del-output-csv'
-        icon_cls = 'mdi-file-account'
-        title = I18n.t('submissions.show.your_output')
-      else
-        cls = 'ins-output-csv'
-        icon_cls = 'mdi-file-check'
-        title = I18n.t('submissions.show.expected')
-      end
-
       builder.colgroup do
         builder.col(class: 'line-nr')
-        builder.col(class: cls, span: headers.length)
+        builder.col(span: headers.length)
       end
       builder.thead do
-        builder.tr do
-          builder << "<th class='line-nr' title='#{title}'><i class='mdi mdi-18 #{icon_cls}'/></th>"
-          builder << "<th colspan='#{headers.length}'>#{title}</th>"
+        if is_old
+          icon_cls = 'mdi-file-account'
+          title = I18n.t('submissions.show.your_output')
+        else
+          icon_cls = 'mdi-file-check'
+          title = I18n.t('submissions.show.expected')
         end
         builder.tr do
-          builder << "<th class='line-nr'></th>"
+          builder.th(class: 'line-nr', title: title) do
+            builder.i(class: %(mdi mdi-18 #{icon_cls}))
+          end
+          builder.th(title, colspan: headers.length)
+        end
+        builder.tr do
+          builder.th(class: 'line-nr')
           builder << headers.join
         end
       end
--- a/app/helpers/renderers/csv_differ.rb
+++ b/app/helpers/renderers/csv_differ.rb
@@ -1,6 +1,11 @@
 class CsvDiffer
   require 'builder'
 
+  # Render a table displaying the accepted output
+  #
+  # @param [Builder::XmlMarkup] builder   The xml builder that is used to generate the html
+  # @param [String]             generated The csv-encoded output as received from the judge
+  # @return [Builder::XmlMarkup]          Generated table as an xml object
   def self.render_accepted(builder, generated)
     generated = CSV.parse((generated || '').lstrip, nil_value: '')
     gen_headers, *generated = generated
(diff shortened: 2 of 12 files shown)


app/assets/stylesheets/models/submissions.css.scss
```diff
@@ -134,6 +134,10 @@
       color: $feedback-diff-table-header-fg;
       padding-left: 10px;
       padding-right: 10px;
+      width: 0px; /* minimize the width */
```
thepieterdc, line 137: Unit is not required if amount is 0
```suggestion
      width: 0; /* minimize the width */
```

app/helpers/renderers/feedback_table_renderer.rb
```diff
@@ -257,7 +257,11 @@ def messages(msgs)
 
   def test_accepted(t)
     @builder.div(class: 'test-accepted') do
-      @builder.span(t[:generated], class: 'output')
+      if t[:format] == 'csv'
+        LCSCsvDiffer.render_accepted(@builder, t[:generated])
+      else
+        LCSHtmlDiffer.render_accepted(@builder, t[:generated])
+      end
```
thepieterdc, line 264: Could get rid of the duplicated `if`-statement everywhere if you make a function `differ(t)` that returns either the `LCSCsvDiffer` or `LCSHtmlDiffer`.

public/schemas/judge_output.json
```diff
@@ -89,7 +90,14 @@
         "columns": { "$ref": "#/definitions/index" }
       }
     },
-    "format": {
+    "test-format": {
+      "type": "string",
+      "enum": [
+        "text",
+        "csv"
+      ]
```
thepieterdc, line 98: Indicate the default value here in the description

app/helpers/renderers/lcs_csv_differ.rb
```diff
@@ -0,0 +1,343 @@
...
+          unified_simple builder
+        else
+          @diff.each do |chunk|
+            row = old_row chunk, false
+
+            unless row.nil?
+              full_row = Array.new(@combined_headers.length) { '<td></td>' }
+              row.each.with_index do |el, idx|
+                full_row[@gen_header_indices[idx]] = el
+              end
```
thepieterdc, line 94: Can oneline this with `map`.

pdawyndt:
Is there support for large tables (say > 20 rows, > 10 colums). We might need some specific strategies to handle those:

- specific handling if number of rows differs, e.g. explicitly show the number of rows (might be useful to always report the number of rows)
- specific handling if number of cols differs, e.g. explicitly show the number of cols (might be useful to always report the number of cols)
- show only selection of rows, e.g. always show first few rows (irrespective if they are correct or not), and only show a limited number of wrong rows with indications where rows have been left out and where rows are situated (e.g. row numbers)
- show only some columns: some good thinking needed here

inteon:
> Is there support for large tables (say > 20 rows, > 10 colums). We might need some specific strategies to handle those:
> 
> * specific handling if number of rows differs, e.g. explicitly show the number of rows (might be useful to always report the number of rows)
> * specific handling if number of cols differs, e.g. explicitly show the number of cols (might be useful to always report the number of cols)
> * show only selection of rows, e.g. always show first few rows (irrespective if they are correct or not), and only show a limited number of wrong rows with indications where rows have been left out and where rows are situated (e.g. row numbers)
> * show only some columns: some good thinking needed here

Currently, the column problem is mostly solved by horizontal scrolling:
image

Large number of rows are just displayed as-is, and if there are more than 100 row, a simplified view is generated to reduce compute time (@bmesuere).

BTWS2:
> Is there support for large tables (say > 20 rows, > 10 colums, possibly lengthy content). We might need some specific strategies to handle those:
> 
> * specific handling if number of rows differs, e.g. explicitly show the number of rows (might be useful to always report the number of rows)
> * specific handling if number of cols differs, e.g. explicitly show the number of cols (might be useful to always report the number of cols)

In the new SQL judge (in development by @inteon) differences between row en column counts are shown (if the counts match, there is no warning):
image


> * show only selection of rows, e.g. always show first few rows (irrespective if they are correct or not), and only show a limited number of wrong rows with indications where rows have been left out and where rows are situated (e.g. row numbers)
> * show only some columns: some good thinking needed here

We have to assume students don't have access to the full database/dataset. I prefer outputting all rows/columns (limited to 100).

> * add ellipsis to lengthy fields
> * may be easier (in some cases) if diffed tables are not shown side-by-side, but unified; but also here there are some options, e.g. unify per cell
> 
> Some inspiration:
> 
> * [daff](https://github.com/paulfitz/daff) (got this one from here [hacker news](https://news.ycombinator.com/item?id=17924798)); Ruby bindings and supports Sqlite format

I find it difficult to see the difference when two tables are unified, I prefer the split view for tables. However, I have to admit daff manages to clearly show the differences in a unified table. 

The output of the student and solution may be shown in a html table in a pop-up (as is, without diff), similar to the Python Tutor pop-up ([example in jsfiddle](https://jsfiddle.net/vxktcn4L/)). This may be especially useful for large datasets, because we have to assume students don't have access to the full dataset.
- Sticky header with types
- Adaptive lay-out with horizontal and vertical scrollbars (when needed). Students sometimes use Dodona in split screen with their IDE.

image

inteon:
Just did some experimentation with the daff online example:
- In most cases the tool yields identical results to this current diff implementation.
- The main improvement of daff over this implementation is the situation where there is a difference in the number of columns and number of rows; in this case daff will better match the columns & rows because it uses the header values to match columns. This is something that we can add too.
- In my opinion, fixing the row/column order should be the responsibility of the judge. This way we can also better adhere to the way the text-base compare works.

I think this current implementation should already be a good baseline for diffing tables.
I propose we implement this feature throughout multiple PRs & limit the scope of this PR to:
- adding format option in judge json output
- adding simple html + csv for table compare
- adding simple differ implementation

Follow-up PRs can include:
- adding sub-headers (eg. datatype in sql)
- updating layout: sticky columns & headers
- adding ellipsis to lengthy fields (maybe expand contents on click)
- adding a popup for better full-screen viewing
- updating the differ implementation

app/assets/stylesheets/models/submissions.css.scss
```diff
@@ -134,6 +134,10 @@
       color: $feedback-diff-table-header-fg;
       padding-left: 10px;
       padding-right: 10px;
+      width: 0; /* minimize the width */
```
bmesuere, line 137: why is this needed? (I mostly concerned with breaking existing things)

app/assets/stylesheets/models/submissions.css.scss
```diff
@@ -134,6 +134,10 @@
...
+    }
+    .line-nr:empty::before {
+      content: " "; /* make sure the line-height is respected */
```
bmesuere, line 140: seems a bit hackish to me

app/assets/stylesheets/models/submissions.css.scss
```diff
@@ -161,6 +175,14 @@
...
     display: table;
   }
+  .diffs.show-split .split-diff.csv-diff {
+    display: inline-block;
+    width: 50%;
+    vertical-align: top;
+  }
+  .diffs.show-unified .unified-diff.csv-diff {
+    display: inline-block;
+  }
```
bmesuere, line 185: Are these table? If so, is there a reason to use inline-block instead of inline-table?

app/helpers/renderers/feedback_table_renderer.rb
```diff
@@ -255,9 +255,17 @@ def messages(msgs)
     end
   end
 
+  def differ(t)
+    if t[:format] == 'csv'
+      LCSCsvDiffer
```
bmesuere, line 260: The naming is inconsistent with `LCSHtmlDiffer`. The html part there refers to the output format of the diff, not the input. Although the CSV differ also uses LCS (if I understand right), maybe CSVDiffer is a better name?

app/helpers/renderers/lcs_csv_differ.rb
```diff
@@ -0,0 +1,336 @@
+class LCSCsvDiffer
+  require 'builder'
+
+  def self.render_accepted(builder, generated)
```
bmesuere, line 4: Why do we need a class method here? The old differ didn't need one.

app/helpers/renderers/lcs_csv_differ.rb
```diff
@@ -0,0 +1,336 @@
+class LCSCsvDiffer
```
bmesuere, line 1: Would we benefit from making a super class and somehow forcing the used functions to be present? Ruby isn't statically typed so abstract methods don't exist, but we can raise errors in base class or use a module to achieve something similar.

app/helpers/renderers/lcs_csv_differ.rb
```diff
@@ -0,0 +1,336 @@
...
+
+    @gen_headers, @generated = @generated.split("\n", 2)
+    @gen_headers = @gen_headers.nil? ? [] : CSV.parse_line(@gen_headers)
+    @gen_headers = @gen_headers.map { |el| el.nil? ? '' : el }
+    @generated ||= ''
+
+    @exp_headers, @expected = @expected.split("\n", 2)
+    @exp_headers = @exp_headers.nil? ? [] : CSV.parse_line(@exp_headers)
+    @exp_headers = @exp_headers.map { |el| el.nil? ? '' : el }
+    @expected ||= ''
```
bmesuere, line 49: duplicate code, so extract to separate function?

app/helpers/renderers/lcs_csv_differ.rb
```diff
@@ -0,0 +1,336 @@
...
+                gen_result = (CSV.parse_line(chunk.old_element || '') || Array.new(@gen_headers.length) { '' }).map { |el| el.nil? ? '' : el }
+                exp_result = (CSV.parse_line(chunk.new_element || '') || Array.new(@exp_headers.length) { '' }).map { |el| el.nil? ? '' : el }
+                if chunk.action == '!'
+                  gen_result, exp_result = diff_arrays(gen_result, exp_result)
+                else
+                  gen_result = gen_result.map { |el| CGI.escape_html el }
+                  exp_result = exp_result.map { |el| CGI.escape_html el }
+                end
+                Diff::LCS::ContextChange.new(chunk.action, chunk.old_position, gen_result, chunk.new_position, exp_result)
+              end
```
bmesuere, line 68: I find it hard to understand this, especially lines 59 and 60, but this could be me. Can you maybe add some comments? You can add newlines in the chained calls and add comments at the end of the line.

app/helpers/renderers/lcs_csv_differ.rb
```diff
@@ -0,0 +1,336 @@
...
+    @exp_headers = @exp_headers.map { |el| el.nil? ? '' : el }
+    @expected ||= ''
+
+    @generated_linecount = @generated&.lines&.count || 0
+    @expected_linecount = @expected&.lines&.count || 0
+    @simplified_table = @generated_linecount > 100 || @expected_linecount > 100
+
+    @gen_header_indices, @exp_header_indices, @gen_headers, @exp_headers, @combined_headers = diff_header_indices(@gen_headers, @exp_headers)
+
+    @diff = unless @simplified_table
```
bmesuere, line 57: What happens in the other case?

public/schemas/judge_output.json
```diff
@@ -89,7 +90,15 @@
         "columns": { "$ref": "#/definitions/index" }
       }
     },
-    "format": {
+    "test-format": {
+      "type": "string",
+      "description": "Indicates how to show the test output (defaults to 'text').",
```
bmesuere, line 95:
```suggestion
      "description": "Indicates how to diff and show the test output (defaults to 'text').",
```

public/schemas/partial_output.json
```diff
@@ -152,12 +153,20 @@
...
-        "format":      { "$ref": "#/definitions/format" },
+        "format":      { "$ref": "#/definitions/message-format" },
         "description": { "$ref": "#/definitions/description" },
         "permission":  { "$ref": "#/definitions/permission" }
       }
     },
-    "format": {
+    "test-format": {
+      "type": "string",
+      "description": "Indicates how to show the test output (defaults to 'text').",
```
bmesuere, line 163:
```suggestion
      "description": "Indicates how to diff and show the test output (defaults to 'text').",
```

bmesuere (changes requested):
I think this can be a valuable addition to Dodona. Not only for the sql judge, but also for other judges. Next to comments I left on the code, I have these general remarks:

- Expand the initial post of the pull request and add more info about the changes and design choices you made. Slightly change the wording of the title as well: this is more about adding a new diffing and display strategy for csv-based output, not about the tests itself.
- Add tests
- As I already said on slack, this is the most performance critical part of Dodona. This code is executed on the webserver (of which we only have one) and handles user-generated content. This can crash the Dodona website. In fact, the few times we had issues, they were all related to the performance of generating the feedback table. I would like you to do a few timings on large test output (e.g., 1000 tests, each with 100+ lines; a user generating 1000+ columns; etc).

app/assets/stylesheets/models/submissions.css.scss
inteon: `inline-table` breaks the horizontal scrolling; `inline-block` is the only display option that worked for me.

app/helpers/renderers/lcs_csv_differ.rb
inteon: The `render_accepted` logic is different for the csv and text output; I moved that logic from `feedback_table_renderer.rb` to `lcs_csv_differ.rb` and `lcs_html_differ.rb` so both can have a different implementation.

app/assets/stylesheets/models/submissions.css.scss
inteon: I'll remove that line, seems to have no effect anymore.

app/assets/stylesheets/models/submissions.css.scss
inteon: In the current implementation; the diff table always has at least some text on each row. This csv diff however, uses 2 tables (1 left and 2 right) and now it is possible to have rows that are totally empty. This hack makes sure that such an empty row has the correct height.

I'll have a 2nd look and try to simplify this bit of css.

app/helpers/renderers/feedback_table_renderer.rb
inteon: It seems like we can only fix this by renaming `LCSHtmlDiffer` right?
What do you think about `TextDiff` and `CSVDiff`?

app/helpers/renderers/lcs_csv_differ.rb
inteon: a simplified table is shown instead (much like the text table)
This table does no diffing and combines all column content (newlines instead of rows)

app/helpers/renderers/lcs_csv_differ.rb
inteon: The interface would be simple:
```
.new(generated, expected)
.unified
.split
.render_accepted(builder, generated)
```

app/helpers/renderers/lcs_csv_differ.rb
inteon: fixed this using lambda function

app/helpers/renderers/lcs_csv_differ.rb
bmesuere: But why use a class method and not an instance method as with the rest of the methods in this file?

app/helpers/renderers/feedback_table_renderer.rb
bmesuere: @chvp What do you think? Rename to `LineDiffer`/`TextDiffer` and `CsvDiffer`?

app/helpers/renderers/lcs_csv_differ.rb
bmesuere: This is not a common construct in our code base and unless I'm missing something has no advantages over using a real method. In fact, in the face of testability a separate method is preferred.

app/helpers/renderers/feedback_table_renderer.rb
chvp: `TextDiffer` and `CsvDiffer` seem fine to me.

app/helpers/renderers/lcs_csv_differ.rb
inteon: because in case of an accept, there will be no instance created of the class (the constructor runs code specific to a non-accepted situation)

app/helpers/renderers/diff_csv.rb
```diff
@@ -0,0 +1,310 @@
...
+
+  def unified_simple(builder)
+    gen_cols = CSV.parse(@generated).transpose.map { |col| col.join("\n") }
+    builder.tr do
+      builder.td(class: 'line-nr') do
+        builder << (1..@generated_linecount).to_a.join("\n")
+      end
+      builder.td(class: 'line-nr')
+
+      builder << Array.new(@combined_headers.length) { |i| @gen_header_indices.index(i) }.map { |idx| idx.nil? ? '<td></td>' : %(<td class="del">#{CGI.escape_html gen_cols[idx]}</td>) }.join
```
chvp, line 183: This can still create a **lot** of HTML elements. Too many HTML elements on a page can noticeably slow down browsers. The text differ takes special care to do this with only a few elements. This is probably harder here if you still want this feedback to look somewhat good, but we might have to abandon that in extreme cases.

app/helpers/renderers/diff_csv.rb
```diff
@@ -0,0 +1,310 @@
...
+        end
+      end
+    end.html_safe
+  end
+
+  def split
+    builder = Builder::XmlMarkup.new
+
+    builder.div do
+      build_table = lambda { |headers, is_old, cls, icon_cls, title|
```
chvp, line 135: Again, move this to a separate function.

app/helpers/renderers/diff_csv.rb
```diff
@@ -0,0 +1,310 @@
...
+              builder << "<th colspan='#{headers.length}'>#{title}</th>"
+            end
+            builder.tr do
+              builder << "<th class='line-nr'></th>"
+              builder << headers.join
+            end
+          end
+          builder.tbody do
+            if @simplified_table
+              method(is_old ? :simple_old_row : :simple_new_row).call(builder)
```
chvp, line 153: Just use an if statement instead of adding extra indirection. (Same below.)

app/helpers/renderers/diff_csv.rb
chvp: This is also a pretty long line. Use `.<call> do |arg| ... end` syntax to split this over multiple lines.

chvp (changes requested):

app/helpers/renderers/diff_csv.rb
```diff
@@ -0,0 +1,310 @@
...
+              builder << (CSV.parse_line(line || '') || Array.new(gen_headers.length) { '' }).map { |el| %(<td>#{CGI.escape_html el || ''}</td>) }.join
+            end
+          end
+        end
+      end
+    end
+  end
+
+  def initialize(generated, expected)
+    extract_headers = lambda { |text|
```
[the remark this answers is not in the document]
inteon: I'm not sure if we can make Diff::LCS.sdiff work on a table, now I use the function on the lines first and then parse and apply sdiff on the columns.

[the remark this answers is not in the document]
chvp: I'm pretty sure the LCS differ should be able to handle that.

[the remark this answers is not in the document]
chvp: See for example this output:
```
irb(main):003:0> Diff::LCS.sdiff [[:a, :b], [:d, :e]], [[:a, :b, :c], [:d, :e]]
=> [#<Diff::LCS::ContextChange: ["!", [0, [:a, :b]], [0, [:a, :b, :c]]]>, #<Diff::LCS::ContextChange: ["=", [1, [:d, :e]], [1, [:d, :e]]]>]
```

app/helpers/renderers/diff_csv.rb
inteon: It should 'only' create 1 element per column. Will first try some other changes to improve performance.

app/helpers/renderers/diff_csv.rb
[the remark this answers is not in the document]
inteon: ok, cool; I'll try that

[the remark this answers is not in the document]
inteon: Seems to work, thank you for the improvement!

inteon:
I updated the code as requested, below are the timing results of some requests on my local computer:
```
CsvDiffer - 32 cols 1000 rows
501.2ms | Allocations: 608735
486.6ms | Allocations: 556962
526.4ms | Allocations: 556948
399.3ms | Allocations: 556942
649.3ms | Allocations: 556941
424.6ms | Allocations: 556947
502.2ms | Allocations: 556945
405.6ms | Allocations: 556945

TextDiffer - 32 cols 1000 rows
241.5ms | Allocations: 143148
174.0ms | Allocations: 94615
283.1ms | Allocations: 94610
376.1ms | Allocations: 94607
176.9ms | Allocations: 94608
161.0ms | Allocations: 94606
  
CsvDiffer - 32 cols 100 rows
111.8ms | Allocations: 145770
63.4ms | Allocations: 96886
74.9ms | Allocations: 96882
76.6ms | Allocations: 96877
40.7ms | Allocations: 96877
49.2ms | Allocations: 96877
40.5ms | Allocations: 96881

TextDiffer - 32 cols 100 rows
69.2ms | Allocations: 66493
19.7ms | Allocations: 18007
23.3ms | Allocations: 17996
53.6ms | Allocations: 18000
19.9ms | Allocations: 17993
24.2ms | Allocations: 18000
31.3ms | Allocations: 17996
18.2ms | Allocations: 18000
18.0ms | Allocations: 17993

CsvDiffer - 100 cols 100 rows
135.4ms | Allocations: 301609
191.0ms | Allocations: 252706
204.2ms | Allocations: 252698
91.0ms | Allocations: 252708
85.4ms | Allocations: 252710
85.7ms | Allocations: 252717
96.8ms | Allocations: 252724

TextDiffer - 100 cols 100 rows
73.6ms | Allocations: 72305
21.8ms | Allocations: 23818
36.3ms | Allocations: 23804
24.5ms | Allocations: 23811
25.8ms | Allocations: 23804
38.0ms | Allocations: 23814
23.8ms | Allocations: 23811
22.6ms | Allocations: 23811
```
It seems like the CsvDiffer is slower than the TextDiffer; it also does a lot more allocations (I guess this is due to the extra csv-decode step).

chvp:
Is this always one table or multiple?

inteon:
> Is this always one table or multiple?

This is 1 big table (as described eg. 1000 rows and 32 cols) and 1 small table (probably negligible).
The output was initially generated by the sql judge eg.:
image
image

bmesuere:
Can you retest with multiple tables/tests in the same result? We have no control over how many tests a teacher might include in their exercise (we recently had someone using 5000 tests).

The timing is for the entire feedback set. Part of that is a fixed cost, part of it will scale with the number of tests. It would be interesting to see how it scales.

inteon:
I added an extra constraint, now the constraints are:
- if more than 20 columns > use text rendering instead of csv rendering
- if more than 100 rows > use simplified csv rending

This results in the following benchmarks:
```
ALL: 6 chars per field, 50 tables (rows x columns)

100 x 100 -> (text fallback)
Duration: 635.1ms | Allocations: 299391
Duration: 577.8ms | Allocations: 251725
Duration: 602.0ms | Allocations: 251720
Duration: 553.1ms | Allocations: 251727
500 x 20  -> (simplified view)
Duration: 1214.1ms | Allocations: 1715646
Duration: 1051.4ms | Allocations: 1668200
Duration: 1187.4ms | Allocations: 1668202
Duration: 1200.1ms | Allocations: 1668200
Duration: 1237.9ms | Allocations: 1668197
100 x 20  -> (no fallback)
Duration: 1405.0ms | Allocations: 3819983
Duration: 1185.6ms | Allocations: 3773643
Duration: 1383.5ms | Allocations: 3772518
Duration: 1473.9ms | Allocations: 3826372
Duration: 1253.0ms | Allocations: 3772573
1000 x 3    -> (simplified)
Duration: 609.9ms | Allocations: 1368229
Duration: 788.6ms | Allocations: 1320804
Duration: 678.4ms | Allocations: 1320797
Duration: 658.1ms | Allocations: 1320805
3    x 1000 -> (text fallback)
Duration: 195.4ms | Allocations: 152149
Duration: 233.6ms | Allocations: 104550
Duration: 132.3ms | Allocations: 104543
Duration: 144.0ms | Allocations: 104539
```

app/helpers/renderers/csv_differ.rb
```diff
@@ -0,0 +1,337 @@
...
+              builder << line.map { |el| %(<td>#{CGI.escape_html el}</td>) }.join
+            end
+          end
+        end
+      end
+    end
+  end
+
+  def self.usable?(raw)
+    first_line, = raw.lstrip.split("\n", 2)
```
bmesuere, line 35:
```suggestion
    first_line = raw.lstrip.lines.first
```

app/helpers/renderers/csv_differ.rb
```diff
@@ -0,0 +1,337 @@
...
+              builder.th(idx + 1, class: 'line-nr')
+              builder << line.map { |el| %(<td>#{CGI.escape_html el}</td>) }.join
+            end
+          end
+        end
+      end
+    end
+  end
+
+  def self.usable?(raw)
```
bmesuere, line 34:
```suggestion
  def self.limited_columns?(raw)
```

app/helpers/renderers/feedback_table_renderer.rb
```diff
@@ -255,9 +255,17 @@ def messages(msgs)
     end
   end
 
+  def differ(t)
+    if t[:format] == 'csv' && CsvDiffer.usable?(t[:generated]) && CsvDiffer.usable?(t[:expected])
```
bmesuere, line 259:
```suggestion
    if t[:format] == 'csv' && CsvDiffer.limited_columns?(t[:generated]) && CsvDiffer.limited_columns?(t[:expected])
```

app/helpers/renderers/csv_differ.rb
```diff
@@ -0,0 +1,337 @@
...
+          end
+          builder.th(class: 'line-nr', title: I18n.t('submissions.show.expected')) do
+            builder.i(class: 'mdi mdi-18 mdi-file-check')
+          end
+          builder.th(colspan: @combined_headers.length)
+        end
+        builder.tr do
+          builder.th(class: 'line-nr')
+          builder.th(class: 'line-nr')
+          builder << @combined_headers.join
```
bmesuere, line 91: I think this is user input and needs to be escaped?

app/helpers/renderers/csv_differ.rb
```diff
@@ -0,0 +1,337 @@
...
+      builder.td(class: 'line-nr') do
+        builder << (1..@expected_rowcount).to_a.join("\n")
+      end
+      exp_cols.each do |col|
+        builder.td(class: 'ins') do
+          builder << CGI.escape_html(col)
+        end
+      end
+    end
+  end
```
bmesuere, line 252: These functions look very similar, can they be combined?

app/helpers/renderers/csv_differ.rb
```diff
@@ -0,0 +1,337 @@
...
+
+    builder.div do
+      split_build_table(builder, @gen_headers, true)
+      split_build_table(builder, @exp_headers, false)
+    end.html_safe
+  end
+
+  private
+
+  def split_build_table(builder, headers, is_old)
```
bmesuere, line 139: what is the meaning of `is_old`? Is this really the best name?

bmesuere (changes requested):
The code is already much improved 👍 

I left some comments, but as a general remark I am still missing more comments. For each of the functions, it should be clear what the function does and what the format of the input and output is.

I also took the liberty to rewrite the start post of the pull request a bit.

app/helpers/renderers/csv_differ.rb
inteon: @combined_headers is escaped already:
https://github.com/dodona-edu/dodona/pull/3065/files#diff-92f0d0cfac0130a6a7673e9ce05d27033bebb84e9428a7aa354a74e3430c46f9R294-R337

app/helpers/renderers/csv_differ.rb
inteon: yes, done

app/helpers/renderers/csv_differ.rb
inteon: changed it to `is_generated_output`

bmesuere (approved):

chvp (approved):

PR #3078, opened 2021-09-06
Make grading debounce independent per score item

### LINKED ISSUE #3072: Only last score item is updated when entering scores too fast
When grading exercises, it often occurs that a student only solves the first few steps of an exercise. In this case, I often just fill in all other score items with a zero by clicking the first box and then pressing 0, TAB, 0, TAB, 0, TAB, etc. However, if I do this too fast, only the last score item is updated and the loading indicator is never hidden. No errors are shown in the console, but when I take a look in the network tab of the browser's developer tools I see that only one request is performed.

I double checked and the entered scores are not saved and need to be entered again.

niknetniko:
This pull request fixes #3072.

The existing `delay` function uses a global timer for the delaying, regardless of what function you call. This is why only the last score was being saved (since the previous calls are cancelled when the next score saving function is called). I have now introduced a `createDelayer` function, which allows you to create multiple `delay`s that are independent.

- [x] Tests were added, and the test file was cleaned up a bit.

### DIFF
--- a/app/assets/javascripts/util.js
+++ b/app/assets/javascripts/util.js
@@ -26,13 +26,11 @@ import { isInIframe } from "iframe";
  *  @return {function(TimerHandler, number): void}
  */
 function createDelayer() {
-    return (function () {
-        let timer = 0;
-        return function (callback, ms) {
-            clearTimeout(timer);
-            timer = setTimeout(callback, ms);
-        };
-    })();
+    let timer = 0;
+    return function (callback, ms) {
+        clearTimeout(timer);
+        timer = setTimeout(callback, ms);
+    };
 }
 
 /*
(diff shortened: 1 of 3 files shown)


chvp (approved):

app/assets/javascripts/util.js
```diff
@@ -2,17 +2,47 @@
...
+ */
+function createDelayer() {
+    return (function () {
+        let timer = 0;
+        return function (callback, ms) {
+            clearTimeout(timer);
+            timer = setTimeout(callback, ms);
+        };
+    })();
+}
```
chvp, line 36:
```suggestion
function createDelayer() {
        let timer = 0;
        return function (callback, ms) {
            clearTimeout(timer);
            timer = setTimeout(callback, ms);
        };
}
```
Actually, isn't this equivalent?

niknetniko: Ah yes it is, good catch

bmesuere (approved):
👍

PR #3079, opened 2021-09-09
Implement OIDC identity provider

### LINKED ISSUE #3022: Add support for sign ins from vlaanderen.be
We will use OpenID Connect for this with our existing omniauth plugin. A button should be added on the sign in page but since Vlaanderen is no school, we will need to change a few things.

In Dutch the "selecteer je onderwijsinstelling" should become "selecteer je instelling". We could add an "Other" category to the bottom of the page (below the autocomplete) with Vlaanderen.be and move Elixir.

thepieterdc:
This pull request implements OIDC as an identity provider. The first use case is the Flemish government, documentation is available at https://authenticatie.vlaanderen.be/docs/beveiligen-van-toepassingen/integratie-methoden/oidc/technische-info/discovery-url/.

:exclamation: If you receive an `OpenIDConnect::Discovery::DiscoveryFailed` exception, this can be because of a present or absent trailing slash (`/`) in the issuer value. Some issuers include this, others don't. 

## Setup test
Add a new provider with the following settings:
```
irb(main):001:0> Provider::Oidc.create!(client_id: 'the-client-id', issuer: 'https://authenticatie-ti.vlaanderen.be/op')
```

## Setup production
Add a new provider with the following settings:
```
irb(main):001:0> Provider::Oidc.create!(client_id: 'the-client-id', issuer: 'https://authenticatie.vlaanderen.be/op')
```

## Displaying a provider in the frontend
Similar to `LTI` and `SAML`, the `provider` must be passed along in the request:
```ruby
 <%= link_to omniauth_authorize_path(:user, Provider::Oidc.sym, provider: provider_comes_here) do %>
Sign in!
<% end %>
```

## Manual testing
To test this, create a provider with the following settings. This issuer has whitelisted the callbacks for both `dodona.localhost:3000`, `mestra.ugent.be`, `naos.ugent.be` and `dodona.ugent.be`.
```ruby
irb(main):001:0> Provider::Oidc.create!(client_id: '4v22yzEh92pcRylWfEsXe1wtMlgQgE9x', issuer: 'https://dodona-oidc-test.eu.auth0.com/')
```

## Integration tests
This flow has been tested extensively using integration tests, the code coverage is more than 90%.
image

Related to #3022


lib/OIDC/auth/settings.rb
```diff
@@ -0,0 +1,31 @@
...
+    module Settings
+      def base_settings
+        # Support only third-parties that are discoverable.
+        {
+          client_options: {
+            redirect_uri: "https://#{Rails.configuration.default_host}/users/auth/oidc/callback"
+          },
+          discovery: true,
+          response_mode: :form_post,
+          response_type: :id_token,
```
bmesuere, line 14:
```suggestion
          response_type: :code,
```

lib/OIDC/auth/settings.rb
```diff
@@ -0,0 +1,31 @@
+require 'omniauth'
+
+module OIDC
+  module Auth
+    module Settings
+      def base_settings
+        # Support only third-parties that are discoverable.
+        {
+          client_options: {
+            redirect_uri: "https://#{Rails.configuration.default_host}/users/auth/oidc/callback"
```
bmesuere, line 10:
```suggestion
            redirect_uri: "https://#{Rails.configuration.default_host}/users/auth/vla/callback"
```

lib/LTI/jwk.rb
```diff
@@ -19,13 +19,13 @@ def keyset
...
       # Download the jwks keyset from the provider.
-      keys = JSON.parse(get_jwsk_content(uri)).with_indifferent_access
+      keys = JSON.parse(get_jwks_content(uri)).with_indifferent_access
 
       # Parse the keys.
       JSON::JWK::Set.new keys[:keys]
     end
 
-    def get_jwsk_content(uri)
+    def get_jwks_content(uri)
```
bmesuere, line 28: I'm not sure why, but by fixing this typo, one of the tests fails: 
```
      SocketError: getaddrinfo: Name or service not known (stark.name:80)
          lib/LTI/jwk.rb:29:in `get_jwks_content'
          lib/LTI/jwk.rb:22:in `parse_jwks_uri'
          lib/LTI/messages.rb:14:in `parse_message'
          app/controllers/concerns/set_lti_message.rb:11:in `set_lti_message'
      # lib/LTI/jwk.rb:29
      # lib/LTI/jwk.rb:22
      # lib/LTI/messages.rb:14
      # app/controllers/concerns/set_lti_message.rb:11
      # test/controllers/lti_controller_test.rb:24
```

This error is because we try to fetch key urls from an auto-generated address which of course fails. This can probably be fixed by stubbing the call, but I don't know why it passed in the first place.

bmesuere: I fixed this. The reason was that this call was in fact stubbed, but only for the wrongly spelled name.

bmesuere (changes requested):
When the callback url is triggered, we to use the correct settings to proceed. For this, we need to know which provider to use. A mechanism was implemented that can extract the provider id from the returned state, however it doesn't seem to be correctly set in the first place.

thepieterdc:
> When the callback url is triggered, we to use the correct settings to proceed. For this, we need to know which provider to use. A mechanism was implemented that can extract the provider id from the returned state, however it doesn't seem to be correctly set in the first place.

Should be fixed in https://github.com/dodona-edu/dodona/pull/3079/commits/68a1a4d68790e237a983ecbccde5946e85848eb8

lib/OIDC/auth/setup.rb
```diff
@@ -0,0 +1,72 @@
...
+        def provider_from_state
+          # If there is no state, we will not find anything.
+          return nil if params[:state].blank?
+
+          # Attempt to split the state.
+          state = params[:state].split("-", 2)
+          return nil unless state.count == 2
+
+          # Attempt to parse the id as an integer.
+          parsed = state[1].to_i
```
bmesuere, line 66:
```suggestion
          parsed = state[0].to_i
```

lib/OIDC/auth/strategy.rb
```diff
@@ -0,0 +1,31 @@
...
+
+      private
+
+      def user_info
+        return @user_info if @user_info
+
+        # Set the email address alias.
+        decoded = decode_id_token(access_token.id_token).raw_attributes
+        decoded["email"] = decoded["vo_email"]
+        @user_info = ::OpenIDConnect::ResponseObject::UserInfo.new(decoded)
```
chvp, line 25: This is specific to the Vlaamse Overheid, right? I feel like that should at least be noted in a comment that this will have to be extracted/changed when we add another OIDC provider.

chvp (commented):
Code looks good, thanks a lot :+1: 

(I obviously wasn't able to test the functionality.)

lib/OIDC/auth/strategy.rb
thepieterdc: It is indeed. Will do so

lib/OIDC/auth/settings.rb
```diff
@@ -0,0 +1,51 @@
...
+          },
+          discovery: true,
+          response_mode: :form_post,
+          response_type: :code,
+          scope: [:openid, :profile]
```
bmesuere, line 17: Don't we also need the `vo` scope for things like email?

lib/OIDC/auth/settings.rb
```diff
@@ -0,0 +1,51 @@
...
+        raise 'Not an OIDC provider.' unless provider.is_a?(Provider::Oidc)
+
+        {
+          client_auth_method: :jwt_bearer,
+          client_options: {
+            identifier: provider.client_id,
+            private_key: private_key
+          },
+          issuer: provider.issuer,
+          state: lambda { format("%d-%s", provider.id, SecureRandom::hex(16)) }
```
bmesuere, line 31: Add comment about encoding the provider id in the state to be able to retrieve it later at callback time.

lib/OIDC/auth/strategy.rb
```diff
@@ -0,0 +1,31 @@
...
+# This strategy enables OIDC to be used by Dodona.
+module OmniAuth
+  module Strategies
+    class OIDC < OmniAuth::Strategies::OpenIDConnect
+      include Rails.application.routes.url_helpers
+
+      option :name, 'oidc'
+
+      def client
+        @client ||= ::OIDC::Client.new(client_options.merge(audience: options.issuer))
```
bmesuere, line 14: In the end, was the audience returned by VO according to the spec or did you make changes to it? If yes, can you add a comment explaining what was the default and what you changed?

lib/OIDC/auth/strategy.rb
```diff
@@ -0,0 +1,31 @@
...
+      end
+
+      private
+
+      def user_info
+        return @user_info if @user_info
+
+        # Set the email address alias. This is specific to Vlaamse Overheid.
+        decoded = decode_id_token(access_token.id_token).raw_attributes
+        decoded["email"] = decoded["vo_email"]
```
bmesuere, line 24: Related to my previous question, is the vo_email field present also if we don't request the scope?

lib/OIDC/client.rb
```diff
@@ -0,0 +1,35 @@
+require "json/jwt"
+require "rack/oauth2/client/grant/authorization_code"
+
+module OIDC
+  class Client < OpenIDConnect::Client
+    attr_optional :audience
+
+    def authorization_code=(code)
```
bmesuere, line 8: This was monkey patched from the library? If yes, add a comment about why + changes.

bmesuere (changes requested):

lib/OIDC/auth/settings.rb
thepieterdc: Indeed, done :)

lib/OIDC/auth/strategy.rb
thepieterdc: It is not, we will need to change this when implementing other OIDC providers. My proposal is to not abstract this logic away just yet, but wait until we in fact require additional OIDC providers. This way we can extract everything Vlaamse Overheid-specific at once.

lib/OIDC/client.rb
thepieterdc: Not really monkey patched (as in, a change to the underlying library should/will not affect us), just a different client was injected.

bmesuere (approved):
👍 great work!

PR #3080, opened 2021-09-09
Make evaluation creation a full stepper

### LINKED ISSUE #2525: Use stepper component in evaluation wizard
image

This is the first screen a user gets when starting an evaluation and this might be optimized a bit. A user has to set a deadline here and indicate if he wants to enable grades. The rest is just extra information (eg, the series overview) but this isn't super clear.

- include something about grades in the info box
- make it clearer that the user has to pick a deadline and can enable grades. Some ideas which may or may not work:
  - slightly change the wording, eg Deadline -> Pick a deadline
  - add a "Settings" header to separate the top part of the page from the bottom part
- Add some text to the top of the series overview that makes it clear that the listing of exercises is just informational 
- Maybe also tweak the card title to make it more clear that we started some sort of wizard, eg "Configuring a series evaluation"

_Originally posted by @bmesuere in https://github.com/dodona-edu/dodona/issues/2462#issuecomment-785808920_

_The score-related options have been removed from this page, but we can still check if we can apply some of the other suggestions_.

### LINKED ISSUE #3037: Issues with manual selection of students for evaluation
I was making a selection of the students to be included in an evaluation (during the process of creating the evaluation). After selecting all students that submitted at least one solution, I manually added some additional students. I found these two issues:

- manual selection of students did not properly update the number of students selected (as mentioned on top of the page)
- in a long list of students, there's no easy way to see which students have been selected (e.g. by filtering the list to only those students that have been selected)

chvp:
Also change the edit page to lead back to a stepper.

Closes #2525, closes #3037.


bmesuere (changes requested):
:+1: 

The code looks good. Functionally, I have the following remarks:
- [x] Step 1 is about selecting a deadline but this is not super clear. Maybe rename this to pick/choose deadline? Can we add the overview and callout to a card above step 1? The info below the deadline picker can then be promoted to a callout to the right of the input.
- [x] If you go to step 2, the label for step 3 is already visible, but if you are at step 1, the labels for 2 and 3 aren't shown yet.
- [x] The list of users in step 2 feels long, especially because you need to scroll to the bottom to click next. Maybe reduce the number of users on a page to 15 or 20?
- [x] There's a lot of white space above the filter bar. Maybe add a h-border between the two sections would help
- [x] If I clicked the right preset, users in the table were selected, but the search bar progress kept spinning. The console showed an error:
```
Uncaught (in promise) TypeError: window.dodona.initCheckboxes is not a function
    at eval (eval at <anonymous> (index.js:101), <anonymous>:3:15)
    at index.js:101
```
-  [x] After selecting users, the top of the page reads "Je hebt op dit moment x gebruikers geselecteerd. Alle gebruikers verwijderen". Maybe use deselect instead of remove?
- It would be nice to be able to quickly see all selected users but this might be for a future PR
- [x] When a step is done, the chosen value isn't shown next to the step title (similar to how we do this for the course copy). It could list the chosen deadline and the number of users selected.
- [x] The page has no title, but this could be fixed with my first remark

chvp:
Also fixes #3037 now.

niknetniko (commented):
Looks good, some minor stuff I've noticed:

- When step 2 & 3 are not clickable, the mouse pointer still is the little hand, like when hovering on links
- When adding a score item to all exercises, the max score in the description isn't updated.
- When clicking on step 1, the selection in step 2 & 3 are cleared, while when clicking step 2, step 3 isn't cleared. I'm not sure if this is intentional or not.
- If you don't select any users and click though, you go to the edit page, which has step 2 & 3. I don't think it's 100% clear why (perhaps a message "you need to select users" or something would fix this)

bmesuere (approved):
Looks really good!

On the edit page, it is not super clear that all edits are saved directly. Especially because there is a "to evaluation" button at the bottom. Maybe add a message somewhere in the edit view like "all changes are saved immediately" and tweak the button text to "Back to evaluation"?

niknetniko (approved):

Issue #2195, opened 2020-08-29
New activity: multiple choice


pdawyndt:
The R courses now abuse "exercises" as a poor man's version of multiple choice questions.

![image](https://user-images.githubusercontent.com/5736113/91634478-0ba56400-e9f1-11ea-97d2-47bb6bfa6770.png)

This has opened the debate whether we should support a new activity type for multiple choice and to what extent that support should lead us (mockup based on [this snippet](https://bootsnipp.com/snippets/DVv0a))

![image](https://user-images.githubusercontent.com/5736113/91634625-f977f580-e9f1-11ea-90f2-e5dc401f3619.png)

This is a stub to collect good practices and ideas for multiple choice support.

pdawyndt:
[QTI](https://en.wikipedia.org/wiki/QTI) seems to be the leading data exchange format for multiple choice questions. [GIFT](https://en.wikipedia.org/wiki/GIFT_(file_format)) is an exchange format associated with Moodle. [SCORM](https://en.wikipedia.org/wiki/Sharable_Content_Object_Reference_Model) is much broader in scope.

bmesuere:
It will be very important to strictly scope this before implementing.
- Which format will we use to store the question, answers and feedback? QTI is a standard but this is very verbose xml which is not user editable without a custom editor.
- Will we allow markdown/html for the questions/answers/feedback? Or restrict it to inline elements for answers?
- Will all questions have a single correct answer or will we allow multiple?
- Will we give feedback to the user after answering?
- Will we allow a user to answer a question multiple times until they get it right? And will we store all answers?
- If we make those last 2 things configurable (e.g. for tests) it probably will not be enough to configure this in question itself because a certain question might be used in multiple courses with different feedback requirements. Those settings should thus be associated with the use of a question inside a series (maybe a setting at series level?). How do we prevent that a student looks up the question in the general listing and hands in multiple times there to retrieve the answer?

I suggest we go for a minimal implementation that mimics the current use of multiple choice questions but with a nicer user interface. We don't want to give the impression that Dodona is suitable for multiple choice exams. It was not built for that and the options to shield questions/answers are inadequate. The multiple choice implementation is merely meant as a study aid. So,
- Only one correct answer per question
- Always give feedback
- Users always can hand in multiple times but we store each result

pdawyndt:
[...]

PR #3088, opened 2021-09-10
Add overview for activity read states

### LINKED ISSUE #2385: Improve inspection of "marked as read"
Inspection when (date) students have marked a reading activity as read is quite limited:

- [ ] no listing available when students have marked as read, as we have list of submissions for an exercise (linked to the exercise)
- [ ] no link to the "digitally signed" document in places where we link to submission for exercises: series, status overview (series), ...

chvp:
- [ ] Tests were added

Closes #2385.


bmesuere (approved):
Functionally, this looks good. If this would be used by many people, I would opt for adding two tabs to the existing page instead of the icon, bit it's ok for now.

app/controllers/activity_read_states_controller.rb
```diff
@@ -0,0 +1,91 @@
...
+        format.json { head :ok }
+      end
+    else
+      render json: { status: 'failed', errors: read_state.errors }, status: :unprocessable_entity
+    end
+  end
+
+  private
+
+  def set_activity_read_states
```
niknetniko, line 60: Perhaps a comment here would be useful for future programmers that this logic should be very similar to the one used by the submission controller (and the reverse comment as well); e.g. if a bug is found here, it will also probably be in the submission logic and vice versa.

app/controllers/activity_read_states_controller.rb
```diff
@@ -0,0 +1,91 @@
...
+    @read_states = apply_scopes(policy_scope(ActivityReadState))
+
+    if params[:user_id]
```
niknetniko, line 63: Perhaps we can include some simple tests for the lines that aren't covered, e.g. viewing with a user id. No clue how much work this would be, so if it is too much, ignore this comment.

niknetniko (changes requested):
A bug I found:
- When going to [Your Account] -> My Submissions and then clicking the read activity button, the page loses for which user information is being shown.

I also agree with Bart; I think we should create an issue to not forget to eventually evaluate if this needs to be a tab instead.

app/controllers/activity_read_states_controller.rb
jorg-vr: I have added a comment, but your statement is relevant for most code in this pr. I started most code by duplicating submissions pages and making changes where relevant.

When we introduce more activity types such as multiple choice (as in #2195 ) we should probably look into using abstraction or mixins to solve this code duplication

app/controllers/activity_read_states_controller.rb
jorg-vr: I have added some simple tests which improve code coverage

niknetniko (approved):

PR #3125, opened 2021-09-21
Rewrite series graph legends in HTML instead of SVG

### LINKED ISSUE #3019: Reimplement series graph legends in html
The bottom legends in the graphs from #2598 are implemented using svg. This causes problems with text reflow on narrow screens. It would be better to move these legends to a normal div that's positioned in the same place as the current legend.

tibdhond:
The visualisation legends have been rewritten into html (instead of svg). Several styling attributes have been moved to the stylesheet. I also used the opportunity to give the colour boxes rounded colours since I feel like it looks a bit better.

image

image

Closes #3019 .

### DIFF
--- a/app/assets/stylesheets/components/visualisations.css.scss
+++ b/app/assets/stylesheets/components/visualisations.css.scss
@@ -25,6 +25,7 @@
 
 .stats-container {
   display: flex;
+  flex-direction: column;
 }
 
 .graph_placeholder {
@@ -34,4 +35,29 @@
 .metric-container {
   fill: none;
   stroke-width: 2;
+}
+
+.legend {
+  display: flex;
+  flex-direction: row;
+  justify-content: center;
+  margin-top: -20px;
+
+  .legend-item {
+    margin-left: 8px;
+    margin-right: 8px;
+    display: flex;
+    flex-direction: row;
+    align-items: center;
+
+    .legend-box {
+      width: 15px;
+      height: 15px;
+      border-radius: 5px;
+    }
+
+    .legend-text {
+      margin-left: 4px;
+    }
+  } 
 }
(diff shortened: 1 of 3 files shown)


chvp (commented):
Can we not generate the legends in the template files now? Or would that make coloring the indicators too complex?

bmesuere (changes requested):
Text reflow with many items/narrow screens (one of the main reasons of this PR) isn't fixed.

image

The squares should always be square, only the labels should shrink.

tibdhond:
@chvp For the stacked status graph this is probably possible, but it'll probably be a lot harder to for the cumulative graph since the amount of legend items can vary.  I'm not sure if there's an easy/elegant fix for this using ruby/css.

bmesuere:
> @chvp For the stacked status graph this is probably possible, but it'll probably be a lot harder to for the cumulative graph since the amount of legend items can vary. I'm not sure if there's an easy/elegant fix for this using ruby/css.

I'm ok with doing this in JS

tibdhond:
A 'flex-wrap' seems to have fixed both the text and the square issue, but I also added a 'min-width' to the boxes for good measure.

app/assets/stylesheets/components/visualisations.css.scss
```diff
@@ -35,3 +36,29 @@
   fill: none;
   stroke-width: 2;
 }
+
+.legend {
+  display: flex;
+  flex-direction: row;
+  justify-content: center;
+  flex-wrap: wrap;
```
bmesuere, line 44:
```suggestion
  flex-wrap: wrap;
  overflow: hidden;
```

app/assets/stylesheets/components/visualisations.css.scss
```diff
@@ -35,3 +36,29 @@
...
+
+  .legend-item {
+    margin-left: 8px;
+    margin-right: 8px;
+    display: flex;
+    flex-direction: row;
+    align-items: center;
+
+    .legend-box {
+      min-width: 15px;
```
bmesuere, line 54:
```suggestion
```

bmesuere (approved):

Issue #925, opened 2019-08-01
Add visualisation of submission history


pdawyndt:
Related to #924, I think it would benefit both students and teachers if we combined

1. the listing of all submissions of a user for a particular exercise (in a particular course)
2. the page showing the details of a single submission

by adding iteration (like pagination) to the page showing the details of a single submission, so that one could iterate back and forth between successive submissions. Here's a quick mockup of some sort of submission timeline slider that could be used to jump to a particular submission

![image](https://github.ugent.be/storage/user/6/files/e103af00-4315-11e9-888b-ff08f43aeb87)

This would give a visual impression on the time the user spent on the exercise. And I could image additional features like

1. a `play` button that shows the evolution of the coding process by automatically moving from the first submission to the last submission
2. a double pointer on the submission timeline to compare (diff) two submissions
![image](https://github.ugent.be/storage/user/6/files/ed880780-4315-11e9-83da-e80830221a48)
3. a kind of gradient between submissions that indicates how "big" the change is between submissions (e.g. to detect big jumps in strategy to solve an exercise), could be done by diffing successive submissions and scoring the degree of change based on insertions/deletions
![image](https://github.ugent.be/storage/user/6/files/fc6eba00-4315-11e9-9925-89c416fe4beb)
4. an extra marker to indicate the deadline of an evaluation
5. more detailed statistics about the sequence of submissions (e.g. estimation of total time spent, taking into account time in between first and last submission that was spent on programming)

_[Original issue](https://github.ugent.be/dodona/dodona/issues/925) by @pdawyndt on Sun Mar 10 2019 at 09:19._

pdawyndt:
Here's a ][way of representing the history](https://twitter.com/amasad/status/1570488498205126656) as I somewhat had it in mind.

jorg-vr:
easier navigation already implemented, no need for the fancier stuff

PR #3136, opened 2021-09-29
Add info icon when question has newer submission

### LINKED ISSUE #2540: Display whether a student has submitted a correct answer after asking a question
Sometimes students will send in a question in Dodona (great feature!), but will have solved their issue at a later point. It would be really useful if we could see whether a correct submission has been made, OR see all of the students submissions for the question (correct and incorrect). The latter might make for a better judgment of whether the student has answered their own question, in case the student opted for an alternative solution.

chvp:
screenshot_2021-10-02-201451

Closes #2540 .


niknetniko:
Do we want to show this (with the current wording) to the students themselves?

pdawyndt:
Would be handy if the warning also links to the other submissions of the student for this exercise. Why not show the number of upstream and downstream submissions as we also do in evaluations?

chvp:
> Would be handy if the warning also links to the other submissions of the student for this exercise. Why not show the number of upstream and downstream submissions as we also do in evaluations?

There is no room for this in the annotation interface. It's also far too involved for the usecase (which is a quick check that the student hasn't fixed their own problem already).

> Do we want to show this (with the current wording) to the students themselves?

Right, forgot that students will also see this when picking the wording. Showing to students doesn't seem like a problem to me (might remind them to mark their own question as answered), but wording should obviously be changed to accommodate that. Problem is that we don't have easy access to the user type in the location where we pick the translation, so it's preferable to come up with something generic that works for both.

chvp:
I tweaked the wording so that it works for both students and teachers.

app/models/question.rb
```diff
@@ -37,6 +37,11 @@ def create_notification; end
 
   def destroy_notification; end
 
+  def newer_submission
+    # Submissions are sorted newest first by default
+    Submission.where(id: (id + 1).., exercise_id: submission.exercise_id, course_id: course_id, user_id: submission.user_id).first
```
niknetniko, line 42:
```suggestion
    Submission.where(id: (submission.id + 1).., exercise_id: submission.exercise_id, course_id: course_id, user_id: submission.user_id).first
```
If I understand correctly.

niknetniko (changes requested):

niknetniko (approved):

config/locales/js/en.yml
```diff
@@ -157,6 +157,7 @@ en:
       unresolve: 'Re-open question'
       deleted: This question was deleted by another user
       conflict: The status of this question was already changed by another user
+      has_newer_submission: There is a newer solution since this question. Click here to go to their newest submission.
```
bmesuere, line 160:
```suggestion
      has_newer_submission: There was a newer submission since this question. Click here to go to the latest submission.
```

bmesuere (changes requested):
- hasWarning and warningUrl are implemented in a generic way in the Annotation class (which is good), but the tooltip title is very specific. Is there a reason for this?
- Can we use the bootstrap (popper) tooltip instead of the title attribute (which can take a while to show up).
- I'm not sure about the size of the icon. There is a big contrast in size and color with the gray small text. Maybe mdi-12 is more appropriate here? Or a blue info icon because there is nothing actually wrong?
- Do we also want to show this warning for completed questions? Or only for open questions?

Regarding the remark of @pdawyndt: I have been thinking about something like this when I was working on the redesign. The idea was to show some kind of timeline at the top of the submissions since there we only use the left side of the page to show the status. However, I haven't come up with a good design yet.

chvp:
> * hasWarning and warningUrl are implemented in a generic way in the Annotation class (which is good), but the tooltip title is very specific. Is there a reason for this?

Other code in this file does this as well. We could get the warning text in the same way as the URL, but like for the other code it feels to me like a bridge we should cross once it's necessary.

> * Can we use the bootstrap (popper) tooltip instead of the title attribute (which can take a while to show up).

Sure, I'll change this.

> * I'm not sure about the size of the icon. There is a big contrast in size and color with the gray small text. Maybe mdi-12 is more appropriate here? Or a blue info icon because there is nothing actually wrong?

I'll change it to a blue info icon.

> * Do we also want to show this warning for completed questions? Or only for open questions?

Especially with the icon change it doesn't feel like it does any harm to show it for completed questions as well.

bmesuere (approved):
Looks good! Do you have a screenshot to update the PR start?

pdawyndt:
Some iDeal for a submission history in #925

PR #3166, opened 2021-10-14
Finalize the heatmap in the series visualization

### LINKED ISSUE #3017: Finalize the heatmap in the series visualization
A first version of this graph was implemented in #2598 but wasn't included because it needed a bit more work.
More specifically, the binning should be dynamic depending on the time domain of the graph. 

The improved binning could also be used for the line graph.

tibdhond:
This pull request hopes to finalize the timeseries graph. 
The graph now uses a custom binning system that can handle anywhere from bins per 5 minutes to bins per 4 weeks (and can easily be expanded).

imageimage

Closes #3017 .


tibdhond:
- The issue with the tooltip numbers has been fixed.
- The range detection was working as intended, the configuration just wasn't broad enough. It has been expanded to also look for bin steps of 5, 10 and 30 minutes.
- Can't reproduce the not enough data bug on Naos.
- There was an error in the aligning part causing the first bins to be empty. This has been fixed.
- A first attempt at displaying the bin size has been added. Screenshots at the top of the page have been updated.
- Datetime formatting has been improved

bmesuere (changes requested):
It seems that the functional issues are indeed fixed 👍 
Some additional suggestions:
- The legend now reads "1x[x] = 15 minutes". I would remove the "1x" and "=", and make the square white with only the outline border shown. If we don't do the below suggestion, it could probably be moved to the bottom left corner to the left of the tick marks.
- Would it be useful to add a legend mapping the color intensity to values? In don't think we really need this, so I wouldn't to this for now.
- The ticks are much improved!
- Tooltip suggestions:
  - Do we need the "van: " and "tot: "? A dash between the two values is shorter. If we keep them, they shouldn't be in bold but the values should.
  - For shorter intervals (e.g. 15 minutes), the day isn't mentioned. Suggestion: **Van:** 14:00 **Tot:** 14:15 -> **14:00 - 14:15** op woensdag 6 september.
  - 1 day: this has a "ingediende oplossingen op" prefix that's redundant
  - 1 week/2 weeks: **Maandag 2 januari - 16 januari**
  - Pluralization for the submissions e.g. "1 oplossingen"
-  Sometimes, submissions from (1 interval) after the deadline are taken into account. You can see this clearly on naos in course 3, week 12
- how does it look if we add a border (in a light-ish blue?) to the squares? The ones with low values are hard to see now in light mode on my screen

tibdhond:
Feedback has been processed, screenshots at the top of the page have been updated accordingly.

app/assets/javascripts/visualisations/timeseries.ts
```diff
@@ -71,9 +83,10 @@ export class TimeseriesGraph extends SeriesExerciseGraph {
             .style("z-index", 5);
 
         // make sure cell size isn't bigger than bandwidth
-        // this is commented for now awaiting more data
-        // const rectSize = Math.min(this.y.bandwidth()*1.5, this.innerWidth / this.dateRange - 5);
-        const rectSize = this.y.bandwidth();
+        const rectSize = Math.min(
+            this.y.bandwidth()*1.5,
```
chvp, line 87:
```suggestion
            this.y.bandwidth() * 1.5,
```

app/assets/javascripts/visualisations/timeseries.ts
```diff
@@ -51,17 +54,26 @@ export class TimeseriesGraph extends SeriesExerciseGraph {
...
-            .domain([this.minDate, end])
-            .range([5, this.innerWidth]);
+        this.x = d3.scaleBand()
+            .domain(this.binTicks)
+            .range([0, this.innerWidth]);
 
         // Axis
         this.graph.append("g")
-            .attr("transform", `translate(0, ${this.innerHeight})`)
+            .attr("transform", `translate(0, ${this.innerHeight-this.y.bandwidth()/2})`)
```
chvp, line 63:
```suggestion
            .attr("transform", `translate(0, ${this.innerHeight-this.y.bandwidth() / 2})`)
```

app/assets/javascripts/visualisations/timeseries.ts
```diff
@@ -99,6 +113,35 @@ export class TimeseriesGraph extends SeriesExerciseGraph {
                     .attr("height", rectSize)
                     .attr("fill", d => d.sum === 0 ? "" : this.color(d.sum));
             });
+        const unitTl = I18n.t("time.time_units");
```
chvp, line 116: Not clear what `Tl` means.

app/assets/javascripts/visualisations/timeseries.ts
```diff
@@ -177,7 +214,36 @@ export class TimeseriesGraph extends SeriesExerciseGraph {
...
             .style("opacity", .9);
-        let message = `<b>${this.longDateFormat(d.date)}</b><br><b>${d.sum} ${I18n.t("js.submissions")}</b>`;
+        let message = "";
+        if (this.binStep < 24) {
+            const on = I18n.t("js.date_on");
+            const timeFormat = d3.timeFormat(I18n.t("time.formats.plain_time"));
+            const dateFormat = d3.timeFormat(I18n.t("date.formats.weekday_long"));
+            message = `
+                <b>${timeFormat(d.date)} - ${timeFormat(new Date(d.date+this.binStep*3600000))}
+                <br>${on} ${dateFormat(d.date)}<br>
```
chvp, line 224: This makes days in Dutch capitalized, which should not be the case.

config/locales/defaults/en.yml
```diff
@@ -27,11 +27,19 @@ en:
     formats:
       weekday_short: "%a %b %d"
       weekday_long: "%A %B %d"
+      monthday_long: "%B %d"
   time:
+    time_units:
```
chvp, line 32:
```suggestion
    units:
```

chvp (changes requested):

app/assets/javascripts/visualisations/timeseries.ts
chvp: Just changing the casing in the translation strings is not good enough. In the other string the day is the first element of a sentence, so should be capitalised in that case.

chvp (changes requested):
Please add whitespace around all mathematical operators for consistency with our codebase. I only noted a few in the previous review, but there are others still.

tibdhond:
All of the spacing issues should have been fixed now.
The tooltip dates are now capitalized if they are the first word in the sentence.

bmesuere:
I would also reduce the size of the square in the legend some more to make it less prominent.

bmesuere (commented):
I would also reduce the size of the square in the legend some more to make it less prominent.

tibdhond:
The box in the legend is now shrunk by two thirds.
I was also wondering if using the status icons in the tooltip would give added value (in addition to the textualstatus )?

bmesuere:
> The box in the legend is now shrunk by two thirds (screenshots updated accordingly). I was also wondering if using the status icons in the tooltip would give added value (in addition to the textualstatus )?

This is hard to say without trying, so maybe create a mockup and compare screenshots of the 2 versions?

tibdhond:
Tooltip with and without the statusicons.
I personally think it looks a lot better with the icons, it breaks up the wall of text a bit.

image
image

config/locales/defaults/nl.yml
```diff
@@ -21,18 +21,26 @@
 
 nl:
   date:
-    day_names: ["Zondag", "Maandag", "Dinsdag", "Woensdag", "Donderdag", "Vrijdag", "Zaterdag"]
+    day_names: ["zondag", "maandag", "dinsdag", "woensdag", "donderdag", "vrijdag", "zaterdag"]
```
chvp, line 24: Is this used anywhere else? Because if so this will have an impact in those places as well, which we do not necessarily want.

chvp (commented):
Code looks good to me. I quite like the icons in the tooltip.

config/locales/defaults/nl.yml
tibdhond: A global search for `day_names` (in vscode) only returns hits in `series_graph.ts` and the translation files, so I would assume it's not used anywhere else.

chvp: Okay.

bmesuere (changes requested):
Sorry for the delay in reviewing. Right now, everything in the tooltip is in bold which defies the purpose. How about in the example below but the following things in bold:
- 23:35 - 24:40
- 51 oplossingen
- 17, 9, ... (so only the numbers for the detailed part)
image


I do like the icons!

tibdhond:
Everything in the tooltip being bold was a bug caused by a missing closing tag.
Nevertheless, changes have been made as per the suggestion:
image

bmesuere (approved):

chvp (approved):

PR #3203, opened 2021-11-13
Make email address on user profiles a mailto link

### LINKED ISSUE #3201: Make email addresses on user profiles and course member pages a mailto link
I don't see any problems with this, and it could be convenient. (Just found myself wishing for this on mobile.)

BTWS2:
image
image

Closes #3201.

### DIFF
--- a/app/views/course_members/_members_table.html.erb
+++ b/app/views/course_members/_members_table.html.erb
@@ -32,7 +32,7 @@
             <%= link_to user.full_name, course_member_path(@course, user), title: user.full_name, class: "ellipsis-overflow" %>
           <% end %>
         </td>
-        <td><%= user.email %></td>
+        <td><%= mail_to(user.email, nil, { title: t('.send_mail') }) %></td>
         <td><%= user.institution&.name %></td>
         <td class="d-none d-sm-table-cell">
           <%= render partial: 'application/progress_chart',
--- a/app/views/users/show.html.erb
+++ b/app/views/users/show.html.erb
@@ -30,7 +30,7 @@
             <% if @user.institution&.name.present? and @user.email.present? %>
               &middot;
             <% end %>
-            <%= @user.email %>
+            <%= mail_to(@user.email, nil, { class: 'text-muted', title: t('course_members.show.send_mail') }) %>
           </span>
           <p class="timezone">
             <%= t ".timezone" %> <b><%= @user.time_zone %></b>.
--- a/app/views/course_members/show.html.erb
+++ b/app/views/course_members/show.html.erb
@@ -32,7 +32,7 @@
             <% if @user.institution&.name.present? and @user.email.present? %>
               &middot;
             <% end %>
-            <%= @user.email %>
+            <%= mail_to(@user.email, nil, { title: t('.send_mail'), class: 'text-muted' }) %>
           </span>
         </div>
         <div class="row">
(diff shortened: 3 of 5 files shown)


bmesuere (approved):
Not sure if the tooltip is needed, but ok for me.

Thanks!

niknetniko (approved):

chvp (changes requested):
Like the issue asks, I would also add this to the user profile. For the Dodona admins, this can be useful as well.

app/views/course_members/_members_table.html.erb
```diff
@@ -32,7 +32,7 @@
             <%= link_to user.full_name, course_member_path(@course, user), title: user.full_name, class: "ellipsis-overflow" %>
           <% end %>
         </td>
-        <td><%= user.email %></td>
+        <td><span title="<%= t '.send_mail' %>"><%= mail_to(user.email) %><span></td>
```
chvp, line 35: The extra span isn't required. You can use the [`html_options` argument of `mail_to`](https://api.rubyonrails.org/classes/ActionView/Helpers/UrlHelper.html#method-i-mail_to) to set the title. `mail_to` also supports a `name` argument, which we can use as well.

chvp (changes requested):
Right now, the emailaddress is also colored as a link. Personally I think this is a bit too attention-grabbing.

app/views/users/show.html.erb
```diff
@@ -30,7 +30,7 @@
             <% if @user.institution&.name.present? and @user.email.present? %>
               &middot;
             <% end %>
-            <%= @user.email %>
+            <%= current_user.admin? ? mail_to(@user.email) : @user.email %>
```
chvp, line 33: `admin?` also includes staff. `zeus?` should be used.

bmesuere:
> Right now, the emailaddress is also colored as a link. Personally I think this is a bit too attention-grabbing, especially on the profile pages. @bmesuere opinions?

If it is a link, it should look like a link? I don't have a problem with the current styling because the profile page isn't used that much.

app/views/course_members/_members_table.html.erb
```diff
@@ -32,7 +32,7 @@
             <%= link_to user.full_name, course_member_path(@course, user), title: user.full_name, class: "ellipsis-overflow" %>
           <% end %>
         </td>
-        <td><%= user.email %></td>
+        <td><%= mail_to(user.email, nil, {title: t('.send_mail'), class: 'text-muted'}) %></td>
```
chvp, line 35: We can fill in this `nil` argument, we have the name of the user (here and in the other instances). I would also not add the text-muted class here since the link works visually in our tables.

chvp (changes requested):
Could you rebase on latest develop when making changes? That way the checks will run for your changes.

app/views/course_members/_members_table.html.erb
BTWS2: I replaced `nil` with the user email. I would like to continue displaying the e-mail, because not everyone has their mailto client configured properly. I removed text-muted.

BTWS2:
I had some trouble with the rebase because of multiple laptops, but it should be ok now.

app/views/course_members/_members_table.html.erb
chvp: I did not realize filling in the argument would change the shown content of the `a`. In that case, leaving it empty seems like the better option. I'll change this, approve and merge.

chvp (approved):

PR #3204, opened 2021-11-14
Allow users to set time bounds of series visualisations

tibdhond:
This pull request gives the user control over the time scope of the visualisation on the course page.

image


chvp (changes requested):
I was trying to test this on naos, but the filtering doesn't seem to work correctly. I tried to apply bounds to [this series](https://naos.ugent.be/en/courses/3/#series-264-second-evaluation-group-1) to zoom in on the day of the evaluation, but got a "no data" message (even though there should have been data, when looking at the submission list).

tibdhond:
The bug seems fixed, however the initialization of the date picker doesn't seem to be on point yet. This is mostly because it's hard to find a good date to initialize on (especially for the graphs that don't use the timestamps). For now I'll leave it like this until we decide whether to keep the time scope for the non-timestamp graphs.

bmesuere:
@tibdhond can you rebase this?

tibdhond:
I moved the time scoping the the frontend. This has several advantages, but one disadvantage:
\+ No need for sanitation
\+ Data fetching is easier (since the before parts doesn't have to be calculated, D3 easily takes care of that)
\+ Date pickers are easier to configure
\- Sometimes, more data than necessary is sent to the frontend

This raises two possible ways how we can use the time scoping:

- Keep things as they are and send a request to the backend every time 'apply' is pressed (and keep the data up to date)
- Apply the scoping on previously fetched data (but potentiality miss out on new submissions)

app/assets/javascripts/visualisations/series_graph.ts
```diff
@@ -62,20 +72,54 @@ export abstract class SeriesGraph {
...
+            };
+            if (I18n.locale === "nl") {
+                options.locale = Dutch;
+            }
+            this.fpStart = flatpickr(`#scope-start-${seriesId}`, options);
+            this.fpEnd = flatpickr(`#scope-end-${seriesId}`, options);
+            this.scopeApply = document.getElementById(`scope-apply-${seriesId}`);
+            this.scopeApply.onclick = () => this.applyScope();
+        }
+
```
chvp, line 91: This feels like it should be solved by inheritance instead of an `if` statement.

app/policies/series_policy.rb
```diff
@@ -110,7 +110,7 @@ def course_admin?
   end
 
   def statistics?
-    course_admin?
+    course_admin? || true
```
chvp, line 113:
```suggestion
    course_admin?
```

chvp (changes requested):

app/assets/javascripts/visualisations/series_graph.ts
tibdhond: While I do agree that this does seem like an inheritance thing, I feel like it would complicate things a bit. We would either have to:
- Duplicate everything inside the if to both the cumulative graph and heatmap (and any future graph that uses the datepickers)
- Use mixins (which might impact execution of 'super' methods/calls)
- Create a more complex hierarchy (something like: SeriesGraph > ExerciseGraph > TimeExerciseGraph > Heatmap + SeriesGraph > TimeGraph > Cumulative)
- Decouple the heatmap from its ExerciseGraph super class, copy any code inside that super class to the heatmap and make a common super class between the heatmap and the cumulative graph

tibdhond: Moved some things around. The initialization of the time pickers has now been put in a function which is called from the heatmap init and cumulative init.

bmesuere (changes requested):
In chrome, the pickers are full width instead of inline:
image

The ticks seem to show the end time of the interval? Is there a reason for this? The start seems more intuitive to me.

If I load the graph for for course 3, exam group 6 on naos and select 15h as starting time, the first block of the graph seems bugged:
image

tibdhond:
@bmesuere 
> In chrome, the pickers are full width instead of inline:

This is probably due to the style sheets not being compiled anymore (issue #3289). Not much I can do about this until this issue is fixed.

> The ticks seem to show the end time of the interval? Is there a reason for this? The start seems more intuitive to me.

I had changed this to make it consistent with the cumulative graph, but in hindsight this wasn't necessary. It has been switched back now.

> If I load the graph for for course 3, exam group 6 on naos and select 15h as starting time, the first block of the graph seems bugged:

The old data wasn't discarded on updates. This should be fixed now.

chvp (approved):
Code looks good. The styling of the pickers was indeed due to #3289 (clearing the cache and recompiling the assets on naos fixed it). Not sure about the UI (it's just two time pickers and a button, without ever making it clear what the pickers and button are for), but I'll let @bmesuere comment on that.

bmesuere (commented):
I agree, this probably looks better if you use the `calendar-start` and `calendar-end` icons. You could leave out the button and update automatically if any of the two fields are changed.

bmesuere (approved):

bmesuere:
can I merge this, @tibdhond ?

tibdhond:
Looks good to me.

PR #3207, opened 2021-11-15
Fix average dot of violin plot being placed outside of the graph when it is high

### LINKED ISSUE #3206: Average dot outside of graph
The average dot  of the 'Number of submissions per user' plot is placed outside the graph when the average is larger than 20.

image

tibdhond:
This pull request fixes the bug where the average dot of the violin graph is outside of the graph when the average is larger than 20.
image

Closes #3206 .


bmesuere (approved):

niknetniko (approved):

chvp (commented):
I have a slight concern that it won't be immediately obvious that the average is (in some cases much) higher than 20.

tibdhond:
> I have a slight concern that it won't be immediately obvious that the average is (in some cases much) higher than 20.

How about if we replace the dot with a simple arrow (something like '>') when the average is outside of the graph?

chvp:
Trying a different icon seems like a good idea. Can you post a screenshot if you try that? I would definitely mention the average in a tooltip if that is not already the case.

tibdhond:
I was thinking of something like this:

image

bmesuere (changes requested):
I'm not sure if I like the icon for 20+ (but can live with it).

I would change the tick mark on the axis from "20" to "20+", this will also be more correct for the violin plot itself.

bmesuere (approved):

chvp (approved):

PR #3215, opened 2021-11-20
Add copy to clipboard button for submission link

### LINKED ISSUE #2213: Direct link to individual submissions
Students find it hard to get access to the URL of one of their submissions (e.g. to ask for specific feedback on that submission). Something like this might solve it.

image

BTWS2:
Default:
image
On hover:
image

I chose this lay-out because the link refers to the submission and aligns with the other symbol.

Closes #2213.

### DIFF
--- a/app/views/submissions/_description.html.erb
+++ b/app/views/submissions/_description.html.erb
@@ -30,7 +30,11 @@
   </span>
   <div class="row">
     <div class="col-md-6">
-      <span class='status-icon float-start'><%= submission_status_icon(submission, 36) %></span>
+      <span class='status-icon float-start'>
+        <button class="btn btn-icon" data-clipboard-text="<%= submission.id %>" title="Copy submission link to clipboard.">
+          <%= submission_status_icon(submission, 36) %>
+        </button>
+      </span>
       <span class="status-line">
         <%= Submission.human_enum_name(:status, submission.status) %>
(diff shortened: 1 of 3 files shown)


app/views/submissions/_description.html.erb
```diff
@@ -30,7 +30,11 @@
   </span>
   <div class="row">
     <div class="col-md-6">
-      <span class='status-icon float-start'><%= submission_status_icon(submission, 36) %></span>
+      <span class='status-icon float-start'>
+        <button class="btn btn-icon" data-clipboard-text="<%= 'https://' + request.domain.to_s + submission_path(submission.id).to_s %>" title="<%= t "submissions.show.copy_submission_link" %>">
```
chvp, line 34:
```suggestion
        <button class="btn btn-icon" data-clipboard-text="<%= submission_url(submission.id) %>" title="<%= t "submissions.show.copy_submission_link" %>">
```

chvp (changes requested):
I'm not sure I agree with this placing. Even though there is a tooltip, this will not be obviously a button. Since this has been requested independently by a number of teachers, we need this to be a lot more discoverable, IMO.

bmesuere (commented):
I agree that we need a better solution for this than the current proposal. The entire result area must be taken into account. For now, I/we don't have time to design this.

bmesuere:
This was mostly solved by adding clickable and linkable submission numbers to the ui in the previous release.

PR #3222, opened 2021-11-24
Add dynamic binning to the cumulative graph

tibdhond:
This pull request adds dynamic binning to the cumulative graph.

image
image


chvp (changes requested):
There's some problems with the labels: 
screenshot_2021-12-01-141323

I also got an unexpected "no data" message for [this series](https://naos.ugent.be/en/courses/3/#series-286-exam-group-1).

tibdhond:
The issue with the tick labels was more of a design flaw than a functional one.
It happened when the data was binned per 12 hours. Because every other label was omitted, the strange pattern observed in the screenshot was created. This has now been fixed by using slanted labels instead of omitting every other label.

As for the second issue:
The "no data" message is displayed when the frontend receives an answer from the backend, but the data field is empty. This means this is a backend problem. I have checked the backend code, but can't seem to find where the fault lies. Since I don't have access to the data of that series, it's hard to pinpoint what could cause this bug.

chvp (changes requested):
Can you add some test cases that catch the no data edge case(s) we encountered?

bmesuere (commented):
Right now, the deploy fails on mestra (with a syntax error). Is this also the case with you locally?

Before I deployed, there still seemed to be an issue with series without (correct) submissions. If there were none, I correctly received the "no data available" message but there still seem to be edge cases where I got an empty graph with 1 time stamp on the x-axis and 0% in the middle of the y-axis.

tibdhond:
> Right now, the deploy fails on mestra (with a syntax error). Is this also the case with you locally?

I was not able to reproduce this error locally. I took a look at the mestra deploy logs, but it shows as succeeded for me, so I have no idea what might have caused this error.

> Before I deployed, there still seemed to be an issue with series without (correct) submissions. If there were none, I correctly received the "no data available" message but there still seem to be edge cases where I got an empty graph with 1 time stamp on the x-axis and 0% in the middle of the y-axis.

This was caused by a bug in the binning when there was only one bin. This was surprisingly hard to fix and has led to some changes in the workings of the graph:

- The tick labels now display the endpoint of a bin instead of the starting point (for the sake of consistency this has also been changed for the heatmap timeseries).
- The first bin in the cumulative graph now represent a 'before' section, containing any submissions not in the selected time range (also reflected in the tooltip as shown in the screenshot). At present this bin will always be empty, but should see some use once the time scope is being utilized.
image

bmesuere (changes requested):
The code looks ok, but I have found 2 small remaining issues with the ticks:

On the y-axis, you explicitly use the `.0%` format which causes an incorrect axis for small values. I think just `%` should do the trick.

image

With some hour-bins, the x-axis ticks aren't correctly formatted:
image

tibdhond:
Problems with both axes should now be resolved.

On the x-asis midnight tick labels have now been replaced with the date.
image

bmesuere (approved):

tibdhond:
Just noticed some problems might arise when the timezone switches (summertime/wintertime). Whenever this happens either 01:00 or 23:00 is now replaced.

image

chvp (approved):

bmesuere (approved):

PR #3274, opened 2022-01-05
Rights request: Show correct error message when motivation is not filled out

### LINKED ISSUE #3124: Error message when motivation is missing in rights request is very misleading
Instead of an actual error message, we tell the user that a rights request already exists, which is obviously wrong.

jorg-vr:
This pull request fixes a bug where the user was shown misleading information when he hadn't filled out a motivation message.

Right now the following message is shown:
image

The error message saying you already applied still appears when it should:
image

Closes #3124 .

### DIFF
--- a/app/views/rights_requests/new.html.erb
+++ b/app/views/rights_requests/new.html.erb
@@ -3,14 +3,14 @@
     <div class="card">
       <div class="card-title card-title-colored">
         <h2 class="card-title-text"><%= t ".title" %></h2>
-        <% if policy(RightsRequest).create? or current_user&.rights_request.errors.any?%>
+        <% if policy(RightsRequest).create? or @rights_request.errors.any? %>
           <div class="card-title-fab">
             <%= render 'fab_button', form: 'new_rights_request', icon: 'send' %>
           </div>
         <% end %>
       </div>
       <div class="card-supporting-text">
-        <% if policy(RightsRequest).create?  or current_user&.rights_request.errors.any?%>
+        <% if policy(RightsRequest).create?  or @rights_request.errors.any? %>
           <%= render 'form', rights_request: @rights_request %>
         <% else %>
           <div class="callout callout-info"><p>
--- a/app/views/rights_requests/new.html.erb
+++ b/app/views/rights_requests/new.html.erb
@@ -3,14 +3,14 @@
     <div class="card">
       <div class="card-title card-title-colored">
         <h2 class="card-title-text"><%= t ".title" %></h2>
-        <% if policy(RightsRequest).create? %>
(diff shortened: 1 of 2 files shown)


app/views/rights_requests/new.html.erb
```diff
@@ -3,14 +3,14 @@
     <div class="card">
       <div class="card-title card-title-colored">
         <h2 class="card-title-text"><%= t ".title" %></h2>
-        <% if policy(RightsRequest).create? %>
+        <% if policy(RightsRequest).create? or current_user&.rights_request.errors.any?%>
```
bmesuere, line 6: It seems strange get the rights_request from the current user. You should be able to access it through `@rights_request`. 

In addition, we always add a space before the `%>`. Our linter doesn't check for this in erb files :(

bmesuere (changes requested):

bmesuere (approved):

app/views/rights_requests/new.html.erb
```diff
@@ -3,14 +3,14 @@
     <div class="card">
       <div class="card-title card-title-colored">
         <h2 class="card-title-text"><%= t ".title" %></h2>
-        <% if policy(RightsRequest).create? %>
+        <% if policy(RightsRequest).create? or @rights_request.errors.any? %>
```
niknetniko, line 6:
```suggestion
        <% if policy(RightsRequest).create? || @rights_request.errors.any? %>
```

I think we generally prefer `&&` and `||` over `and` and `or`, as those last ones have some non obvious precedence rules.

app/views/rights_requests/new.html.erb
```diff
@@ -3,14 +3,14 @@
...
           <div class="card-title-fab">
             <%= render 'fab_button', form: 'new_rights_request', icon: 'send' %>
           </div>
         <% end %>
       </div>
       <div class="card-supporting-text">
-        <% if policy(RightsRequest).create? %>
+        <% if policy(RightsRequest).create?  or @rights_request.errors.any? %>
```
niknetniko, line 13:
```suggestion
        <% if policy(RightsRequest).create?  || @rights_request.errors.any? %>
```

niknetniko (changes requested):

app/views/rights_requests/new.html.erb
jorg-vr: Thanks for the feedback, learned something new now by looking it up :)

niknetniko (approved):

chvp:
I'm not sure this is the correct fix. I feel like the policy's `create?` should not return false.

PR #3277, opened 2022-01-06
Sort copy courses list to show own courses first

### LINKED ISSUE #1661: Copy course page does not list own courses first

jorg-vr:
This pull request changes the order a user sees the course list when he wants to copy a course. Now their own courses are displayed first. Original order is kept as secondary ordering.

image

Closes #1661.
Closes #3049.

### DIFF
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -32,8 +32,9 @@ class CoursesController < ApplicationController
       @courses = @courses.where(featured: true)
     elsif params[:copy_courses]
       @courses = @courses.reorder(featured: :desc, year: :desc, name: :asc)
-      # use index to force stable sorting
-      @courses = @courses.sort_by.with_index { |course, i| [current_user.course_admin?(course) ? 0 : 1, i] }
+      @own_courses = @courses.select { |course| current_user.course_admin?(course) }
+      @other_courses = @courses.reject { |course| current_user.course_admin?(course) }
+      @courses = @own_courses + @other_courses
     end
 
     @courses = apply_scopes(@courses)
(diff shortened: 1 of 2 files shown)


bmesuere (changes requested):
In this implementation, it isn't really clear why some courses are sorted first. It would be clearer if the ones where the user is admin have an additional icon as is done in the general course listing: 

image

I'm not sure about the way the sorting is done (do we already require will_paginate/array in other pages?). Is this done at the ruby side or using a query?

jorg-vr:
I have added the Icon
image

@bmesuere  `require 'will_paginate/array'` is not used anywhere else in the code base
This was needed because `sort_by` returns an array instead of an `Active Record Relation`
sort_by preforms the sort in ruby, not in a database query

The query order/reorder does not support custom functions like `current_user.course_admin?(course)`, which is why I chose `sort_by`
I could write a database statement with the same result, but that would be rewriting the functionality of `current_user.course_admin?(course)` which could lead to two differing implementations in the future

Other option is using two or three separate select statements as in #3049 (but merge them instead of duplicating the listing code as @chvp  suggested)

jorg-vr:
Updated code `sort_by` to a `select` and `reject` statement to move the filtering logic to the database and also reduce complexity

`require 'will_paginate/array'` is still needed

jorg-vr:
By using `concat` instead of `+` I removed the need for `require 'will_paginate/array'`

bmesuere (approved):
Ok for me if it is for @chvp 

We generally try to keep the PR start post up to date before merging so it reflects the state of what is being merged. This usually means updating the screenshots to the final version.

chvp (approved):

niknetniko (approved):

PR #3278, opened 2022-01-06
Add external URL to machine annotations

### LINKED ISSUE #2845: Add external URL to machine annotations
We could add a `link` or `external` (the name is debatable, e.g. `info`, `external-info`, `url`, `external-url`, ...) field to the annotation data, which would contain a URL. Dodona can then render this link.

One use-case is linking to additional information for linter messages (see https://github.com/dodona-edu/universal-judge/issues/210). TESTed used to do this itself by rendering HTML, but this doesn't work anymore since #2844.

jorg-vr:
This pull request adds `externalUrl` to machine annotations. This externalUrl can be provided by judges and will be displayed as a note. 

image

image

This pr was tested with the TESTed judge https://github.com/dodona-edu/universal-judge/pull/294/files

Closes #2845 .


bmesuere (commented):
We generally use "draft pull requests" instead of a WIP label. This prevents the PR from accidentally being merged (and doesn't trigger the automatic review request).

niknetniko (changes requested):
We probably also want to add this to the full format, in `public/schemas/judge_output.json`.

bmesuere (approved):
I agree with @niknetniko his comment, but looks good otherwise.

niknetniko (approved):

PR #3287, opened 2022-01-11
Mark score item as busy before delay, to let page wait until delay finish

### LINKED ISSUE #3212: When navigating too fast, grades are not stored in grading module
When using the grading module, grades are not always correctly stored if I click the "next" button too fast after filling in the grades. Maybe the "next" button should be disabled as long as the grades are being saved? Or maybe they should also be saved if this button is clicked?

jorg-vr:
This pull request fixes the issue that a user can go to the next exercise before the scores are saved.

I also disabled the max, zero and remove buttons when the input field is already updating. This helps to prohibit concurrent updates.

Closes #3212.

### DIFF
--- a/app/assets/javascripts/feedback/score.ts
+++ b/app/assets/javascripts/feedback/score.ts
@@ -15,9 +15,9 @@ export default class ScoreForm {
     private readonly input: HTMLInputElement;
     private readonly expectedScore: HTMLInputElement;
     private readonly spinner: HTMLElement;
-    private readonly deleteButton: HTMLElement;
-    private readonly zeroButton: HTMLElement;
-    private readonly maxButton: HTMLElement;
+    private readonly deleteButton: HTMLButtonElement;
+    private readonly zeroButton: HTMLButtonElement;
+    private readonly maxButton: HTMLButtonElement;
     private readonly maxText: HTMLElement;
     private readonly form: HTMLFormElement;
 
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/feedback/score.ts
```diff
@@ -70,6 +70,7 @@ export default class ScoreForm {
             // long enough so the delay is useful when using the increment/decrement buttons
             // and the case where we type the value and don't want to wait.
             this.visualiseUpdating();
+            this.markBusy();
```
niknetniko, line 73: This seems to break the case where you use the up/down buttons of the form field, since those buttons immediately trigger the `change` event. This is the reason why we only call `visualiseUpdating` here: we apply visuals to ensure the user is aware we know a change has occurred, but we cannot disable the input yet, since the user might want to make further changes.

I think a potential solution worth investigating is moving the line `this.parent.registerUpdating(this.scoreItemId);` from `markBusy` to `visualiseUpdating`. There is code to make sure if you press 'Next', it will wait for the updating score items to complete.

niknetniko (changes requested):

app/assets/javascripts/feedback/score.ts
jorg-vr: Tried your solution and this also fixed the the `next` button issue, but the issue with being able to press the up and down buttons while updating remained (on slow internet)

Fixed that by disabling those as well in the `disableInputs` method. Now everything works as expected in my tests

jorg-vr: good to know is that the `change` event only triggers after the input looses focus, so I think at this point we can safely disable all inputs

jorg-vr: Ah wait I get what you mean with up/won buttons, will change

jorg-vr: Ok applied your feedback now

niknetniko (approved):

chvp (approved):

bmesuere (approved):

PR #3290, opened 2022-01-13
Fix the amount of pixels of total score input on the grading page

jorg-vr:
Before:
image

After:
image

Issue only happened on specific zoom levels because the input field had height `33.33px` while the surrounding button-tags had height `33px`


bmesuere (approved):
I'm sure @toonijn will be very happy 😉

PR #3293, opened 2022-01-13
Enable additional host names

bmesuere:
This pull request contains changes to make Dodona work with multiple host names on the same instance. This larger goal is tracked in #3299.

The specific changes in this pull request are:
- Set a relative asset host to host the assets from the same domain as the main request. Side effect: the assets from the sandboxed domain are loaded again
- Provide different SAML metadata depending on the hostname of the request
- Use correct host name for OIDC and LTI requests
- Add the OAuth credentials to the staging environment to allow testing
- Allow `/media` requests from the alt hostname
- Enable dodona.be as alternative hostname in production
- Allow all hosts names as parent for sandbox iframes


config/environments/staging.rb
```diff
@@ -18,7 +20,7 @@
   config.hosts << config.default_host << config.alt_host << config.sandbox_host
 
   # Where we host our assets (a single domain, for caching)
-  config.action_controller.asset_host = 'naos.ugent.be'
+  config.action_controller.asset_host = '/'
```
chvp, line 23: I don't immediately see a way around this, but this means that assets in exercise descriptions are served from the sandbox domain and are not using the same cache as assets for the rest of the site. At least the comment above should be updated, since it isn't accurate anymore.

app/controllers/auth/saml_controller.rb
```diff
@@ -2,6 +2,6 @@
 
 class Auth::SamlController < ApplicationController
   def metadata
-    render xml: OmniAuth::Strategies::SAML::Metadata.generate
+    render xml: OmniAuth::Strategies::SAML::Metadata.generate(false, request.host)
```
rien, line 5: We might consider explicitly checking whether `request.host` is in `config.web_hosts`. I don't think anything harmful could be done, but it seems sensible to check?

rien (approved):

lib/SAML/metadata.rb
```diff
@@ -15,9 +15,9 @@ class Metadata
         #                               (No pretty print if you gonna validate the signature)
         # @return [String] XML Metadata of the Service Provider
         #
-        def self.generate(pretty_print = false)
+        def self.generate(pretty_print = false, host = Rails.configuration.default_host)
```
chvp, line 18: Comment above shoud be updated.

config/environments/staging.rb
chvp: I'm still not entirely a fan of this. I assume this is not something we want to do in our production config? Keeping staging as close as possible to production seems like something we want to do?

bmesuere: I added a todo to #3299 to switch this back after the migration.

PR #3295, opened 2022-01-14
Make cloning git repository asynchronous

### LINKED ISSUE #1643: Make creating a new repository asynchronous
When a repository is created, Dodona performs a `git clone` within the request. This can possibly take a while, "hanging" the request.

To improve the user experience, this should be made asynchronous.

jorg-vr:
This pull request makes cloning a repository a delayed job. This improves UX as the frontend does not hang while a git repository is cloning.

The show page of the judge and repository display a warning as long as the clone has not completed succesfully.  (Requires page refresh to check again if warning is relevant)

It also displays an error notice if the clone failed.

image
image
image

image
image
image

Closes #1643 .


config/locales/views/judges/en.yml
```diff
@@ -14,3 +14,4 @@ en:
       webhook_html: "If you set up a <a href='https://developer.github.com/webhooks/creating/#setting-up-a-webhook' target='_blank' rel='noopener'>webhook</a> on GitHub, Dodona will automatically update your judge on each commit:"
       all_submissions: "Submissions for this judge"
       all_exercises: "Exercises for this judge"
+      clone_incomplete: "Cloning in progress. It can take a few minutes before the judge is uploaded to the server and ready for use."
```
bmesuere, line 17:
```suggestion
      clone_incomplete: "We are adding your judge to Dodona. This can take up to five minutes."
```

bmesuere (changes requested):
I added a suggestion to rephrase the message to the user (to be adapted/translated for the other string if ok).

You have added the job to the git_clone queue, but no workers are watching this queue. Option 1 is to add this queue to the current workers in the config/deploy files, option 2 is to reuse an existing queue. @chvp is better placed to decide which option is best.

chvp:
I think using the same queue as the one used for reprocessing the repositories makes the most sense. This is currently the default queue, but we might want to make this a named queue also (in light of efforts related to #3299).

jorg-vr:
I updated the translations and added a 'git' queue for cloning and reprocessing repositories

config/locales/views/judges/en.yml
```diff
@@ -14,3 +14,4 @@ en:
       webhook_html: "If you set up a <a href='https://developer.github.com/webhooks/creating/#setting-up-a-webhook' target='_blank' rel='noopener'>webhook</a> on GitHub, Dodona will automatically update your judge on each commit:"
       all_submissions: "Submissions for this judge"
       all_exercises: "Exercises for this judge"
+      clone_incomplete: "We are adding your judge to Dodona. This can take up to five minutes."
```
bmesuere, line 17:
```suggestion
      clone_incomplete: "We are now adding your judge to Dodona. It can take up to five minutes until it becomes available."
```

config/locales/views/judges/nl.yml
```diff
@@ -14,3 +14,4 @@ nl:
       webhook_html: "Stel op GitHub een <a href='https://developer.github.com/webhooks/creating/#setting-up-a-webhook' target='_blank' rel='noopener'>webhook</a> in zodat de judge op Dodona automatisch geüpdatet wordt bij elke commit:"
       all_submissions: "Inzendingen voor deze judge"
       all_exercises: "Oefeningen voor deze judge"
+      clone_incomplete: "Wij zijn uw judge aan het toevoegen aan Dodona. Dit kan tot 5 minuten duren."
```
bmesuere, line 17:
```suggestion
      clone_incomplete: "We zijn je judge nu aan het toevoegen aan Dodona. Het kan tot 5 minuten duren eer hij beschikbaar is."
```

config/locales/views/repositories/en.yml
```diff
@@ -25,6 +25,7 @@ en:
       public_files_placeholder: This repository has no public files.
       file_name: File name
       file_url: URL
+      clone_incomplete: "We are adding your repository to Dodona. This can take up to five minutes."
```
bmesuere, line 28:
```suggestion
      clone_incomplete: "We are now adding your repository to Dodona. It can take up to five minutes until your exercises show up."
```

config/locales/views/repositories/nl.yml
```diff
@@ -25,6 +25,7 @@ nl:
       public_files_placeholder: "Deze repository bevat geen publieke bestanden"
       file_name: Bestandsnaam
       file_url: URL
+      clone_incomplete: "Wij zijn uw repository aan het toevoegen aan Dodona. Dit kan tot 5 minuten duren."
```
bmesuere, line 28:
```suggestion
      clone_incomplete: "We zijn je repository nu aan het toevoegen aan Dodona. Het kan tot 5 minuten duren eer je oefeningen beschikbaar zijn."
```

bmesuere (changes requested):
Some more minor text suggestions. We try to avoid u/uw in communicating with users. Right now, the additional message is shown as callout on the top of the page, right under the "success" message which is styled differently. I would maybe move the message to inside the card (e.g., right side).

app/models/concerns/gitable.rb
```diff
@@ -30,6 +37,12 @@ def clone_repo
     throw :abort
   end
 
+  def clone_complete?
+    out, error, status = Open3.capture3('ls', chdir: full_path.to_path)
+    errors.add(:remote, error) unless status.success?
+    out != ''
```
chvp, line 43: Not sure about this (for two reasons). First of, I'd use Ruby's builtin File listing instead of shelling out. Also, just checking that there are files is not sufficient for knowing if the checkout finished, since it happens quite often that people add empty repositories to Dodona (sometimes even without commits).

chvp (changes requested):

app/models/concerns/gitable.rb
jorg-vr: Wasn't sure how often that would happen, that does indeed break things

The only other option I can think of is adding a state field to both the judge and the repository table which is set to complete after the run (As I don't think were able to add such a field to `gitable`)

chvp: You could technically check the delayed jobs table, but I don't think that would be a cleaner solution than just keeping an extra field.

db/migrate/20220118102649_add_status_to_judges.rb
```diff
@@ -0,0 +1,5 @@
+class AddStatusToJudges < ActiveRecord::Migration[6.1]
+  def change
+    add_column :judges, :status, :integer
```
chvp, line 3: This column should probably be `null: false` and initialised to complete for any existing repositories/judges?

app/models/concerns/gitable.rb
```diff
@@ -1,6 +1,10 @@
 module Gitable
   extend ActiveSupport::Concern
 
+  included do
+    enum status: { unknown: 0, 'clone queued': 1, cloning: 2, 'clone complete': 3, 'clone failed': 4 }
```
chvp, line 5: Is the `unknown` status required? It doesn't seem used, and I feel like we have a sensible default for existing `Gitable`s? Also maybe `clone_status` as the field name and remove the spaces in the names, just to tighten things up a bit? Although I don't feel too strongly about this.

chvp (changes requested):

bmesuere (approved):
Except for the comment from @chvp, this looks good 👍

config/locales/views/judges/nl.yml
```diff
@@ -14,3 +14,5 @@ nl:
       webhook_html: "Stel op GitHub een <a href='https://developer.github.com/webhooks/creating/#setting-up-a-webhook' target='_blank' rel='noopener'>webhook</a> in zodat de judge op Dodona automatisch geüpdatet wordt bij elke commit:"
       all_submissions: "Inzendingen voor deze judge"
       all_exercises: "Oefeningen voor deze judge"
+      clone_incomplete: "We zijn je judge nu aan het toevoegen aan Dodona. Het kan tot 5 minuten duren eer hij beschikbaar is."
+      clone_failed: "Het clonen van de judge is mislukt. Kijk na of Dodona lees en schrijf rechten heeft op jouw judge."
```
niknetniko, line 18:
```suggestion
      clone_failed: "Het clonen van de judge is mislukt. Kijk na of Dodona lees- en schrijfrechten heeft op jouw judge."
```

app/models/concerns/gitable.rb
niknetniko: Perhaps also useful to add the `_prefix: :clone` option, which will prefix all helper methods with `clone`. This will then result in the `judge.clone_failed?`, `judge.clone_completed?`, etc. methods, instead of `judge.failed?` and `judge.completed?` if I understand correctly how it works.

If we were to use the `prefix` option, you wouldn't need to define `clone_failed?` yourself anymore (on line 53 of this file).

config/locales/views/repositories/nl.yml
```diff
@@ -25,6 +25,8 @@ nl:
       public_files_placeholder: "Deze repository bevat geen publieke bestanden"
       file_name: Bestandsnaam
       file_url: URL
+      clone_incomplete: "We zijn je repository nu aan het toevoegen aan Dodona. Het kan tot 5 minuten duren eer je oefeningen beschikbaar zijn."
+      clone_failed: "Het clonen van de repository is mislukt. Kijk na of Dodona lees en schrijf rechten heeft op jouw repository."
```
niknetniko, line 29:
```suggestion
      clone_failed: "Het clonen van de repository is mislukt. Kijk na of Dodona lees- en schrijfrechten heeft op jouw repository."
```

app/models/concerns/gitable.rb
niknetniko: Perhaps also useful to add the `_prefix: :clone` option, which will prefix all helper methods with `clone`. This will then result in the `judge.clone_failed?`, `judge.clone_completed?`, etc. methods, instead of `judge.failed?` and `judge.completed?` if I understand correctly how it works.

If we were to use the `prefix` option, you wouldn't need to define `clone_failed?` yourself anymore (on line 53 of this file).

niknetniko (approved):

db/migrate/20220118102649_add_clone_status_to_judges.rb
```diff
@@ -0,0 +1,5 @@
+class AddCloneStatusToJudges < ActiveRecord::Migration[6.1]
+  def change
+    add_column :judges, :clone_status, :integer, null: false, default: 3
```
chvp, line 3: I wonder whether it would be a good idea to add an extra statement after this changing the default to 1. The application code already handles this, but it makes more sense for resulting default to be the initial state I think.

pdawyndt:
“lees- en schrijfrechten”

chvp (approved):

PR #3300, opened 2022-01-16
add line blur on hover

tibdhond:
When hovering over a legend item, every line except the one corresponding to the item will be blurred.
This should greatly assist in identifying the correct line for each exercise.

Peek 2022-01-16 01-26

### DIFF
--- a/app/assets/javascripts/visualisations/cumulative_timeseries.ts
+++ b/app/assets/javascripts/visualisations/cumulative_timeseries.ts
@@ -171,8 +171,7 @@ export class CTimeseriesGraph extends SeriesGraph {
                 .x(d => this.x(d.x1))
                 .y(d => this.y(d.cSum / this.studentCount))
                 .curve(d3.curveMonotoneX)(ex.ex_data)
-            )
-            .attr("data-index", ex => ex.ex_id);
+            );
     }
 
     /**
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/visualisations/cumulative_timeseries.ts
```diff
@@ -391,16 +392,29 @@ export class CTimeseriesGraph extends SeriesGraph {
             .data(this.exOrder)
             .enter()
             .append("div")
-            .attr("class", "legend-item");
+            .attr("data-index", exId => exId)
+            .attr("class", "legend-item")
+            .on("mouseover", e => {
+                const lines = this.graph
+                    .selectAll(`.line:not([data-index='${e.target.dataset.index}'])`);
+                lines.style("opacity", 0.2);
```
bmesuere, line 400: You don't need the complicated css selector here (nor the data attribute added before). The d3-way to do this is to use a function accessing the data ("d"), with a conditional operator returning 0.2 or 1.

bmesuere (changes requested):
This is a great improvement :+1:

app/assets/javascripts/visualisations/cumulative_timeseries.ts
tibdhond: should be better now

bmesuere (approved):

chvp (approved):

PR #3314, opened 2022-01-17
Move minitest from development to test gem

### LINKED ISSUE #3167: WebMock seems to insert itself in the delayed job runner in development
This makes it so one can't evaluate submissions, since the docker communication is HTTP over a socket, which webmock then breaks.

jorg-vr:
This pull request moves minitest to the test group in the gemfile. This solves the webmock issue as mintest loads webmock as one of its dependencies. 

I have not found any reason why minitest is needed in development as this is clearly a testing library. @bmesuere or @chvp can you confirm or deny this?

Closes #3167 .

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -143,9 +143,6 @@ group :development, :test do
   # test template rendering
   gem 'rails-controller-testing', '~> 1.0.5'
 
-  # I18N default locale & better test reporter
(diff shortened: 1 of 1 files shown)


chvp (approved):
@rien added it there originally, I don't see a reason for it needing to be there.

bmesuere (approved):

PR #3327, opened 2022-01-24
Merge users through the console

### LINKED ISSUE #2668: Merge users through the console
Add the ability to more easily merge users through the console (for now).

This was part of #2452.

jorg-vr:
This pull request adds a function `merge_into` to the user model. This function can be used to merge two user from the command line.

Also added an error if the two users don't have the same permissions. This error can be circumvented by using force, in  which case the highest permissions will be kept for the new user.

For overlapping unique properties, such as right_requests, read_states and identities the values of other are kept.
For overlapping course memberships, the highest permission is kept. 

I have added a rake task to run this from a command line interface: (Without rails console)
example input: invalid, swap, yes
image

- [x] Tests were added

Closes #2668 .


lib/tasks/merge_users.rake
```diff
@@ -0,0 +1,62 @@
+
+task :merge_users, [:arg1, :arg2] => :environment do |task, args|
+  u1_id = args[:arg1].to_i
+  u2_id = args[:arg2].to_i
+
+  MergeUsers.new.merge_users_interactive u1_id, u2_id
+end
+
+class MergeUsers
```
jorg-vr, line 9: Not sure if the rake task is the best place to put this, but I am unsure where else to put it either

app/models/user.rb
```diff
@@ -290,6 +290,73 @@ def set_search
...
+    errors.add(:merge, 'User has different permissions') if other.permission != permission && !force
+    return false if errors.any?
+
+    transaction do
+      other.permission = permission if (permission == 'staff' && other.permission == 'student') \
+                                    || (permission == 'zeus' && other.permission != 'zeus')
+
+      other.institution_id = institution_id if other.institution_id.nil?
+
+      rights_request.update(user: other) if !rights_request.nil? && other.permission == 'student' && other.rights_request.nil?
```
bmesuere, line 304: Are these "if modifiers" enforced by rubocop? If they are more complex like the first and third one, a classical if seems more appropriate.

app/models/user.rb
```diff
@@ -290,6 +290,73 @@ def set_search
...
+
+      submissions.each { |s| s.update(user: other) }
+      api_tokens.each { |at| at.update(user: other) }
+      events.each { |e| e.update(user: other) }
+      exports.each { |e| e.update(user: other) }
+      notifications.each { |n| n.update(user: other) }
+      annotations.each { |a| a.update(user: other, last_updated_by_id: other.id) }
+      questions.each { |q| q.update(user: other) }
```
bmesuere, line 312: I think these updates and the ones below cover most aspects of our users. Some open thoughts:

- How do we handle grading?
- Calling update on each of these will generate a huge number of queries, especially for submissions. Using update_all will do this in a single query, but will skip all validations and callbacks. It might be worth checking if update_all can be used for submissions and read states.
- Somewhat linked with the previous point: the necessary caches should be invalidated (e.g. the number of users that solved an exercise within a course). This might happen automatically using the callbacks/validations but it should be double checked.
- You of course added some tests, but I was curious if you checked the order of the updates. For example, validations (or cache recalculations) that might fail if a users has submissions in a course  which he is at that moment not a member off (the membership is updated last).

lib/tasks/merge_users.rake
bmesuere: Since it's only used in this rake task, I think it's ok.

bmesuere (commented):
👍 This will certainly be useful. I have added a few remarks below.

app/models/user.rb
jorg-vr: they are indeed enforced by rubocop

app/models/user.rb
jorg-vr: I indeed seem to have forgotten about grading (As it was not mentioned as a relationship at the top of the model) Will add code for evaluations

Will explore the update_all statements

I have not yet checked any cached values, will do this now :)

Order of updates has not caused me any trouble in my manual testing. But I see this is not explicitly tested in the test cases, so I will expend upon this to make sure this does not cause any errors

app/models/user.rb
```diff
@@ -290,6 +290,73 @@ def set_search
     self.search = "#{username || ''} #{first_name || ''} #{last_name || ''}"
   end
 
+  def merge_into(other, force: false)
+    errors.add(:merge, 'User belongs to different institution') if other.institution_id != institution_id && other.institution_id.present? && institution_id.present?
+    errors.add(:merge, 'User has different permissions') if other.permission != permission && !force
+    return false if errors.any?
+
+    transaction do
```
niknetniko, line 298: (this is not my full review, but something I noticed while taking a quick look)

What does this `transaction` block intend to make atomic, i.e. when do we foresee it to potentially fail?

I see a lot of `update` and `destroy` statements, but if I'm not mistaken, those won't abort on failure, but return true/false, and thus won't trigger the transaction rollback. If this is intentional, perhaps we could add a comment saying why we need a transaction, as it isn't immediately clear to me. Otherwise, perhaps we should see if we can use the `update!` and `destroy!` methods.

jorg-vr: didn't know this about update and destroy, I expected them to fail 
I will explore the `update!` and `destroy!` versions

chvp (approved):
Looks good. I don't have any comments other than what was already mentioned by Bart or Niko.

app/models/user.rb
jorg-vr: On Grading:
I have added a merge of evaluation users. When two users appear in the same evaluation I keep the evaluation of other.
I am hesitating to just block the merge of users in this case, because removing the grades of a user without manually checking which grade is most applicable can led to real world issues. It also should not happen very often, so it is acceptable to break here, and force this merge to be done manually.

On update_all:
This indeed forgoes validations, which breaks caching. I am weary to combine an update all statement with a custom rewrite of the applicable validations, as this code will diverge over time from the actual validations that need to happen .

On caching:
I have added tests to validate some cached parameters. These broke when testing the `update_all` statements for example

On order:
In general it seems possible to have submissions linked to a course which you are not a member of. So i was unable to produce failing tests. Yet I did update the order a bit to be more logical in case later validations are written that break this. (Although there will always be inconsistencies while the merge is in progress)

bmesuere (approved):
Ok for me, but I would like approval from @niknetniko before merging.

niknetniko (approved):
Looks good to me :+1:

PR #3328, opened 2022-01-24
Store last seen and last sign in in the database

### LINKED ISSUE #3296: Store last seen and last sign in in the database
Add two additional field to the user table: `seen_at` and `sign_in_at`. These timestamps should be updated on page load and sign in respectively. This helps us detect inactive accounts, better tracks sign in events for security purposes, allows us to show certain notices based on last visit and can help us get an overview of the current users on site (of course only in an aggregated form). It is important to note that we only store the last event and not keep a history to minimize privacy implications.

Some technical considerations for implementing:
- The rails `touch` method is a great fit. Touch skips validations and callbacks and thus has minimal overhead.
- Instead of calling touch on every request, we could opt to exclude api/js requests (such as for notifications) and/or update only if the last update was longer than five minutes ago.

jorg-vr:
This pull request adds two database fields to `users`: `seen_at` and `sign_in_at`. These are updated on page load and sign in respectively.

Closes #3296  .

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -27,6 +27,8 @@ class ApplicationController < ActionController::Base
 
   around_action :user_time_zone, if: :current_user
 
+  after_action :set_user_seen_at, if: :current_user
+
   before_action :set_time_zone_offset
 
   before_action :set_notifications, if: :user_signed_in?
@@ -59,6 +61,7 @@ class ApplicationController < ActionController::Base
       raise "User with id #{user.id} should not have a blank email " \
             'if the provider is not LTI, OIDC or Smartschool'
     end
+    user.touch(:sign_in_at)
   end
 
   Warden::Manager.after_set_user do |user, _auth, _opts|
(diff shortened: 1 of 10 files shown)


app/controllers/application_controller.rb
```diff
@@ -27,6 +27,8 @@ class ApplicationController < ActionController::Base
 
   around_action :user_time_zone, if: :current_user
 
+  after_action :set_user_seen_at, if: :current_user
```
bmesuere, line 30: I would also skip this for `remote_request?` calls.

bmesuere (changes requested):
I would add an index on `seen_at` since we will query this often for secondary statistics.

jorg-vr:
@bmesuere I have added the index but I am not 100% sure that it is a good idea. The field is updated very often, each update being slowed down by also having to update the index. 
While the added benefit only applies for our analysis requests, which will probably only happen infrequently.
So overall our database might be slowed down instead of sped up by this index

bmesuere:
@jorg-vr you are right. In addition, we will query this on elysium so the select will have no impact on the production database. We could always add an index later.

niknetniko (approved):

bmesuere (approved):
We can monitor the number of update queries in datadog and add a timeout if needed.

Issue #2809, opened 2021-06-04
Add `Copy series` button and move `Delete series` button down


BTWS2:
To make an identical list for students with facilities, the teacher has to copy paste the series description and add the same exercises one by one. Having a `Copy series` button would ask the teacher to modify the series name and make a series with the same description, exercises and properties.

`Delete series` is the second button next to two important buttons: `Edit series` and `Show evaluation`. This is prone to accidental clicking. Deleting a series doesn't happen that often and is irreversible (after a warning), so I would argue it can be moved down in the list.

Before:
![image](https://user-images.githubusercontent.com/56451049/120841863-5057c000-c56c-11eb-8a7d-afd6bd9510e9.png)
After:
![image](https://user-images.githubusercontent.com/56451049/120843238-2b644c80-c56e-11eb-9405-87fa240b83ef.png)

chvp:
I reworked the title to be only about copying series, since this would not be a trivial feature to add, and the order of the items in the dropdown menu is unrelated.

BTWS2:
> Would be nice if series could be copied (use case: create test/exam series with alternative deadlines). At least with the same course, but also between courses.
> 
> Issue jointly created with @toonijn

Issue by @pdawyndt (#3329)

BTWS2:
> Zou het mogelijk zijn om een reeks gemakkelijk te dupliceren? Dan zou ik een reeks kunnen dupliceren en aan elke reeks een andere deadline geven en via de geheime link de reeks bezorgen aan de juiste studenten.
> Als ik dit nu wil doen moet ik een nieuwe reeks toevoegen en dan handmatig alle oefeningen opnieuw toevoegen.
> Dit zou het bovenstaande probleem al een beetje oplossen.

Comment by @IngeVerbeek (https://github.com/dodona-edu/dodona/discussions/3400#discussioncomment-2187543)

PR #3330, opened 2022-01-25
Move delete series button down

BTWS2:
`Delete series` is the second button next to two important buttons: `Edit series` and `Show evaluation`. This is prone to accidental clicking. Deleting a series doesn't happen that often and is irreversible (after a warning), so I would argue it can be moved down in the list. In most GUI's the delete button is near the bottom in the list of options.

Before:
image
After:
image

Was a part of #2809.

### DIFF
--- a/app/views/series/_series.html.erb
+++ b/app/views/series/_series.html.erb
@@ -54,19 +54,6 @@
                 <% end %>
               </li>
             <% end %>
-            <% if policy(series).destroy? %>
-              <li>
-                <%
-                    confirm = t('series.delete.confirm')
-                    if series.evaluation.present?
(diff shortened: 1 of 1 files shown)


bmesuere:
We discussed this today. Edit and delete are semantically related actions and belong together because they both directly modify the series. It would even make sense to add a divider below delete. It therefore seems not appropriate to move the item in the menu. There might be a risk of clicking delete by accident, but that is exactly why we have the confirmation.

The potential data loss is limited: no submissions are deleted on this action. You can recreate the series with the same exercises and the state for the students will automatically be restored. In extreme cases where the description contained crucial information we can restore from the database backup.

Issue #2866, opened 2021-06-17
Automate DB backup restore to naos


bmesuere:
If we want to further automate releases, it is important to validate the release and especially DB migrations. To facilitate this, a first step would be to restore production DB backups to Dodona.

An added benefit is that this way would validate our backups.

Note that simply restoring the backup would probably result in a non-functional deployment because the submissions and results are missing. If this is handled gracefully, it shouldn't be a problem.

The implementation can probably be done with a script on mestra.

PR #3331, opened 2022-01-28
Merge complex institutions with overlapping users

jorg-vr:
This pull request builds upon #3327  but is split off to keep the individual pull request manageable.

It introduces an interactive script to merge two institution who potentially have overlapping users. It uses the merge_users interactive script to merge the overlapping users first.

This all happens in a single transaction, as both the merge of users and that of institutions depend on each other.

Because of this transactional nature I had to introduce a new gem [after_commit_everywhere](https://github.com/Envek/after_commit_everywhere) 
This gem makes sure certain side effects are only applied after the full transaction is successfully committed.

See the test `The script should also rollback filesystem changes` for a test case that was broken without this gem.

Peek 2022-01-31 16-59

The relevant command is printed in the error message when merging institutions using the web interface has failed
image

- [x] Tests were added


app/models/submission.rb
```diff
@@ -497,8 +497,10 @@ def update_fs
     new_path = fs_path
     return if old_path == new_path
 
-    FileUtils.mkdir_p File.dirname new_path
-    FileUtils.move old_path, new_path
+    AfterCommitEverywhere.after_commit do
```
bmesuere, line 500: maybe add a little comment here

bmesuere (changes requested):
Right now, we already offer an institution merge on the website but this sometimes fails due to overlapping users. Since this PR provides a way to merge these institutions semi-automatically, it would be nice if the command that is needed for this is printed along with the error message that is shown when the automatic merge fails.

I'm thinking about a way on how to test this with actual data without risking data loss. The only way I can think of is copying some problematic data from the production environment to naos but this might be a lot of work (this is why #2866 would be nice).

The code itself looks ok.

app/runners/merge_institutions.rb
```diff
@@ -0,0 +1,73 @@
+class MergeInstitutions
```
chvp, line 1: I'm not sure `app/runners/` is the correct location for this. `app/runners` is currently for everything related to the running of submissions, which this is unrelated to.

chvp (changes requested):

app/runners/merge_institutions.rb
bmesuere: You're right, I missed the file path of this one. Since its strictly related to the rake task, was there a specific reason for not putting it in that same file?

jorg-vr: It imports from MergeUsers. Was not sure if rake tasks importing classes from other rake tasks is really the way to go.
It should probably be moved to a different folder?

chvp: Are `.rb` files allowed in `lib/tasks`. Or does that conceptually not make sense? Otherwise, storing them alongside the tasks seems fine to me.

bmesuere: maybe in a separate subdirectory in lib/tasks?

chvp (approved):

niknetniko (approved):

bmesuere (approved):
Looking forward to merge some problematic institutions :)

PR #3336, opened 2022-02-01
Replace sassc-rails with css-bundling and dart sass

### LINKED ISSUE #3289: scss files no longer rebuild automatically
This bug was created by: https://github.com/dodona-edu/dodona/pull/3272

The issue is also mentioned here: https://github.com/sass/sassc-rails/issues/55
https://files.slack.com/files-pri/TNLH8C0LC-F02TF607XN1/image.png

### Workaround:
`rm -rf tmp/cache`

### Tested  and failed solutions:

Keeping only `.css` extensions on import statements results in the following bug:
```
ActionView::Template::Error (Error: Undefined variable: "$development".
        on line 70:5 of app/assets/stylesheets/theme/theme-light.css.scss
        from line 11:9 of app/assets/stylesheets/base.css.scss
>> @if $development == true {

   ----^
):
```

removing all extensions on import results in the following bug:
```
Error: File to import not found or unreadable: base.
        on line 6:1 of app/assets/stylesheets/application.css.scss.erb
>> @import "base";
```

jorg-vr:
This pull request replaces `sassc-rails` with  [cssbundling-rails](https://github.com/rails/cssbundling-rails) as a new gem to handle css compilation. This because `sassc-rails` depends on [libsass](https://github.com/sass/libsass) which is deprecated.

`cssbundling-rails` is supported by the core rails team, and the recommended library for css bundling in Rails 7.

Followed this migration guide: https://dev.to/kolide/how-to-migrate-a-rails-6-app-from-sass-rails-to-cssbundling-rails-4l41

This PR also solves some deprections warnings that appeared now that we use a more recent version of sass: namely replacing `/` by `math.divide`

Closes #3289.
Part of #3316

Note here for future reference of a very nasty bug that appeared when `sassc-rails` was removed:
An scss import in `vendor.scss` used the `.css` extension instead of `.css.scss`
This resulted in the imported file not being bundled by `dart-sass`.
This resulted in the browser requesting this css file seperatly to rails. Rails sees that the css file is actually a `.scss` file, which it does not have in its cache, so it tries to process the `scss` itself . (This all happens in sprockets)
For this it tries to load `sassc` which is no longer installed because `sassc-rails` was removed.
Which resulted in the error `cannot load such file -- sassc`

Bug fix was simply changing the extension to `.css.scss`


bmesuere:
I tested this locally and worked great. The watcher recompiled all files automatically.

chvp (commented):
Is the `/app/assets/builds` path standard? It feels a bit weird to me to have build output in the `app` folder. Otherwise looks good.

jorg-vr:
@chvp  App/assets/builds is de nieuwe standaard inderdaad

Uit de [README](https://github.com/rails/cssbundling-rails)
> This gem provides installers to get you going with the bundler of your choice in a new Rails application, and a convention to use app/assets/builds to hold your bundled output as artifacts that are not checked into source control (the installer adds this directory to .gitignore by default).

chvp (approved):

niknetniko (approved):
Works for me :+1:

bmesuere (approved):

PR #3340, opened 2022-02-03
Remove Google Analytics

### LINKED ISSUE #3297: Remove Google Analytics
Our use of Google Analytics is already quite minimal. All "data sharing" options are disabled in the settings to minimize the privacy impact. In practice, we only use this to monitor the number of real time active users (e.g. at the start of an exam). If #3296 is implemented, we can get the same information from aggregated queries on phocus.

- Remove the tracking code
- Remove custom event tracking code
- Remove the mention of GA in our privacy policy and "your data" page
- Check if this was an exception to our CSP rules and remove if so

bmesuere:
This pull request removes Google Analytics from our code and privacy statement. After this is in production, we can request a removal of all our data from Google.

We will lose
- historical statistics regarding visitors, sessions duration and page views. The most relevant number here is the number of users and this is more accurately captured by the submission data which we have in the database.
- real time data on the number of users. The newly added "last seen" will mostly replace this insight. It will require an additional dashboard on phocus.
- tracking of "events" such as tutor invocations. We have no direct replacement for this, but they weren't used a lot anyway.

This should only be merged after 5.3 is released.

Closes #3297


app/assets/javascripts/exercise.js
```diff
@@ -1,5 +1,5 @@
 /* globals Bloodhound,ace,ga */
```
jorg-vr, line 1: `ga` is also used in this file, and these usages should probably be removed as well

niknetniko (approved):

config/initializers/content_security_policy.rb
```diff
@@ -12,12 +12,10 @@
   if Rails.env.development?
     # Allow webpack-dev-server
     policy.connect_src :self, Rails.configuration.tutor_url.to_s,
-                       'https://www.google-analytics.com',
                        'https://*.googleapis.com',
                        'http://localhost:3035', 'ws://localhost:3035'
   else
     policy.connect_src :self, Rails.configuration.tutor_url.to_s,
-                       'https://www.google-analytics.com',
                        'https://*.googleapis.com'
```
chvp, line 19: Why do we still have this in our `connect-src`? AFAICT we don't do any XHR requests to this domain? Could this be a leftover from our firebase usage?

bmesuere: I have searched for the domain in the code base and only found 2 uses:
- the inclusion of a stylesheet for the roboto webfont
- the inclusion of a hosted version of jquery in the python tutor file

None of these fall under the connect-src directive so I think it can be removed.

chvp (approved):

PR #3341, opened 2022-02-03
Upgrade to rails 7

### LINKED ISSUE #3316: Upgrade to rails 7

jorg-vr:
This pull request upgrades rails to version 7.0.1

This upgrade guide was used: https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#upgrading-from-rails-6-1-to-rails-7-0

This pr also fixes some bugs that arose with the new rails version. Normally each has a relevant comment.

Two other prs build on this to bring us closer to the new rails 7 philosophy: #3366  and #3368 

Closes #3316.


app/controllers/score_items_controller.rb
```diff
@@ -93,7 +93,7 @@ def preload_eval_exercise(item)
     # Reload the evaluation, since one of the items changed.
     @evaluation.score_items.reload
     # Preload the stuff for one exercise and an existing instance.
-    ActiveRecord::Associations::Preloader.new.preload(item.evaluation_exercise, [score_items: :scores])
+    ActiveRecord::Associations::Preloader.new(records: item.evaluation_exercise, associations: [score_items: :scores])
```
jorg-vr, line 96: Fixes a bug caused by caused by an API change of `ActiveRecord::Associations::Preloader` [old](https://www.rubydoc.info/docs/rails/ActiveRecord/Associations/Preloader) [new](https://msp-greg.github.io/rails_stable/ActiveRecord/Associations/Preloader.html)

New api docs state _"Do not use. This class is for internal use only."_ so the api of this class will probably not be guaranteed to be stable in the future and other solutions for this should be explored in the future

app/controllers/exports_controller.rb
```diff
@@ -86,7 +86,7 @@ def create(item, list)
       params.delete(:with_labels)
     end
 
-    Export.create(user: current_user).delay(queue: 'exports').start(item, list, ([@user] if @user), params)
+    Export.create(user: current_user).delay(queue: 'exports').start(item, list, ([@user] if @user), params.to_unsafe_h)
```
jorg-vr, line 89: Fixes a bug.

To avoid the use of `unsafe` we should specifically lists which params are permitted here. (I did not do this myself as I was not sure which params we are expecting in this part of the code)

app/views/series/show.js.erb
```diff
@@ -1,4 +1,4 @@
-<% series_card = render partial: 'series.html.erb',
```
jorg-vr, line 1: Fixes a bug that appeared on the course detail page

Gemfile
```diff
@@ -83,8 +86,8 @@ gem 'pundit', '~> 2.1.1'
 # impersonate users
 gem 'pretender', '~> 0.4.0'
 
-# db annotations
-gem 'annotate', '~> 3.1.1'
+# db annotations Temporary use of github branch to support rails 7, should be reverted on new release
+gem 'annotate', github: 'dabit/annotate_models', branch: 'rails-7'
```
chvp, line 90: How will we get notified of this new release?

app/controllers/exports_controller.rb
chvp: What was the bug? I prefer to do bugfixes in separate PRs, to make sure that it we are explicit why the change was necessary (or is the bug introduced by the upgrade to rails 7?).

app/controllers/score_items_controller.rb
chvp: An issue should be created then, IMO.

Gemfile
niknetniko, line 90: Is the new 3.2.0 release enough? (https://github.com/ctran/annotate_models/releases/tag/v3.2.0) Otherwise we should probably add a comment with which PR/commit needs to be released to be able to go back to normal versions

config/initializers/new_framework_defaults_7_0.rb
```diff
@@ -0,0 +1,117 @@
+# Be sure to restart your server when you modify this file.
+#
+# This file eases your Rails 7.0 framework defaults upgrade.
+#
+# Uncomment each configuration one by one to switch to the new default.
+# Once your application is ready to run with all new defaults, you can remove
+# this file and set the `config.load_defaults` to `7.0`.
```
chvp, line 7: Did you follow these instructions?

Gemfile
jorg-vr: new realease is enough, seems to be released two hours ago, so will update this pr

Gemfile
```diff
@@ -86,8 +86,8 @@ gem 'pundit', '~> 2.1.1'
 # impersonate users
 gem 'pretender', '~> 0.4.0'
 
-# db annotations Temporary use of github branch to support rails 7, should be reverted on new release
-gem 'annotate', github: 'dabit/annotate_models', branch: 'rails-7'
+# db annotations
+gem 'annotate','~> 3.2.0'
```
chvp, line 90:
```suggestion
gem 'annotate', '~> 3.2.0'
```

app/controllers/exports_controller.rb
jorg-vr: The bug was introduced by rails 7, although I don't understand how it ever worked before. (The bug happens in a testcase, so it would also be visible if the problem existed in other branches)

The problem is that the delay function persists its input for later use. For this it probably calls some kind of to string/ to hash function on the objects

Reading the api docs of params, they say that certain functions called upon it will error as long as it contains not permitted parameters. This because params could contain more parameters then your function expects.

You can either explicitly permit the params you are expecting, or call an unsafe version of the api.
I chose the `unsafe` version because I have no idea which parameters are expected here, but I can look in to it further if desired.

config/initializers/new_framework_defaults_7_0.rb
jorg-vr: Not really, as the code introducing this file, also set `config.load_defaults` to `7.0` itself. 
So I ended up debugging all issues before discovering this file.

Afterwards I did read through all of them, but none of the settings seemed to impact dodona

chvp: Ah, I missed the change in `application.rb` when going over these changes. Never mind then.

app/controllers/score_items_controller.rb
jorg-vr: #3369

niknetniko (approved):

app/controllers/exports_controller.rb
jorg-vr: https://stackoverflow.com/questions/46029084/rails-unable-to-convert-unpermitted-parameters-to-hash

Problem popped up completely unrelated in commit 82857627402ce120558bcf5e0671e37189daff5a showing that the params issue is not unique to rails 7. Yet the test failing on this specific usecase of params does seem to be caused by this pr

jorg-vr:
Works on naos

bmesuere:
I can confirm I was signed out on naos after deploying, but I'm not sure if that isn't always the case.

I clicked around on naos and everything seems to work.

config/initializers/new_framework_defaults_7_0.rb
chvp: We've not done this in the past, but I wouldn't be against it.

PR #3343, opened 2022-02-03
Fix deadline warning not showing up

### LINKED ISSUE #3342: Deadline message visible?
This is the second exam session where I try to find a student who sees the message that alerts an approaching deadline (from five minutes before the deadline), but I don't see any student that gets this message displayed on top of the editor. Can you check if this message is still there, and if it is, isn't it to subtle for the students?

chvp:
Since Bootstrap 5, the CSS for `[hidden]` has been changed to `display: none !important`. jQuery's [show](https://api.jquery.com/show/) can't override this, so the `show` and `hide` in the code was useless. Since we're slowly moving away from jQuery anyway, this pull request replaces the jQuery usage with `removeAttribute` and `setAttribute` where appropriate.

I verified locally that the message shows up again. Writing time-based tests seemed like more effort than I was willing to put into it, so I didn't do it.

Closes #3342.

### DIFF
--- a/app/assets/javascripts/exercise.js
+++ b/app/assets/javascripts/exercise.js
@@ -389,17 +389,17 @@ function initExerciseShow(exerciseId, programmingLanguage, loggedIn, editorShown
         if (!_deadline) {
             return;
         }
-        const $deadlineWarning = $("#deadline-warning");
-        const $deadlineInfo = $("#deadline-info");
+        const deadlineWarningElement = document.getElementById("deadline-warning");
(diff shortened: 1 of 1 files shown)


niknetniko (approved):
I have a slight preference for `element.hidden = true / false` over the `set/getAttribute` API, but it's only a small preference, since this also works.

chvp:
> I have a slight preference for `element.hidden = true / false` over the `set/getAttribute` API, but it's only a small preference, since this also works.

I didn't know this is an attribute that can be set in this way. I'll change before merging.

PR #3356, opened 2022-02-07
Fix filtering on copy courses list

### LINKED ISSUE #3351: Filtering courses in course copy is broken
Filtering the course list is broken in the first step of the course copy flow. This is due to the changes in #3277.

```
undefined method `by_filter' for #<Array:0x00007f3c51c5fbe0>
Did you mean?  filter
  app/controllers/courses_controller.rb:40:in `index'
```

jorg-vr:
This pull request fixes the filtering feature on the copy courses list.

I did the ordering with some custom sql this time, which allows `@courses` to stay of the type `ActiveRecord:Relation` instead  of an` array`. (This type difference was the cause of the associated bugs)

Closes #3351  .

### DIFF
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -1,5 +1,3 @@
-require 'will_paginate/array'
-
 class CoursesController < ApplicationController
   include SetLtiMessage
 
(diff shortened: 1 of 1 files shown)


app/controllers/courses_controller.rb
```diff
@@ -31,10 +29,7 @@ def index
     elsif params[:tab] == 'featured'
       @courses = @courses.where(featured: true)
     elsif params[:copy_courses]
-      @courses = @courses.reorder(featured: :desc, year: :desc, name: :asc)
-      @own_courses = @courses.select { |course| current_user.course_admin?(course) }
-      @other_courses = @courses.reject { |course| current_user.course_admin?(course) }
-      @courses = @own_courses.concat(@other_courses)
+      @courses = @courses.reorder(Arel.sql("id in (#{current_user.administrating_courses.pluck(:id).join(', ')}) DESC"), featured: :desc, year: :desc, name: :asc)
```
bmesuere, line 32: Although this seems safe, I'm generally against using raw sql, especially in controllers. This action will not be performed a lot, so cleaner but potentially slower code is worth the tradeoff here.

jorg-vr: Which is why I chose the old option first.

But this causes bugs, because doing the filtering in ruby results in arrays in stead of `ActiveRecordRelation`. 

Some functions later in the code expect `@courses` to be an `ActiveRecordRelation`. 

There was a simple solution to allow pagination on arrays, but allowing scopes seems to be harder. (And doesn't solve the root problem that we are not sure which class `@courses` will have)

I could try to rewrite that part of the code, to split of the `copy_courses` method more from the other course listing, but I doubt that will result in `cleaner` code

bmesuere (approved):

chvp (approved):

PR #3365, opened 2022-02-08
Include build-css in README

winniederidder:
This pull request includes an extra step in the Development Setup section of the README, so that users can obtain the required css-files.


chvp (approved):

bmesuere (commented):
@jorg-vr do we need these separate command? I think running bin/server works and starts everything in watch mode?

jorg-vr:
yes but this also starts the workers, which will only work if the steps later in the readme are applied

(And running `rails s` as the line below says won't work without a manual run of `bin/rake css:build`)

But maybe we should restructure it a bit and guide people to the `bin/server` command before explaining all the separate commands

PR #3366, opened 2022-02-08
Replace webpacker by jsbundling rails with webpack

### LINKED ISSUE #3275: Running ./bin/webpack-dev-server breaks js load
Reported errors are of this form (for all javascript on every page):
```
Started GET "/packs/js/course-c9f406c7729a5b8eb701.js" for 127.0.0.1 at 2022-01-06 09:49:56 +0100
  
ActionController::RoutingError (No route matches [GET] "/packs/js/course-c9f406c7729a5b8eb701.js"):
````

This is caused by the fact that the files in `public/packs/js/` don't match with those mentioned in `public/packs/manifest.json`

Steps to reproduce:
```
./bin/webpack-dev-server
# wait till everything is compiled, then exit
./bin/server
```

Workaround: `rm -rf public/packs/`

### PR DESCRIPTION
**Pull request by jorg-vr**
This pull request replaces webpacker by [jsbundling rails](https://github.com/rails/jsbundling-rails) combined with webpack.

I used this migration guide: https://github.com/rails/jsbundling-rails/blob/main/docs/switch_from_webpacker.md

As far as I understood not all our js was bundled by webpacker. I only updated those part of the code that already used webpacker to webpack.

To do this I did have to rename `application.js` In `app/assets/javascripts` to clarify linking.
I also had to rename `application.js` in`app/javascript/packs` to fix caching issues on naos (which would probably also appear on dodona)

Mapping the `webpacker.yml` file to the new `webpacker.config.js` was not easy as I couldn't find any clear guides and not every setting exists in both specifications.  I could use an extra pair of eyes to see if all important settings are kept.

I did remove the [webpack dev server](https://webpack.js.org/configuration/dev-server/), as i guess this was only used for HMR, which is not supported by jsbundling

Closes #3275

### CODE REVIEW

File: app/assets/config/manifest.js [hunk 1]
```diff
@@ -1,3 +1,3 @@
 //= link_tree ../images
-//= link application.js
+//= link assets_application.js
```
jorg-vr, line 2: As far as I understood this, this file was not build by webpacker, but directly by rails/sprockets?

I had to change the name here because it overlapped with packs/application.js (Which used to have a different linking method but doesn't anymore)

Building the file with webpack doen't work as it only contains comments from the perspective of webpack...

config/webpacker.yml
```diff
@@ -1,98 +0,0 @@
-# Note: You must restart bin/webpack-dev-server for changes to take effect
```
jorg-vr, line 1: Not sure if i mapped all settings in here correctly to webpack.config.js, as not every setting seems to have an equivalent in webpack

app/assets/config/manifest.js
jorg-vr: thanks :)

app/models/concerns/gitable.rb
```diff
@@ -1,3 +1,4 @@
+require 'open3'
```
jorg-vr, line 1: Unrelated bug fix, should check if the `mising open3` issue also appears in `develop` or `rails-7` branch

app/views/series/show.js.erb
```diff
@@ -1,4 +1,4 @@
-<% series_card = render partial: 'series.html.erb',
+<% series_card = render partial: 'series',
```
jorg-vr, line 1: ?Unrelated? bug fix, should check if the the issue of not being able to load series on the course detail page also appears in `develop` or `rails-7` branch

Gemfile
```diff
@@ -13,14 +13,16 @@ gem 'puma', '~> 5.6.1'
...
 gem 'cssbundling-rails', github: 'rails/cssbundling-rails', ref: 'fa6151d'
 
-# Load sprockets ourselves for now => planned fo remove this soon
-gem 'sprockets-rails'
+# Use jsbundling to bundle javascript in app/javascript with webpack
+gem 'jsbundling-rails', '~> 1.0.0'
+
+# Load sprockets ourselves because rails 7 no longer autoloads this
+# This is still used for all javascript in app/assets/javascripts
+gem 'sprockets-rails', '~> 3.4.2'
```
chvp, line 21: Can we/do we want to get rid of sprockets? Seems like the moment to do it if we're messing with our bundling setup anyway.

bin/server
```diff
@@ -1,4 +1,4 @@
 #!/bin/bash
 bundle install
 yarn install
-bundle exec rails s -p 3000 & bundle exec rails jobs:work & yarn build:css --watch
+bundle exec rails s -p 3000 & bundle exec rails jobs:work & yarn build:css --watch &  NODE_ENV=development yarn build:js --watch
```
chvp, line 4:
```suggestion
bundle exec rails s -p 3000 & bundle exec rails jobs:work & yarn build:css --watch & NODE_ENV=development yarn build:js --watch
```

db/schema.rb
```diff
@@ -17,7 +17,7 @@
     t.string "record_type", null: false
     t.bigint "record_id", null: false
     t.bigint "blob_id", null: false
-    t.datetime "created_at", precision: 6, null: false
+    t.datetime "created_at", null: false
```
chvp, line 20: This seems weird?

package.json
```diff
@@ -4,30 +4,44 @@
...
     "glightbox": "^3.1.0",
+    "glob": "^7.2.0",
     "iframe-resizer": "^4.3.2",
     "jquery": "^3.6.0",
     "moment": "^2.29.1",
     "sass": "^1.49.4",
     "typeahead.js": "^0.11.1",
-    "typescript": "^4.5.5"
+    "typescript": "^4.5.5",
+    "webpack-dev-server": "^4.7.4"
```
chvp, line 44: Feels like this should be a devDependency?

niknetniko: It also not clear to me why this is added here, since the PR says:

> I did remove the [webpack dev server](https://webpack.js.org/configuration/dev-server/)[](https://github.com/jorg-vr), as i guess this was only used for HMR, which is not supported by jsbundling

niknetniko (approved):
Besides the existing comments, it seems to work for me

Gemfile
bmesuere: I think sprockets is still used as a last step to (at least) add hashes to the filenames of all assets

jorg-vr: I can look into this in a separate pr?

I thought it would overcomplicate this one as it is a seperate issue from 'moving from webpacker to jsbundling`

package.json
jorg-vr: Well spotted, my mistake, I'll remove it :)

app/models/concerns/gitable.rb
jorg-vr, line 1: Weirdly related bug.

Reproduce run `bundle exec rails db:reset`
Results in:
```
NameError: uninitialized constant Gitable::Open3
Did you mean?  OpenURI
/home/jorg/Documents/dodona/app/models/concerns/gitable.rb:55:in `repo_is_accessible'
```

I could not reproduce this bug on branch `develop` nor on branch `3316-rails-7`. Yet on this branch it happens consistently.

chvp: Maybe webpacker was doing this require somewhere?

chvp (approved):

bmesuere (commented):
Things that should be checked:
- Is transpiling working correctly? It should use the browser versions file in the root of the repo
- Are the polyfills included? We used core-js for this.
- What about rails-ujs? Is this still supported with the switch to hotwire in rails 7 (which we don't want to use). Are the remote links still working?

jorg-vr:
Working on Naos

@bmesuere 
- Tested the pollyfills => They still work
- Not sure how to test the transpiling? Everything seems to be working on different browsers
- Rails ujs is still working, they dont mention any thing about ending support in  their [readme](https://github.com/rails/rails/tree/main/actionview/app/assets/javascripts)

bmesuere (commented):
- .keep file
- application_old
- babel debug check
- chunk plugin

PR #3394, opened 2022-02-15
Check the user institution when logging in from a new provider

### LINKED ISSUE #3318: Add extra check institution check when adding identities to existing users
When a user signs in using an unknown identity, we will try to add the new identity to an existing user (see [here](https://github.com/dodona-edu/dodona/blob/01586464d369834ea5d87861f5176c3756b0170b/app/controllers/auth/omniauth_callbacks_controller.rb#L117) and [here](https://github.com/dodona-edu/dodona/blob/01586464d369834ea5d87861f5176c3756b0170b/app/controllers/auth/omniauth_callbacks_controller.rb#L74)).

Existing users are retrieved using only their email. To prevent unauthorized access to existing dodona accounts from (untrusted) providers, we should only return users from the same institution as the new idp.

Note: this is not an exploitable vulnerability in our code. Users can only be associated with a single institution, so adding an extra identity to an existing user from a different institution will cause a validation failure. It is however better to be more explicit about this in the omniauth controller.

jorg-vr:
This pull request adds an extra check for the user institution when looking for an existing user with a new provider.

Closes #3318 .

### DIFF
--- a/app/controllers/auth/omniauth_callbacks_controller.rb
+++ b/app/controllers/auth/omniauth_callbacks_controller.rb
@@ -119,7 +119,7 @@ class Auth::OmniauthCallbacksController < Devise::OmniauthCallbacksController
     identity = Identity.find_by(identifier: auth_uid, provider: provider)
     return [identity, identity.user] if identity.present? && auth_uid.present?
 
-    # No username was provided, try to find the user using the email address.
+    # No username was provided, try to find the user using the email address and institution id.
     user = User.from_email_and_institution(auth_email, provider.institution_id)
     return [nil, nil] if user.blank?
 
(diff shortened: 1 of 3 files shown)


niknetniko (approved):

app/controllers/auth/omniauth_callbacks_controller.rb
```diff
@@ -120,7 +120,7 @@ def find_identity_and_user
     return [identity, identity.user] if identity.present? && auth_uid.present?
 
     # No username was provided, try to find the user using the email address.
```
chvp, line 122: This comment should probably be updated?

chvp (approved):

PR #3401, opened 2022-02-16
Papyros integration

### LINKED ISSUE #3232: Initial Papyros integration
Once we have a first bundled 0.1 release of dodona-edu/papyros, we can explore integrating this in Dodona.

UI-wise, I would add it to all python exercise pages as a start. I would use a pull-tab design as a trigger: an icon button of which only the left rounded side is visible using position sticky on the right side of the page, somewhere near the bottom.

This would trigger a bootstrap [offcanvas](https://getbootstrap.com/docs/5.1/components/offcanvas/) component to slide out from the right. (hovering over the button might hint at the sliding behaviour by animating to the left) I would use a single column layout with the code on top and the output at the bottom.

We must make sure that pyodide is only loaded after the user clicks on the button for the first time and that it isn't reloaded when the scrashpad is hidden and reopened. In fact, once initialized it should keep running in the background as long the user is on the same page. (We might think about asking to terminate the code when it's still running and the user hides the scratchpad). 

This is only an initial proof of concept integration that will probably require one or more additional papyros releases to get right.

If this works, we can consider releasing this as a beta to gather feedback. Future iterations can add more features such as copying code from and to papyros from the Dodona editor, opening the editor full screen or in a separate tab, including it on more pages, etc. The goal of this issue is **not** to gather potential future ideas, but to track the initial integration.

winniederidder:
This pull request integrates Papyros into Dodona to allow users to write and run code within their browser.

Papyros makes use of WebWorkers, requiring SharedArrayBuffers to communicate. Enabling this feature in the browser requires specific HTTP headers to be set on resources used in the pages where this feature is enabled. Examples of such resources are /packs and /assets. For production/staging, these headers must be enabled on the server. In development, this can be taken care of with configuration in config/development.rb .
Some HTTP headers must also be added to responses concerning exercises and their description to allow the iframes to load everything correctly and securely.

- [ ] Tests were added

Closes #3232  .
Depends on #3366


winniederidder:
Current visuals of the integration:

app/assets/javascripts/exercise.js
```diff
@@ -2,6 +2,7 @@
 import { initTooltips, updateURLParameter } from "util.js";
 import { Toast } from "./toast";
 import GLightbox from "glightbox";
+import { InputMode, Papyros, plFromString } from "@dodona/papyros";
```
bmesuere, line 5: `plFromString` seems like this should be a static function under Papyros or something instead of a separate export since it is not useful outside this context.

app/assets/javascripts/exercise.js
```diff
@@ -148,6 +149,9 @@ function initExerciseShow(exerciseId, programmingLanguage, loggedIn, editorShown
     let lastSubmission;
     let lastTimeout;
 
+    let papyros;
+    let launched = false;
```
bmesuere, line 153: rename to something useful, probably papyrosLaunched or papyrosActive?

app/assets/javascripts/exercise.js
```diff
@@ -183,6 +187,55 @@ function initExerciseShow(exerciseId, programmingLanguage, loggedIn, editorShown
...
+                const initialCode = editor.getValue();
+                if (initialCode) {
+                    papyros.setCode(initialCode);
+                }
+            });
+
+            $("#papyros-code-copy-btn").on("click", function () {
+                editor.setValue(papyros.getCode());
+            });
+        } catch (e) {
```
bmesuere, line 232: Error handling for flow control is bad style. But part of this will probably be fixed by conditionally calling the papyros initialisation from ruby-land

app/assets/stylesheets/components/offcanvas.css.scss
```diff
@@ -0,0 +1,3 @@
+.papyros-offcanvas {
+    width: 700px;
```
bmesuere, line 2: I would prefer overriding the bootstrap variable for this (https://getbootstrap.com/docs/5.1/components/offcanvas/#variables).

app/controllers/activities_controller.rb
```diff
@@ -177,6 +194,14 @@ def media
     end
   end
 
+  # Serve the inputServiceWorker file required to handle input in Papyros
+  # Could potentially changed to a Dodona-specific service worker importing a Papyros-function
+  def isw
```
bmesuere, line 199: this needs a better name

app/controllers/activities_controller.rb
```diff
@@ -177,6 +194,14 @@ def media
...
+    send_file(Rails.root.join('node_modules', '@dodona', 'papyros', 'dist', 'inputServiceWorker.js'),
```
bmesuere, line 200: This feels very ugly. In theory, the node_modules directory shouldn't even be available at runtime.

app/views/activities/show.html.erb
```diff
@@ -129,6 +132,30 @@ end %>
           <div class="tab-pane fade hidden" id="feedback">
             <div id="submission-wrapper"></div>
           </div>
+
```
bmesuere, line 135: Move this code to a partial since there's a chance is will be reused on other pages such as grading.

config/initializers/content_security_policy.rb
```diff
@@ -14,11 +14,15 @@
       # Allow webpack-dev-server
       policy.connect_src :self, Rails.configuration.tutor_url.to_s,
                          'https://*.googleapis.com',
-                         'http://localhost:3035', 'ws://localhost:3035'
+                         'http://localhost:3035', 'ws://localhost:3035',
```
bmesuere, line 17: why is localhost used here?

bmesuere (changes requested):
I would move the papyros initialization to a separate function instead of integrating it in initExerciseShow because we won't always run this code.

Layout:
- The button to trigger papyros should not be a floating action button. What about something like https://cdn.phpjabbers.com/files/frees/sliding-contact-form.jpg? (I wouldn't show the button when expanded)
- I wouldn't use "Papyros" branding in the editor. The title should be something like "Run Python code in your browser" so it's clear what this is. Maybe add a BETA label?
- The help text should be bigger, especially since it's still in beta. Tweak the text so it's more aimed at something younger children can understand. Link them to the contact form instead of just saying to contact us. The margin should be in line with the rest of the page. The top needs a bit more spacing.
- Run and stop buttons feel a bit off here design-wise. I don't have a clear suggestion at this time.
- The vertical alignment for "waiting for input" if off.
- It's not really clear what the output area is when you load the page now that the title is gone. Maybe add a "your output will appear here" placeholder?
- The output area is too prominent compared to the editor itself. What about reducing the height a bit + increase the height of the editor area by giving it the leftover vertical space instead of a fixed height?
- The copy button at the bottom isn't clear. There is enough space so at least include a label

@rien can you check the CSP header changes?

app/controllers/activities_controller.rb
winniederidder: That is correct. A possible solution might be as follows:
I could try to make Papyros search for the serviceworker at a given path (using configureInput). We could then move the isw-path to a different controller (pages is probably not the most appropriate, but something like that) and give it a suitable name (something related to service_worker). Finally, I could add an installation script that copies the default service worker to a given path (such as the public directory or from wherever the application wants to serve the service worker if it doesn't provide one themselves).

config/initializers/content_security_policy.rb
winniederidder: Localhost was already present before this PR. I believe it has to do with the webpack-dev-server, as mentioned in the comment above. Someone else might be able to shed more light on this specifically (@rien , @jorg-vr ?)

app/controllers/activities_controller.rb
winniederidder: > I would move the papyros initialization to a separate function instead of integrating it in initExerciseShow because we won't always run this code.
> 

That method won't always be called, because you mentioned e.g. using it in the Grading page? Would a separate method that can be given an editor argument (in case of an exercise page) work then? Being able to access the ace-editor adds useful functionality which I would like to preserve.

> Layout:

> * Run and stop buttons feel a bit off here design-wise. I don't have a clear suggestion at this time.

I could move the buttons away from the status bar and put them elsewhere (similar to [futurecoder](https://futurecoder.io/course/#ide) or [Codecademy](https://i0.wp.com/learn.onemonth.com/wp-content/uploads/2019/05/image1.png?resize=1030%2C532&ssl=1)). [This issue](https://github.com/dodona-edu/papyros/issues/36) also describes the addition of keybinds which can replace the buttons, but students might prefer also having physical buttons in a good location.


> * It's not really clear what the output area is when you load the page now that the title is gone. Maybe add a "your output will appear here" placeholder?

The title attribute is used (shown when the user hovers over the output field). We can also add a title ourselves above the respective div, or go back to the placeholder.

> * The copy button at the bottom isn't clear. There is enough space so at least include a label

Is a floating action button okay here, or should the button style itself also be changed?

winniederidder:
> I would move the papyros initialization to a separate function instead of integrating it in initExerciseShow because we won't always run this code.
> 

That method won't always be called, because you mentioned e.g. using it in the Grading page? Would a separate method that can be given an editor argument (in case of an exercise page) work then? Being able to access the ace-editor adds useful functionality which I would like to preserve.

> Layout:

> * Run and stop buttons feel a bit off here design-wise. I don't have a clear suggestion at this time.

I could move the buttons away from the status bar and put them elsewhere (similar to [futurecoder](https://futurecoder.io/course/#ide) or [Codecademy](https://i0.wp.com/learn.onemonth.com/wp-content/uploads/2019/05/image1.png?resize=1030%2C532&ssl=1)). [This issue](https://github.com/dodona-edu/papyros/issues/36) also describes the addition of keybinds which can replace the buttons, but students might prefer also having physical buttons in a good location.


> * It's not really clear what the output area is when you load the page now that the title is gone. Maybe add a "your output will appear here" placeholder?

The title attribute is used (shown when the user hovers over the output field). We can also add a title ourselves above the respective div, or go back to the placeholder.

> * The copy button at the bottom isn't clear. There is enough space so at least include a label

Is a floating action button okay here, or should the button style itself also be changed?

winniederidder:
Some late screenshots:
Button to show Papyros
papyros-button
Current layout with new button, beta label, ...
papyros-shown

winniederidder:
Closed in favor of #3453

PR #3402, opened 2022-02-16
Replace use of ActiveRecord::Associations::Preloader by includes

### LINKED ISSUE #3369: Usage of ActiveRecord::Associations::Preloader should be removed from dodona
After the upgrade to rails 7 (#3341), the api of `ActiveRecord::Associations::Preloader` changed. [old](https://www.rubydoc.info/docs/rails/ActiveRecord/Associations/Preloader) [new](https://msp-greg.github.io/rails_stable/ActiveRecord/Associations/Preloader.html)

New api docs state "Do not use. This class is for internal use only." so the api of this class will probably not be guaranteed to be stable in the future.

jorg-vr:
This pull request replaces the usage of ActiveRecord::Associations::Preloader by includes.

It was introduced in https://github.com/dodona-edu/dodona/commit/e2cfa5f8120157fff147c3a1d933dc5acb031ec4 but the API of preloader is not guaranteed (as we noticed with a recent api change see #3369). 

Our expected usage is also supported by the default rails api, example see here: https://stackoverflow.com/questions/38749960/rails-eager-load-deep-associations-on-an-instance

Closes #3369 .

### DIFF
--- a/app/controllers/score_items_controller.rb
+++ b/app/controllers/score_items_controller.rb
@@ -93,6 +93,6 @@ class ScoreItemsController < ApplicationController
     # Reload the evaluation, since one of the items changed.
     @evaluation.score_items.reload
     # Preload the stuff for one exercise and an existing instance.
-    EvaluationExercise.includes({score_items: :scores}).find(item.evaluation_exercise.id)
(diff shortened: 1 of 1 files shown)


niknetniko (approved):
(Except the complaint by rubocop)

PR #3403, opened 2022-02-16
Don't fetch first and last submission for timeseries if none are present

### LINKED ISSUE #3367: Timeseries fails on empty data
Timeseries data prep sometimes throws an error:

```
A NoMethodError occurred in statistics#timeseries:

  undefined method `created_at' for nil:NilClass
  app/models/submission.rb:432:in `timeseries_matrix'

-------------------------------
Request:
-------------------------------

  * URL        : https://dodona.ugent.be/nl/stats/timeseries?series_id=12607

```

jorg-vr:
This pull request fixes a bug when timeseries tried to fetch the first and last submission when none are present.

Now a message is shown explaining that no data is availlable:
image

Closes #3367 .

### DIFF
--- a/app/models/submission.rb
+++ b/app/models/submission.rb
@@ -429,8 +429,8 @@ class Submission < ApplicationRecord
     submissions = submissions.judged
     submissions = submissions.from_students(options[:series].course)
 
-    first_sub = submissions.least_recent.first.created_at
-    last_sub = submissions.most_recent.first.created_at
+    first_sub = submissions.present? ? submissions.least_recent.first.created_at : nil
+    last_sub = submissions.present? ? submissions.most_recent.first.created_at : nil
 
     submissions = submissions.in_time_range(options[:start], options[:end]) if options[:end].present?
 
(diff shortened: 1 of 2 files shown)


app/models/submission.rb
```diff
@@ -429,8 +429,8 @@ def self.timeseries_matrix(options = {})
     submissions = submissions.judged
     submissions = submissions.from_students(options[:series].course)
 
-    first_sub = submissions.least_recent.first.created_at
-    last_sub = submissions.most_recent.first.created_at
+    first_sub = submissions.present? ? submissions.least_recent.first.created_at : nil
+    last_sub = submissions.present? ? submissions.most_recent.first.created_at : nil
```
chvp, line 433: I feel like `.any?` instead of `.present?` makes more sense semantically?

chvp (changes requested):
With small bugfixes like this, I always try to add a simple test covering the wrong behaviour. Just to make sure we don't regress later.

chvp (approved):

bmesuere (approved):

niknetniko (approved):

PR #3404, opened 2022-02-16
Adjust color scale of timeseries graph if the time period is updated

### LINKED ISSUE #3352: Heatmap color scale isn't updated
In the new heatmap graph, the color scale isn't updated when changing the time period. This causes all squares to be very light in color.

image

jorg-vr:
This pull request adjusts the color scale of timeseries graph if the time period is updated.

Closes #3352  .


chvp (approved):

bmesuere (approved):
I agree with the fix, but the same 2 comments on the seeds as on #3405

jorg-vr:
Will remove the seeding change here and fix it in https://github.com/dodona-edu/dodona/pull/3405

PR #3405, opened 2022-02-17
Clear scores when another submission is selected to grade

### LINKED ISSUE #3344: Scores should be cleared when selecting another submission in an evaluation
Especially relevant when students who received more time didn't submit before the usual deadline: they have 0 in every category.

jorg-vr:
This pull request clears the scores when an other submission is selected to grade.

Peek 2022-02-17 13-37

- [x] Tests were added

Closes #3344 .

### DIFF
--- a/app/controllers/feedbacks_controller.rb
+++ b/app/controllers/feedbacks_controller.rb
@@ -67,11 +67,7 @@ class FeedbacksController < ApplicationController
     attrs = permitted_attributes(@feedback)
     attrs['scores_attributes'].each { |s| s['last_updated_by_id'] = current_user.id } if attrs['scores_attributes'].present?
 
-    # Reset scores if a new exercise is chosen
-    @feedback.scores.each { |s| s.destroy } if attrs['submission_id'].present? && attrs['submission_id'] != @feedback.submission_id
-
     updated = @feedback.update(attrs)
-
     # We might have updated scores, so recalculate the map.
     @score_map = @feedback.scores.index_by(&:score_item_id)
 
--- a/app/controllers/feedbacks_controller.rb
+++ b/app/controllers/feedbacks_controller.rb
@@ -67,7 +67,11 @@ class FeedbacksController < ApplicationController
     attrs = permitted_attributes(@feedback)
     attrs['scores_attributes'].each { |s| s['last_updated_by_id'] = current_user.id } if attrs['scores_attributes'].present?
 
+    # Reset scores if a new exercise is chosen
+    @feedback.scores.each { |s| s.destroy } if attrs['submission_id'].present? && attrs['submission_id'] != @feedback.submission_id
+
     updated = @feedback.update(attrs)
+
     # We might have updated scores, so recalculate the map.
     @score_map = @feedback.scores.index_by(&:score_item_id)
 
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -272,7 +272,7 @@ if Rails.env.development?
       s.exercises << series_exercises
       series_exercises.each do |exercise|
         course.enrolled_members.sample(5).each do |student|
-          (1..rand(10)).each do
+          (1..rand(1..10)).each do
             status = if rand() < 0.5
                        :correct
                      else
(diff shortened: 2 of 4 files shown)


db/seeds.rb
```diff
@@ -272,21 +272,24 @@ def submission_summary(status)
...
-                            course: s.course,
-                            exercise: exercise,
-                            evaluate: false,
-                            skip_rate_limit_check: true,
-                            status: status,
-                            accepted: status == :correct,
-                            summary: submission_summary(status),
-                            code: "print(input())\n",
-                            result: File.read(Rails.root.join('db', 'results', "#{exercise.judge.name}-result.json"))
+          (1..rand(10)).each do
```
niknetniko, line 275: Wouldn't this create no submissions if `rand` returns 0?

niknetniko (changes requested):

app/controllers/feedbacks_controller.rb
```diff
@@ -67,7 +67,11 @@ def update
     attrs = permitted_attributes(@feedback)
     attrs['scores_attributes'].each { |s| s['last_updated_by_id'] = current_user.id } if attrs['scores_attributes'].present?
 
+    # Reset scores if a new exercise is chosen
+    @feedback.scores.each { |s| s.destroy } if attrs['submission_id'].present? && attrs['submission_id'] != @feedback.submission_id
```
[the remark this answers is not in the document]
jorg-vr: Good idea, will edit :)

db/seeds.rb
jorg-vr: could happen, although that is not really an issue. The random is meant to create some variability in the amount of submissions. But will update to `rand(1,10)`

niknetniko (approved):
:+1: (apart from Rubocop's complaint)

chvp (approved):

db/seeds.rb
```diff
@@ -272,21 +272,24 @@ def submission_summary(status)
...
-                            course: s.course,
-                            exercise: exercise,
-                            evaluate: false,
-                            skip_rate_limit_check: true,
-                            status: status,
-                            accepted: status == :correct,
-                            summary: submission_summary(status),
-                            code: "print(input())\n",
-                            result: File.read(Rails.root.join('db', 'results', "#{exercise.judge.name}-result.json"))
+          (1..rand(1..10)).each do
```
bmesuere, line 275: Do we really need this many submissions? I know this seems like a small number, but they quickly add up, especially in nested loops. After a while our tests and db seeds (and thus mestra deploys) become painfully slow.

db/seeds.rb
```diff
@@ -272,21 +272,24 @@ def submission_summary(status)
...
+                     end
+            Submission.create user: student,
+                              course: s.course,
+                              exercise: exercise,
+                              evaluate: false,
+                              skip_rate_limit_check: true,
+                              status: status,
+                              accepted: status == :correct,
+                              summary: submission_summary(status),
+                              created_at: (Time.now - 1.day).beginning_of_day + 6.hour + rand(0..540).minutes + rand(0..540).minutes,
```
bmesuere, line 289: What is the goal if this construction? If there is a specific reason for this time range, maybe add a comment with the explanation.

Next to that, it can be cleaned up. The 6.hours and double rand can be replace by a single rand. The first part can probably be replaced by something like Date.yesterday or 1.day.ago.

bmesuere (approved):
I suppose this will also fix the issue reported in [this comment](https://github.com/dodona-edu/dodona/discussions/3400#discussioncomment-2187874)?

db/seeds.rb
jorg-vr: Created a separate course for visualizations. Because yes we need so many for decent graphs and testing features with multiple submission per student for example, but no need to have it in every course

db/seeds.rb
jorg-vr: I improved the date and added a comment for the other part. Using two rands creates a bell curve, so it is not the same result as a single rand

PR #3407, opened 2022-02-18
 Improve labeling of time periods on cumulative plot

### LINKED ISSUE #3348: Improve labeling of time periods on cumulative plot
I guess the cumulative plots have an off-by-one issue when results are binned per 12 hours. It's just a labelling issue but the period that is for example marked as (12:00-0:00 Tuesday September 28) actually is the second half of the day Monday September 27 (only the 0:00 minute is on Tuesday September 28). I don't know if anything is counted double at the 0:00 and 12:00 at the moment, but at least I would suggest to have time periods of 0:00-11:59 and 12:00-23:59 in this case (instead of 0:00-12:00 and 12:00-00). I was actually confused because at first it seems as if the largest group of submissions came on Tuesday and Saturday, whereas we have hands-on sessions on Mondays and Fridays, which are traditionally the periods where a lot of submissions are made. But nothings wrong with the data itself, it's only an issue of the labelling: 12:00-00 Tuesday September 28 is actually Monday afternoon (when the very first hands-on session happened).

image

image

jorg-vr:
This pull request removes the start time from the tooltip of the cumulative plot as it incorrectly indicates some kind of binning behavior, which is not what the graph represents.

The capitalization of dates in the tooltip has also changed to follow behavior as specified by I18n.

image
image

I also fixed a bug that scaling is reset when the date range is changed.

Closes #3348 .

### DIFF
--- a/app/assets/javascripts/visualisations/cumulative_timeseries.ts
+++ b/app/assets/javascripts/visualisations/cumulative_timeseries.ts
@@ -362,6 +362,7 @@ export class CTimeseriesGraph extends SeriesGraph {
             message += `<br><span style="color: ${this.color(e)}">&FilledSmallSquare;</span> ${d3.format(".1%")(ex.ex_data[i].cSum / this.studentCount)}
                     (${ex.ex_data[i].cSum}/${this.studentCount})`;
         });
+        message += "</b>";
         return message;
     }
 
--- a/app/assets/javascripts/visualisations/cumulative_timeseries.ts
+++ b/app/assets/javascripts/visualisations/cumulative_timeseries.ts
@@ -345,7 +345,7 @@ export class CTimeseriesGraph extends SeriesGraph {
     private tooltipText(i: number): string {
         const date = this.binTicks[i];
         const capitalize = (s: string): string => s.charAt(0).toUpperCase() + s.slice(1);
-        let message = "<b>";
+        let message = "";
 
         if (this.binStep < 24) {
             const on = I18n.t("js.date_on");
@@ -362,8 +362,8 @@ export class CTimeseriesGraph extends SeriesGraph {
             message += `<br><span style="color: ${this.color(e)}">&FilledSmallSquare;</span> ${d3.format(".1%")(ex.ex_data[i].cSum / this.studentCount)}
                     (${ex.ex_data[i].cSum}/${this.studentCount})`;
         });
-        message += "</b>";
-        return message;
+
+        return "<b>" + message + "</b>";
     }
 
     private yRangeToggle(): void {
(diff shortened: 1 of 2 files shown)


chvp:
As I mentioned on Slack, I think removing the start time makes more sense. The values in the graph are calculations at fixed points in time, not in intervals.

jorg-vr:
Updated the behavior, it is indeed much more logical this way :)

app/assets/javascripts/visualisations/cumulative_timeseries.ts
```diff
@@ -343,48 +344,25 @@ export class CTimeseriesGraph extends SeriesGraph {
...
+
+        const weekDay = d3.timeFormat(I18n.t("date.formats.weekday_long"));
+        message += capitalize(weekDay(date));
+
         this.exOrder.forEach(e => {
             const ex = this.data.find(ex => ex.ex_id === e);
             message += `<br><span style="color: ${this.color(e)}">&FilledSmallSquare;</span> ${d3.format(".1%")(ex.ex_data[i].cSum / this.studentCount)}
                     (${ex.ex_data[i].cSum}/${this.studentCount})`;
         });
+        message += "</b>";
```
bmesuere, line 365: I would prefer that the opening b-tag is also added here (so something like `message = "<b>" + message + "</b>"` to reduce the change to generate invalid html by some future change.

app/assets/javascripts/visualisations/cumulative_timeseries.ts
```diff
@@ -343,48 +344,25 @@ export class CTimeseriesGraph extends SeriesGraph {
...
-            endMessage = i ?
-                monthDay(date):
-                weekDay(date);
+            message += timeFormat(date);
+            message += ` ${on} `;
         }
-        let message = `<b>${startMessage} ${endMessage} ${extraMessage}`;
+
+        const weekDay = d3.timeFormat(I18n.t("date.formats.weekday_long"));
+        message += capitalize(weekDay(date));
```
bmesuere, line 358: The first letter of days and months is only capitalized in English and not in Dutch. In addition, the capitalize function is added twice in the graph files, so maybe the fixed version that checks the current language can me moved to the super class.

bmesuere (changes requested):
The tooltip look a lot clearer 👍 
Sorry for nitpicking!

app/assets/javascripts/visualisations/cumulative_timeseries.ts
jorg-vr: I just removed the function as the behavior of I18n was already correct

bmesuere (approved):

PR #3426, opened 2022-02-21
Improve color scheme of dark mode for the  cumulative time series and heatmap

jorg-vr:
This pull request improves the color scheme of dark mode for the cumulative time series and heatmap.

I made this change because in the old color scheme I could not intuitively see which values where highest.

New:
image
image

Old:
image
image

Light mode:
image
image


bmesuere (approved):
Looks clearer indeed!

niknetniko (approved):

pdawyndt:
Maybe you could also give the boxes a darker border instead of the blue border.

pdawyndt:
The same issue actually also holds for the punch card in the user profile and the course stats page: differences are clearer in light mode compared to dark mode. Maybe the same changes can be made there (if they aren't changed by this PR already).

chvp (approved):

PR #3453, opened 2022-03-01
Papyros integration

### LINKED ISSUE #3232: Initial Papyros integration

winniederidder:
This pull request integrates Papyros into Dodona to allow users to write and run code within their browser.

Papyros makes use of WebWorkers, requiring SharedArrayBuffers to communicate. Enabling this feature in the browser requires specific HTTP headers to be set on resources used in the pages where this feature is enabled. Examples of such resources are /packs and /assets. For production/staging, these headers must be enabled on the server. In development, this can be taken care of with configuration in config/development.rb .
Some HTTP headers must also be added to responses concerning exercises and their description to allow the iframes to load everything correctly and securely.

- [x] Tests were added

Closes #3232  .


winniederidder:
Layout:

Button to show Papyros, shown on permitted exercise pages
papyros-button
Current layout with new button, beta label, ...
papyros-shown

app/assets/javascripts/coding_scratchpad.js
```diff
@@ -0,0 +1,55 @@
+import { InputMode, Papyros } from "@dodona/papyros";
+
+function initCodingScratchpad(programmingLanguage, editor = undefined) {
```
bmesuere, line 3: what are you trying to solve with `editor = undefined`? Default parameters are used if no value or undefined is passed, so this seems like a no-op.

app/assets/javascripts/coding_scratchpad.js
```diff
@@ -0,0 +1,55 @@
...
+        let papyrosLaunched = false;
+        const papyros = Papyros.fromElement(
+            {
+                programmingLanguage: Papyros.toProgrammingLanguage(programmingLanguage),
+                standAlone: false,
+                locale: I18n.locale,
+                inputMode: InputMode.Interactive,
+            }, {
+                code: {
+                    parentElementId: "papyros-editor-wrapper",
```
bmesuere, line 14: these id strings should probably be consts.

app/assets/javascripts/coding_scratchpad.js
```diff
@@ -0,0 +1,55 @@
...
+                    parentElementId: "papyros-output-wrapper",
+                    attributes: new Map([["style", "max-height: 28vh;"]])
+                },
+                input: {
+                    parentElementId: "papyros-input-wrapper"
+                }
+            }
+        );
+
+        $("#papyros-offcanvas-show-btn").on("click", async function () {
```
bmesuere, line 30: we are deprecating the use of jquery in Dodona, so new code should use a native equivalent.

app/assets/javascripts/coding_scratchpad.js
```diff
@@ -0,0 +1,55 @@
...
+            if (!papyrosLaunched) {
+                await papyros.configureInput(false, "http://dodona.localhost:3000/", "inputServiceWorker.js");
```
bmesuere, line 32: why is there a local host reference here?

app/assets/javascripts/exercise.js
```diff
@@ -182,6 +183,7 @@ function initExerciseShow(exerciseId, programmingLanguage, loggedIn, editorShown
         $(".activity-description a[target='_blank']").each(function () {
             $(this).attr("rel", "noopener");
         });
+        initCodingScratchpad(programmingLanguage, editor);
```
bmesuere, line 186: This line should probably be called from an erb/ruby page instead of including it here. In fact, it is already added as a separate pack, so including it here at the top of the file will including it twice in our assets.

app/assets/stylesheets/bootstrap_variable_overrides.css.scss
```diff
@@ -34,3 +34,6 @@ $yellow:  $yellow-500;
 $green:   $green-500;
 $teal:    $teal-500;
 $cyan:    $cyan-500;
+
+// bootstrap offcanvas overrides
+$offcanvas-horizontal-width: 700px;
```
bmesuere, line 39: the offcanvas seems to have a white left border. This is not the case with the examples on the bootstrap website.

app/assets/stylesheets/components/offcanvas.css.scss
```diff
@@ -0,0 +1,11 @@
+.offcanvas-show-btn {
```
bmesuere, line 1: rename the file to papyros/scratchpad instead of offcanvas.

app/assets/stylesheets/libraries/coding_scratchpad.css.scss
```diff
@@ -0,0 +1,25 @@
+.margin-bottom-11 {
```
bmesuere, line 1: If these styles are added to the libraries folder, they should probably be prefixed or scoped since margin-bottom and beta-label are quite generic in name.

app/assets/stylesheets/components/offcanvas.css.scss
```diff
@@ -0,0 +1,11 @@
...
+    position: fixed;
+    left: 100%;
+    top: 50%;
```
bmesuere, line 4: I prefer it a bit lower on the page, eg bottom: 100px

app/assets/stylesheets/libraries/coding_scratchpad.css.scss
```diff
@@ -0,0 +1,25 @@
...
+    font-size: 14px;
+    color: $text-muted;
+}
+
+.papyros-header {
+    color: $header-fg;
+    background-color: $header-bg;
+}
+
+.beta-label {
```
bmesuere, line 16: I prefer a more muted approach, e.g. BETA in superscript just to the right of the title.

app/assets/stylesheets/libraries/coding_scratchpad.css.scss
```diff
@@ -0,0 +1,25 @@
...
+    background-color: rgb(255, 0, 106);
+    color: white;
```
bmesuere, line 18: If still needed, replace with existing scss variables

app/controllers/activities_controller.rb
```diff
@@ -177,6 +194,15 @@ def media
     end
   end
 
+  # Serve the inputServiceWorker file required to handle input in Papyros
+  # Could potentially changed to a Dodona-specific service worker importing a Papyros-function
+  # Must be updated when a change happens to the default worker in Papyros
+  def isw
+    send_file(Rails.root.join('app/assets/javascripts/inputServiceWorker.js'),
```
bmesuere, line 201: the files in app/assets are processed through webpack/sprockets/... for serving in production. It is not ok to reference it directly like this in the ruby code.

app/models/exercise.rb
```diff
@@ -80,6 +80,10 @@ def file_extension
     programming_language&.extension || 'txt'
   end
 
+  def scratchpad?
+    programming_language&.name == 'python'
+  end
+
```
bmesuere, line 86: In my opinion, loading the scratchpad (or not) should not be part of the model, so remove this code.

app/views/activities/_coding_scratchpad.html.erb
```diff
@@ -0,0 +1,39 @@
...
+            <div class="flex-spacer"></div>
+            <span class="beta-label">
+                <%= t '.beta' %>
+            </span>
+            <button type="button" class="btn-close btn-close-white" 
+            data-bs-dismiss="offcanvas" aria-label="<%= t '.close' %>">
+            </button>
+        </div>
+    </div>
+    <div class="papyros-info">
```
bmesuere, line 25: There is still too little margin/padding. The info button almost sticks to the menu bar and the left margin is only half of that of the rest of the content.

app/views/activities/_coding_scratchpad.html.erb
```diff
@@ -0,0 +1,39 @@
...
+        <i class="mdi mdi-information mdi-18"></i>
+        <%= t ".experimental_feature_html", :url => contact_path %>
+    </div>
+    <div class="offcanvas-body">
+        <div id="papyros-editor-wrapper"></div>
+        <div id="papyros-panel-wrapper"></div>
+        <div id="papyros-output-wrapper"></div>
+        <div id="papyros-input-wrapper"></div>
+        <button class="btn with-text btn-primary btn-fab-extended">
```
bmesuere, line 34: The added text is better, but we still have a mix of 3 buttons (4 of you count enter) each in a different color and 3 in a different style. Can we add the copy button to the right of the stop button?

app/views/activities/show.html.erb
```diff
@@ -129,6 +129,9 @@ end %>
           <div class="tab-pane fade hidden" id="feedback">
             <div id="submission-wrapper"></div>
           </div>
+          <% if @activity.scratchpad? %>
```
bmesuere, line 132: explicitly check for python here (instead of in the model). In a first version, this check will probably include some user id's as well.

app/views/activities/_coding_scratchpad.html.erb
```diff
@@ -0,0 +1,39 @@
...
+            <i class="mdi mdi-content-copy icon"></i>
+            <span class="text"><%= t '.copy_code' %></span>
+        </button>
+    </div>
+</div>
```
bmesuere, line 39: The javascript initialization of the scratchpad should happen here.

app/views/activities/show.html.erb
```diff
@@ -129,6 +129,9 @@ end %>
...
+            <%= render partial: 'coding_scratchpad', locals: {activity: @activity} %>
+          <% end %>
```
bmesuere, line 134: These elements don't follow the flow of the page, but are inserted at a very specific place (i.e. in the tab content). Why not add them at the bottom of this file?

bmesuere (changes requested):
@rien kan you check the policy stuff?

app/assets/javascripts/inputServiceWorker.js
```diff
@@ -0,0 +1 @@
...ne)return o;switch(n=0,o&&(i=[2&i[0],o.value]),i[0]){case 0:case 1:o=i;break;case 4:return a.label++,{value:i[1],done:!1};case 5:a.label++,n=i[1],i=[0];continue;case 7:i=a.ops.pop(),a.trys.pop();continue;default:if(!((o=(o=a.trys).length>0&&o[o.length-1])||6!==i[0]&&2!==i[0])){a=0;continue}if(3===i[0]&&(!o||i[1]>o[0]&&i[1]<o[3])){a.label=i[1];break}if(6===i[0]&&a.label<o[1]){a.label=o[1],o=i;break}if(o&&a.label<o[2]){a.label=o[2],a.ops.push(i);break}o[2]&&a.ops.pop(),a.trys.pop();continue}i=t.call(e,a)}catch(e){i=[6,e],n=0}finally{r=o=0}if(5&i[0])throw i[1];return{value:i[0]?i[1]:void 0,done:!0}}([i,s])}}}(this,(function(t){switch(t.label){case 0:return[4,o.handleInputRequest(e)];case 1:return t.sent()||e.respondWith(fetch(e.request)),[2]}}))},new((n=void 0)||(n=Promise))((function(e,o){function a(e){try{u(i.next(e))}catch(e){o(e)}}function s(e){try{u(i.throw(e))}catch(e){o(e)}}function u(t){var r;t.done?e(t.value):(r=t.value,r instanceof n?r:new n((function(e){e(r)}))).then(a,s)}u((i=i.apply(t,r||[])).next())}));var t,r,n,i})),addEventListener("install",(function(e){e.waitUntil(skipWaiting())})),addEventListener("activate",(function(e){e.waitUntil(clients.claim())}))})(),n})()}));
```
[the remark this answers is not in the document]
winniederidder: This is the default service worker from Papyros, as we do not provide one within Dodona. It is built using Webpack and copied from node_modules/@dodona/papyros/dist. I picked a location in our assets directory that seemed appropriate, but it must be somewhere where we can find and serve it, without having to resort to node_modules directly.

app/assets/javascripts/coding_scratchpad.js
```diff
@@ -0,0 +1,55 @@
+import { InputMode, Papyros } from "@dodona/papyros";
```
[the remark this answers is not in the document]
winniederidder: I followed the style I found in exercise.js, where I initially implemented my solution. All files there were in JavaScript, so I stuck to that. I will migrate it.

app/assets/javascripts/coding_scratchpad.js
winniederidder: This is to make it clear that the editor argument is optional. Otherwise, a reader of the code might think that both arguments are always required. When migrating to TypeScript, I will be able to use editor? instead.

app/assets/javascripts/coding_scratchpad.js
winniederidder: An artefact from testing, this should indeed become a dynamic url.

app/assets/javascripts/exercise.js
winniederidder: How do I obtain a reference to the editor then? This editor is scoped within that function, so it must be called from there. Is there a different way to get this required reference?

app/views/activities/_coding_scratchpad.html.erb
winniederidder: As mentioned in an earlier comment, having access to the ace editor (when it exists) is quite useful. Currently, it is locally scoped inside exercise.js. Should I instantiate a separate instace of it using ace.edit on the submission field if it exists within the page, or what is your preferred approach?

app/assets/javascripts/inputServiceWorker.js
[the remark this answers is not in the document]
bmesuere: This is definitely not ok.
- The way this file is accessed from the controller is not ok
- It will not be automatically updated when we update papyros 
- it is not clear what this file is doing. Especially because a service worker can break the rest of Dodona and will be hard to fix if that's the case.

If there isn't an acceptable way to fix these issues, I once again propose not using the service worker route.

app/assets/javascripts/exercise.js
bmesuere: This is why we have a global dodona object

app/assets/javascripts/inputServiceWorker.js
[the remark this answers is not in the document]
winniederidder: It is stored within the assets folder. I think assets are served automatically somehow but my Rails knowledge is rather limited. In theory, the ActivitiesController method is not required as it can be served like a stylesheet and does not need any modification. 

Automatic updating is indeed not done. This is also mentioned in the Papyros-README where users will need to manually copy this file from the dist folder if they do not provide a service worker themselves. However, at this point the file is stable so this will rarely be an issue.
What the file is doing, is described in that README as well. It appends headers to requests that belong to the scope of the service worker, which can be made even smaller by giving it its own directory (so it should be moved a bit). It also handles the specific input urls which are not used by standard applications and thus do not interfere. The added HTTP headers to other parts of Dodona should ensure everything keeps working, as is mentioned in the README in Papyros. It is simply built by webpack and thus turned into rather ugly JavaScript, similar to how e.g. bootstrap.min.js or other minified JS files work. 

So, the first issue should be resolvable by serving it as a normal asset.
The other issues I explained above.

If this is not sufficient, we could make our own service worker importing the functionality provided within Papyros. This will ensure automatic updating when installing a new version, as it will be re-bundled. It will be more clear what it does, as programmers don't see the uglified code. It will need a bundling step to make it accessible in the way it is now, so this will perhaps require some more delving in the assets pipeline.

app/assets/javascripts/exercise.js
winniederidder: I should add the editor to the global object then within initExerciseShow? Do we not risk a race condition if the scratchpad is initialized before initExerciseShow is called? Alternatively, an old instance of the editor might be present when switching between exercises, which could also cause problems. Are there guarantees on the order of code executed?

app/controllers/activities_controller.rb
winniederidder: Am I allowed to reference app/assets/builds?

app/assets/javascripts/inputServiceWorker.js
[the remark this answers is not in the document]
winniederidder: To add to this, the service worker cannot be hosted within /assets as Rails disables these requests. We can give it its own secluded folder (e.g. /serviceworker/) to still limit the requests intercepted by the service worker (instead of putting it at the root, where 'normal' service workers tend to operate.

bmesuere (changes requested):
Code-wise, this is a great step forward 👍 

Regarding the service worker: This can only be merged with the service worker if each of these points are fixed:
- [ ] The service worker code is readable and not minimized
- [ ] The service worker file is called correctly instead of referencing the file from assets (e.g. add it to the packs so it's processed by webpack and sprockets.
- [ ] There is a clear upgrade path
- [ ] The service worker is registered using a `scope` so its impact is as minimal as possible

Regarding the UI/flow:
- New copy button location is an improvement but I would make it gray/neutral instead of green
- The button doesn't seem to copy the code to the ACE editor
- If papyros is triggered for a second time, the code in the papyros editor is overwritten by that in the ACE editor. In case of such conflict: ask the user if he wants the code replaced.
- Increase the height of the papyros editor to take up everything that isn't used by the other UI elements.
- The UI text can still use some work. I'll try to do a few suggestions later.

winniederidder:
> * The button doesn't seem to copy the code to the ACE editor

A standard copy button seems to be more logical. If we copy only to the editor, it sounds more logical to just submit it immediately. Copying to the clipboard allows the user to do with it as they please, such as pasting in a file or in the editor to submit. Are you certain you want it to just paste the code into the editor?

winniederidder:
> Regarding the service worker: This can only be merged with the service worker if each of these points are fixed:
> 
> * [ ]  The service worker code is readable and not minimized
> * [ ]  The service worker file is called correctly instead of referencing the file from assets (e.g. add it to the packs so it's processed by webpack and sprockets.
> * [ ]  There is a clear upgrade path
> * [ ]  The service worker is registered using a `scope` so its impact is as minimal as possible
> 

There is now a file called inputServiceWorker.ts in app/assets/javascripts/packs which is built by webpack and processed as requested. The upgrade path is automatic as the core functionality is mported from @dodona/papyros so that will be updated automatically when upgrading versions.
Regarding scope and serving:
When registering a service worker, it becomes active on the path it was registered on. In our ideal case, that is /assets/ as it is served as e.g. /assets/inputServiceWorker.js. It is therefore only able to intercept requests originating from that location. Activities within Dodona are hosted in different locations, such as /activities/:id, /courses/:cid/activities/:aid and /courses/:cid/series/:sid/activities/:aid. Requests happening within these pages cannot be intercepted, as they are outside the scope of the service worker. As implemented earlier, the solution is to host the service worker within the scope of the specific page, using new routes in config/routes.rb leading to ActivitiesController#isw. This controller now uses the application's assets to serve this file, which is hopefully ok. Redirecting to the actual URL using javascript_path is forbidden by the browser (scripts may not be served under redirects for security reasons). These new routes then host the serviceworker under e.g. /activities/:id/inputServiceWorker.js, alllowing requests to be intercepted only within this specific page.

winniederidder:
Updated layout following given suggestions:
scratchpad-integration

bmesuere:
> > Regarding the service worker: This can only be merged with the service worker if each of these points are fixed:
> > 
> > * [ ]   The service worker code is readable and not minimized
> > * [ ]   The service worker file is called correctly instead of referencing the file from assets (e.g. add it to the packs so it's processed by webpack and sprockets.
> > * [ ]   There is a clear upgrade path
> > * [ ]   The service worker is registered using a `scope` so its impact is as minimal as possible
> 
> There is now a file called inputServiceWorker.ts in app/assets/javascripts/packs which is built by webpack and processed as requested. The upgrade path is automatic as the core functionality is mported from @dodona/papyros so that will be updated automatically when upgrading versions. Regarding scope and serving: When registering a service worker, it becomes active on the path it was registered on. In our ideal case, that is /assets/ as it is served as e.g. /assets/inputServiceWorker.js. It is therefore only able to intercept requests originating from that location. Activities within Dodona are hosted in different locations, such as /activities/:id, /courses/:cid/activities/:aid and /courses/:cid/series/:sid/activities/:aid. Requests happening within these pages cannot be intercepted, as they are outside the scope of the service worker. As implemented earlier, the solution is to host the service worker within the scope of the specific page, using new routes in config/routes.rb leading to ActivitiesController#isw. This controller now uses the application's assets to serve this file, which is hopefully ok. Redirecting to the actual URL using javascript_path is forbidden by the browser (scripts may not be served under redirects for security reasons). These new routes then host the serviceworker under e.g. /activities/:id/inputServiceWorker.js, alllowing requests to be intercepted only within this specific page.

I don't think there's a need for this additional path unless I'm entirely mistaken how this works. We use the service worker to intercept "fake" requests to handle input, right? So we can choose whichever path we use for those fake requests as long as we can intercept them? Using the `scope` during registration, we can further restrict the scope of the service worker. So isn't it possible to send the fake requests to /assets/fake-papyros-request and limit the scope to that? This way the SW will be guaranteed to have no unwanted impact.

bmesuere:
@chvp @rien @niknetniko @jorg-vr can you review this as well?

app/javascript/packs/inputServiceWorker.ts
```diff
@@ -0,0 +1,22 @@
+// Specifically import service worker provided by the Papyros-package
+import { InputWorker } from "@dodona/papyros/dist/workers/input/InputWorker";
```
rien, line 2: Since `InputWorker` is exported in [papyros' `Libary.ts`](https://github.com/dodona-edu/papyros/blob/main/src/Library.ts), you should be able to change this to:
```suggestion
import { InputWorker } from "@dodona/papyros";
```

app/views/activities/_coding_scratchpad.html.erb
```diff
@@ -0,0 +1,38 @@
...
+        <div id="scratchpad-editor-wrapper"></div>
+        <div id="scratchpad-panel-wrapper"></div>
+        <div id="scratchpad-output-wrapper"></div>
+        <div id="scratchpad-input-wrapper"></div>
+    </div>
+</div>
+<script type="text/javascript">
+    $(function () {
+        dodona.initCodingScratchpad(
+            "<%= @activity.programming_language&.name %>"
```
rien, line 35: Since this is Javascript, you should probably use [`escape_javascript`](https://api.rubyonrails.org/classes/ActionView/Helpers/JavaScriptHelper.html#method-i-escape_javascript), just in case.

config/initializers/content_security_policy.rb
```diff
@@ -14,11 +14,15 @@
...
                          'https://*.googleapis.com',
-                         'http://localhost:3035', 'ws://localhost:3035'
+                         'http://localhost:3035', 'ws://localhost:3035',
+                         # Pyodide and related packages
+                         'https://cdn.jsdelivr.net/pyodide/', 'https://pypi.org/pypi/', 'https://files.pythonhosted.org/packages/'
     else
-      policy.connect_src :self, Rails.configuration.tutor_url.to_s
+      policy.connect_src :self, Rails.configuration.tutor_url.to_s,
+      # Pyodide and related packages
+      'https://cdn.jsdelivr.net/pyodide/', 'https://pypi.org/pypi/', 'https://files.pythonhosted.org/packages/'
```
rien, line 23: Allowing full access to pypi and python packages is maybe a lot to allow application-wide. Maybe this should be set more specific on the page where this could be allowed?

Is full access to pypi needed and desired? Not all python packages are available inside the python judge.

app/controllers/activities_controller.rb
```diff
@@ -177,6 +194,15 @@ def media
     end
   end
 
+  # Serve the inputServiceWorker asset required to handle input in Papyros
+  # Asset has been preprocessed and built internally
+  # Redirecting to the asset is not possible due to browser security policy
+  def isw
```
rien, line 200: This name isn't very descriptive. Maybe this could be changed to `input_service_worker`?

app/views/activities/_coding_scratchpad.html.erb
jorg-vr, line 35: Could you provide the isw asset path here as an argument `"<%= asset_path("inputServiceWorker.js") %>"`
and use something like `await papyros.configureInput(location.host, isw_path, false);` in the papyros launch?

I think this would avoid the need for a custom path to the asset.
But maybe you have already tried this.

winniederidder:
> I don't think there's a need for this additional path unless I'm entirely mistaken how this works. We use the service worker to intercept "fake" requests to handle input, right? So we can choose whichever path we use for those fake requests as long as we can intercept them? Using the `scope` during registration, we can further restrict the scope of the service worker. So isn't it possible to send the fake requests to /assets/fake-papyros-request and limit the scope to that? This way the SW will be guaranteed to have no unwanted impact.

Sadly, this is not doable. The scope on the service worker tells the worker from which origins it can intercept requests. This scope is relative to where the service worker is hosted. If it is hosted under /assets, it can only intercept requests originating from /assets. Since our requests come from the exercise page to /assets, it is unable to change them. This is why I need to resolve the worker from a local scope /exercise/:id/serviceworker so it can intercept requests originating from /exercise/:id

config/environments/development.rb
```diff
@@ -1,5 +1,23 @@
 require "active_support/core_ext/integer/time"
 
+class AssetHeaders
```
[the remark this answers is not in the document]
winniederidder: The headers served with the exercise pages to support SharedArrayBuffers (COEP/CORP) cause a mismatch between origins when fetching assets inside an iframe, causing the sandboxed origin to not be allowed to fetch e.g. stylesheets anymore (as origins are isolated). The cross-origin headers allow the sandboxed host to still fetch these. These will indeed be needed in other environments as well.

app/views/activities/_coding_scratchpad.html.erb
winniederidder: The custom path is required because of reasons I described in my reply to Bart, which you can find below.

config/locales/js/en.yml
```diff
@@ -168,3 +168,6 @@ en:
       confirm: Are you sure?
     score_item:
       error: Error while updating
+    coding_scratchpad:
+      copy_code: Copy
```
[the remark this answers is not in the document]
winniederidder: That will cause the button to be quite big. I agree "Copy" isn't ideal.

app/javascript/packs/inputServiceWorker.ts
rien: Ah, sad. Maybe it is a good idea to mention the reason for this type of import in a comment then?

winniederidder: The comment indeed does not mention this specific issue, only that I specifically import the worker. I'll update it to be more clear.

config/environments/development.rb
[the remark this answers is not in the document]
chvp: In other environments these headers will be set by Apache, I assume? Then this will need a matching PR in the dodona-ansible repo.

config/environments/development.rb
```diff
@@ -1,5 +1,23 @@
...
+  def initialize(app)
+    @app = app
+  end
+
+  def call(env)
+    request = Rack::Request.new(env)
+    response = @app.call(env)
+    if request.path =~ /^\/assets\//
+      # COEP does not seem required
+      #response[1]['Cross-Origin-Embedder-Policy'] = 'require-corp'
```
chvp, line 14: Remove the commented out code.

config/environments/development.rb
```diff
@@ -38,14 +56,19 @@
 
   # Enable/disable caching. By default caching is disabled.
   # Run rails dev:cache to toggle caching.
+
+  config.public_file_server.headers = {
+     # 'Cross-Origin-Embedder-Policy' => 'require-corp',
```
chvp, line 61: Same here.

config/initializers/content_security_policy.rb
```diff
@@ -14,11 +14,15 @@
...
     end
-
+
```
chvp, line 25: Remove trailing whitespace.

chvp (approved):
No additional comments except for a few small nits.

config/locales/js/en.yml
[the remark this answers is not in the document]
winniederidder: Changed it to "Copy to Dodona" but layout-wise this isn't ideal. @bmesuere thoughts?

chvp:
The sideways "Begin met coderen" looks a little weird to me. Also, toggling to batch input doesn't seem to work for me on Naos. Not sure if this is due to deployment issues or otherwise.

winniederidder:
> The sideways "Begin met coderen" looks a little weird to me. Also, toggling to batch input doesn't seem to work for me on Naos. Not sure if this is due to deployment issues or otherwise.

"Start coding" doesn't seem to have a better translation, but suggestions are welcome. Regarding the switch, a fix will be pushed very soon, it's not deployment related.

config/locales/views/activities/nl.yml
```diff
@@ -126,3 +126,9 @@ nl:
     series_activities_table:
       course_has_no_rights_with_fix: Deze cursus heeft niet de rechten om deze privé-oefening te gebruiken. Klik hier om deze rechten te geven.
       course_has_no_rights_with_contact: Deze cursus heeft niet de rechten om deze privé-oefening te gebruiken. Contacteer de auteur of verwijder deze oefening uit de cursus.
+    coding_scratchpad:
+      title: "Voer %{programming_language} code uit in je browser"
+      experimental_feature_html: Deze experimentele toepassing laat je toe om code te schrijven en uit te voeren in je browser, zodat je meteen kan beginnen. Suggesties en problemen kan je melden bij <a href="%{url}">team Dodona</a>.
```
bmesuere, line 131:
```suggestion
      experimental_feature_html: In dit venster kan je Python-code uitvoeren zonder iets te moeten installeren. De code die je hier schrijft wordt <b>niet</b> automatisch ingediend in Dodona. Deze functie is nog in volle ontwikkeling. We horen graag je suggesties en problemen via <a href="%{url}">het contactformulier</a>.
```

config/locales/views/activities/en.yml
```diff
@@ -125,3 +125,9 @@ en:
     series_activities_table:
       course_has_no_rights_with_fix: This course is not allowed to use this private exercise. Click here to grant this course those rights.
       course_has_no_rights_with_contact: This course is not allowed to use this private exercise. Contact the author to grant rights or remove this exercise from the course.
+    coding_scratchpad:
+      title: "Run %{programming_language} code in your browser"
+      experimental_feature_html: This experimental feature lets you write and run code in your browser to get you started immediately. Please report any suggestions and issues to <a href="%{url}">team Dodona</a>.
```
bmesuere, line 130:
```suggestion
      experimental_feature_html: This window allows you to run Python code without installing a thing. The code you write here is <b>not</b> automatically submitted to Dodona. This feature is still in full development, so we're happy to hear about your suggestions and issues using <a href="%{url}">the contact form</a>. 
```

app/assets/stylesheets/components/scratchpad.css.scss
```diff
@@ -0,0 +1,60 @@
...
+    height: $navbar-height;
+    padding-left: 16px;
+    padding-right: 16px;
+    align-items: center;
+}
+
+.copy-code-button {
+    background-color: $grey-500;
+    color: white;
+    width: fit-content;
```
bmesuere, line 34:
```suggestion
    white-space: nowrap;
```

config/locales/js/nl.yml
```diff
@@ -168,6 +168,9 @@ nl:
       confirm: Bent u zeker?
     score_item:
       error: Fout bij bijwerken
+    coding_scratchpad:
+      copy_code: Kopieer naar Dodona
+      overwrite_code: Wil je de code in het kladblok overschrijven met je huidige oplossing?
```
bmesuere, line 173:
```suggestion
      overwrite_code: Wil je de code in het kladblok overschrijven met je laatste aanpassingen?
```

config/locales/js/en.yml
```diff
@@ -168,6 +168,9 @@ en:
       confirm: Are you sure?
     score_item:
       error: Error while updating
+    coding_scratchpad:
+      copy_code: Copy to Dodona
+      overwrite_code: Do you want to overwrite the code in the scratchpad with your current solution?
```
bmesuere, line 173:
```suggestion
      overwrite_code: Do you want to overwrite the code in the scratchpad with your latest changes?
```

config/locales/views/activities/nl.yml
```diff
@@ -126,3 +126,9 @@ nl:
...
+      close: Sluiten
+      start_coding: Begin met coderen
```
bmesuere, line 133:
```suggestion
      start_coding: Code uitvoeren
```
In the future, we would use "Kladblok" and "Scratchpad" as label text + add some kind of run icon.

bmesuere (changes requested):
I did this review together with @chvp 

- Please tweak the padding of the buttons to use `px-3 py-1` and remove `border-2`. The inset can probably also be removed. 
- Add additional margin to the left-most button (so the total is equivalent to `m-2`) + add the same margin to the right of the status text on the right. 
- Use cursor not-allowed instead of wait where appropriate.
- Set the bg color of the copy button to gray-300
- the line input text has a smaller bottom margin than the text area which causes the toggle text to stick to the input. Add a bit margin here.
- close papyros after clicking the copy button
- can you move the "do you want to update your scratchpad code" prompt to after expanding the sidebar?
- non-interactive input is broken on naos (both chrome and firefox)
   image

Regarding the scope of the service worker: you are correct. The scope seems to filter on the origin/initiator and not the destination. Can you point me to the lines of code where your own request filtering is done?

app/assets/stylesheets/components/scratchpad.css.scss
```diff
@@ -0,0 +1,60 @@
...
+}
+
+.scratchpad-info {
+    padding-left: 20px;
+    font-size: 14px;
+    color: $text-muted;
+}
+
+.scratchpad-header {
+    color: $header-fg;
```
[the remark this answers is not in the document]
winniederidder: Display:flex causes the height of the header to be slightly smaller than navbar-height. Maybe it is best to remove this close button, as clicking anywhere off the offcanvas closes it too so the button is quite redundant? Then we can remove the flex parts (such as the flex-spacer, display-flex and the button itself)

winniederidder:
> I did this review together with @chvp
> 
> * Please tweak the padding of the buttons to use `px-3 py-1` and remove `border-2`. The inset can probably also be removed.
> * Add additional margin to the left-most button (so the total is equivalent to `m-2`) + add the same margin to the right of the status text on the right.
> * Use cursor not-allowed instead of wait where appropriate.
> * Set the bg color of the copy button to gray-300
> * the line input text has a smaller bottom margin than the text area which causes the toggle text to stick to the input. Add a bit margin here.
A PR for styling was opened at [Papyros](https://github.com/dodona-edu/papyros/pull/117)

> * non-interactive input is broken on naos (both chrome and firefox)
>   image

This is unrelated to the browser but related to [comsync](https://github.com/alexmojaki/comsync/issues/3), I patched it until a proper fix is implemented there.

> Regarding the scope of the service worker: you are correct. The scope seems to filter on the origin/initiator and not the destination. Can you point me to the lines of code where your own request filtering is done?

Dodona's service worker is located [here](https://github.com/dodona-edu/dodona/blob/papyros-integration-v2/app/javascript/packs/inputServiceWorker.ts). It uses the [InputWorker](https://github.com/dodona-edu/papyros/blob/main/src/workers/input/InputWorker.ts) provided by Papyros. Since a hostname is not provided by Dodona to limit the impact of the service worker, this boils down to simply using the serviceWorkerFetchListener provided by sync-message defined [here](https://github.com/alexmojaki/sync-message/blob/master/lib/index.ts) where the current implementation only affects urls containing "__SyncMessageServiceWorkerInput__" for reading and writing.

bmesuere (commented):
Would it be possible to add/duplicate the intervention check to the pack in dodona?
e.g. instead of 
```
    if (!await inputHandler.handleInputRequest(event)) {
        return;
    }
```
something like 
```
    if (/* check if url contains SyncMessageServiceWorkerInput*/) {
        inputHandler.handleInputRequest(event);
    } else {
        return;
    }
```

This way, a broken papyros package can't break dodona requests.

winniederidder:
> Would it be possible to add/duplicate the intervention check to the pack in dodona? e.g. instead of
> 
> ```
>     if (!await inputHandler.handleInputRequest(event)) {
>         return;
>     }
> ```
> 
> something like
> 
> ```
>     if (/* check if url contains SyncMessageServiceWorkerInput*/) {
>         inputHandler.handleInputRequest(event);
>     } else {
>         return;
>     }
> ```
> 
> This way, a broken papyros package can't break dodona requests.

We exposed the filter in sync-message, which is now explicitly added in the Dodona service worker to help with this.

bmesuere:
@winniederidder Thank you for the update, but it doesn't fix the concern I have. Instead, it even makes it more visible. I'll try to explain: if due to a bug or misconfiguration the sync-message library always returns true for the call we do know, all legitimate Dodona requests could fail which is a major problem. What I would like here is a check in code that we own (i.e. in the pack file) that limits the number of requests that we hand over to the library so that worst case, when something breaks, only papyros stops working. Note that this additional check must not be the exact same one that happens in the library as long as it limits the scope of impact enough.

winniederidder:
> @winniederidder Thank you for the update, but it doesn't fix the concern I have. Instead, it even makes it more visible. I'll try to explain: if due to a bug or misconfiguration the sync-message library always returns true for the call we do know, all legitimate Dodona requests could fail which is a major problem. What I would like here is a check in code that we own (i.e. in the pack file) that limits the number of requests that we hand over to the library so that worst case, when something breaks, only papyros stops working. Note that this additional check must not be the exact same one that happens in the library as long as it limits the scope of impact enough.

Using the BASE_URL_SUFFIX in the test from the library would then also be risky in this sense. I could copy the value of the this suffix from the library into this file so it is fully under our control.

bmesuere (changes requested):
Two more minor issues:
- The copy to dodona button should have black text instead of white
- Clicking that button correctly copies the code to Dodona, but the "Submit" tab isn't set to active (e.g. if you were on the feedback tab when opening the scratchpad, I would expect it to switch back to submit when the code is copied to show you the change)

Can you also make sure the latest (working) version is deployed at papyros.dodona.be? We will use it in a demo tomorrow.

winniederidder:
> * Clicking that button correctly copies the code to Dodona, but the "Submit" tab isn't set to active (e.g. if you were on the feedback tab when opening the scratchpad, I would expect it to switch back to submit when the code is copied to show you the change)

Do we want to scroll the submit panel into view? I experimented a bit with scrolling to #handin or #activity-handin-link but neither yielded a simple and reliable scrolling behaviour. I'll look into it more if this is something you want added.

niknetniko (approved):

bmesuere:
Report by @pdawyndt 

The "Run your code" offcanvas-button always seems to trigger a horizontal scrollbar, that disappears as soon as Papyros is brought on canvas. Try to avoid this horizontal scrollbar if not needed.

Example: https://naos.ugent.be/en/courses/56/series/591/activities/203584369/

image

image

winniederidder:
Update to properly support dark mode in Papyros:
dodona-dark-mode

winniederidder:
> Report by @pdawyndt
> 
> The "Run your code" offcanvas-button always seems to trigger a horizontal scrollbar, that disappears as soon as Papyros is brought on canvas. Try to avoid this horizontal scrollbar if not needed.
> 
> Example: https://naos.ugent.be/en/courses/56/series/591/activities/203584369/
> 
> image
> 
> image

I do not get a horizontal scrollbar when opening the Manhattan Python exercise in development locally:
no-scrollbar

It might have to do with the button being filled in first, affecting the page width, before being rotated.

bmesuere (changes requested):
Alright, dark mode looks good! Two (small) remaining issues:
- the vertical button to open the scratchpad in dark mode doesn't look like a button because it's missing a border. I suggesting adding border-color hairline gray (it's a existing var) and border-bottom 0
- the tests are still failing

test/application_system_test_case.rb
```diff
@@ -20,7 +20,7 @@
 end
 
 Capybara.configure do |config|
-  config.default_max_wait_time = 10
+  config.default_max_wait_time = 20
```
bmesuere, line 23: Why is this necessary? The value of 10 was there with a reason to keep the total runtime of these tests acceptable. These tests have increased it from 2.5 to 3.5 minutes.

winniederidder: Pyodide intialization takes a while, so running a piece of code and expecting output to show up takes a while. The find methods take their waiting time from the capybara config and it doesn't seem to be possible to update the config for this test file only. This system test probably also takes longer because separate pages are loaded per test.  I can try to do it all on the same page instead and make the waits longer only when required.

bmesuere: you can set it for checks where it fails: https://blog.tomoyukikashiro.me/post/things-to-be-aware-of-when-change-capybara_default_wait_time/

winniederidder: Does the minitest-suite we use support test nesting? Something similar to Jest using 'describe' and 'it'-blocks, as now I combined all tests into a big one separated by comment lines to prevent having to load Pyodide from scratch multiple times.

bmesuere (approved):

PR #3484, opened 2022-03-16
Color code thumbs up/down in evaluation

### LINKED ISSUE #2715: Switch order of thumbs up/down buttons in grading panel
In the grading panel, the 0/max buttons have been replaced by thumbs up/down icons. Where it was clear that with the 0/max buttons, the expected locations of the buttons was zero-left and max-right, this is no longer the case with the thumbs up/down icons.

Proposal: Put the thumbs up button to the left and the thumbs down button to the right. As a signal that we prefer good scores over bad scores. If most students have done a good job at an exercise, the thumbs up button also comes first in the left-to-right reading order while filling up the scores. 

This way

image

would become

image

BTWS2:
This pull request closes #2715.

I don't have strong opinions on the order of thumbs up or down, but I'm in favor of color coding (thumbs up = green, thumbs down = red) to nudge towards the correct button as @rien suggested. I can switch the order if needed:
https://github.com/dodona-edu/dodona/blob/develop/app/views/feedbacks/_feedback_actions.html.erb
https://github.com/dodona-edu/dodona/blob/develop/app/views/feedbacks/_score_actions.html.erb

image
image
image
image

### DIFF
--- a/app/assets/stylesheets/components/btn.css.scss
+++ b/app/assets/stylesheets/components/btn.css.scss
@@ -291,3 +291,11 @@ form.button_to {
 .option-btn:hover {
   @include shadow-z4;
 }
+
+.single-zero-button:hover, #zero-button:hover{
+  background-color: $brand-danger;
+}
+
+.single-max-button:hover, #max-button:hover{
+  background-color: $brand-success;
+}


jorg-vr:
Peek 2022-03-17 10-54

I prefer to update the icon color instead of background color. (But this is just a matter of taste)

Overall I approve of the change

BTWS2:
My goal with this approach is to add a visual confirmation the teacher pressed the right button. This shows the color after pressing the button. When the teacher hovers over the wrong button, a glimpse of the wrong color is shown with the previous approach. I agree with @jorg-vr that changing the color of the icon looks cleaner.

It's based on https://css-tricks.com/different-transitions-for-hover-on-hover-off/.

https://user-images.githubusercontent.com/56451049/158904550-ef138b63-afcf-4de2-8544-4e7b0a821dd7.mp4

bmesuere:
Why not combine both behaviours? Color on mouse over + a slight delay after clicking?

BTWS2:
@bmesuere A color change on mouse over (without animation) and visual feedback after clicking would indeed be better. If I had to choose I prefer the visual feedback after clicking, but both is better.

BTWS2:
Changing the `color` property in both states wasn't possible without javascript . This is the new proposal using pure CSS: small color change on hover and clicking shows more green/red for clear visual feedback.

https://user-images.githubusercontent.com/56451049/159768808-a9711bd5-5807-4b4b-86ac-258c63ab0235.mp4

app/assets/stylesheets/components/btn.css.scss
```diff
@@ -291,3 +291,35 @@ form.button_to {
 .option-btn:hover {
   @include shadow-z4;
 }
+
+.single-zero-button, #zero-button{
```
chvp, line 295:
```suggestion
.single-zero-button, #zero-button {
```
(Same for the other braces.)

app/assets/stylesheets/components/btn.css.scss
```diff
@@ -291,3 +291,35 @@ form.button_to {
...
+  background-color: $brand-danger;
+  color: unset;
+  transition: background-color 0s;
+}
+
+.single-max-button:active, #max-button:active{
+  background-color: $brand-success;
+  color: unset;
+  transition: background-color 0s;
+}
```
chvp, line 325: Please set up your editor to use the `.editorconfig` file in the root of the repository. That should make it so a final newline is inserted.

chvp (changes requested):
The background color is still quite harsh, I think. The end of the animation also looks a little strange to me. Maybe a fade-out would make it less jarring?

BTWS2:
I changed the background to lighter shade of the color. The end of the animation is cut off at the end, because the button is updated when the loading is done (visible at the end of the screen recording of dark mode below). I made the animation shorter and used `ease-out` to mitigate this partly. I added a space before the brackets and added a newline at the end.

At first I chose a lighter color (`$danger-100`), but it resulted in even more harsh colors in dark mode (e.g. `$danger-100: $red-800 !global;`), so I used opacity instead.

Light mode:

https://user-images.githubusercontent.com/56451049/161397648-62fee4d5-b0c6-433c-83c6-1bed54ca3b30.mp4

Dark mode:

https://user-images.githubusercontent.com/56451049/161397654-fadb9aa3-5833-4d10-8022-500b1ab24b8f.mp4

Issue #3456, opened 2022-03-03
Rework submission table summary


bmesuere:
- [x] number the submissions for a student for an exercise in a course
  - to be stored in the database
  - can be shown in the tables e.g. as `#23`
  - can be used to link to a submission #2213 
- [x] something visual with submission history
- [x] motivational message
  - congrats + link to next exercise

PR #3470, opened 2022-03-10
Add vertical submission history to submission detail view

jorg-vr:
This pull request adds a small submission history table to the submission detail page. This allows for easy navigation to previous submissions. It also gives users a better sense of the progress they are making, visualizing both the total submissions count as well as submission summaries (For example containing diminishing numbers of failing tests).

Peek 2022-03-11 10-55

Closes #3469 
Part of #3456

PR #3485, opened 2022-03-17
Number submissions by user, exercise and course

### LINKED ISSUE #2213: Direct link to individual submissions

jorg-vr:
This pull request adds a column number to the submissions table. This column represents the submission number scoped by student, exercise and course.

image
image
image
image

- [x] Tests were added

Part of #3456 .
Closes #2213 .


BTWS2:
@jorg-vr Is there a "copy submission link to clipboard button" implemented in this design? Related: #2213 and #3215.

jorg-vr:
@BTWS2  There is no copy to clipboard in this design. But the numbers are actual links. So a right click `copy link address` will work, as well as clicking on the link and then selecting the new address in the address bar.

app/models/submission.rb
```diff
@@ -96,6 +97,10 @@ class Submission < ApplicationRecord
     correct.group(:exercise_id, :user_id).least_recent
   }
 
+  trigger.before(:insert) do
```
bmesuere, line 100: Can you motivate why you need this (hairtrigger)? It seems to require a number of big dependencies.

app/views/submissions/_submissions_table.html.erb
```diff
@@ -3,6 +3,7 @@
     <thead>
     <tr>
       <th class="status-column"></th>
+      <th><%= t ".number" %></th>
```
bmesuere, line 6: I would not add a title to this column because of the extra width it introduces. Instead a title tag to the individual links such as `submission #23` might provide enough context.

db/migrate/20220317145048_create_trigger_submissions_insert.rb
```diff
@@ -0,0 +1,17 @@
+# This migration was auto-generated via `rake db:generate_trigger_migration'.
+# While you can edit this file, any changes you make to the definitions here
+# will be undone by the next auto-generated trigger migration.
+
+class CreateTriggerSubmissionsInsert < ActiveRecord::Migration[7.0]
+  def up
+    create_trigger("submissions_before_insert_row_tr", :compatibility => 1).
```
bmesuere, line 7: I would not add these using a migration. Because of the size of the submissions table, this will take a very long time (with the deploy also taking this much time). I would make sure the code can handle "missing" numbers and fill these in later using the rails console or runner function.

db/schema.rb
```diff
@@ -461,15 +461,15 @@
...
-    t.datetime "updated_at", precision: 6, null: false
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
     t.string "lang", default: "nl"
     t.string "token"
     t.string "time_zone", default: "Brussels"
     t.bigint "institution_id"
     t.string "search", limit: 4096
-    t.datetime "seen_at", precision: 6
-    t.datetime "sign_in_at", precision: 6
```
bmesuere, line 472: weren't these additional precisions added in a previous PR?

bmesuere (changes requested):

app/models/submission.rb
jorg-vr: I mostly used it to have database independent triggers. It also helped to get the triggers in the test database (But there are probably also other solutions for this)

But it is no big issue for me to leave it out and just use a plain mysql trigger (As I did in my earlier commits)

db/migrate/20220317145048_create_trigger_submissions_insert.rb
jorg-vr: These are not added in using a migration. I only add the trigger here (So that for each future submission the number is set correctly)

jorg-vr: The code is able to handle missing numbers

db/schema.rb
jorg-vr: Damn it looked over it when reviewing my own pr :/ 
I'll have a look at my database setup, because he sometimes removes and re adds these precisions when I am rolling out new migrations, but I am not sure why

BTWS2:
@jorg-vr Would you consider adding [a chainlink icon like Peter suggested](https://github.com/dodona-edu/dodona/issues/2213#issue-692134559) before the `#` (similar to the filter icon)? And also in #3470?

> image
> Suggestion from Peter





Ideally sharing the link should be as simple as possible. In the past we had the really emphasize we want a submission link and not https://dodona.ugent.be/nl/courses/???/activities/???/ for example (where the submission is visible at the bottom of the page for students but not for teachers).

I'm not keen on the right click method. The text is different dependent on the browser and language (in Chrome `Linkadres kopiëren` and in Edge `Koppeling kopiëren`). If students use this regularly I would like it to be more discoverable.

Clicking on the link and selecting the new address in the address bar involves too many steps (new page load, select full link, right click, copy). If a students clicks one more time, the selection is gone and they have to select the URL themselves. Not everyone uses the `Ctrl+C` shortcut, which adds another step.

For young students a single click in the UI, which copies the link to the clipboard is less cumbersome.

db/schema.rb
jorg-vr: Researched it a bit.

It is reletatd to https://github.com/rails/rails/issues/43909 and https://github.com/rails/rails/issues/44571

From one of the comments
> `precision: 6` is the default in schema version `7.0`. (But before `7.0.2`, it was still [unnecessarily] being dumped explicitly.)

So the current rails version removes the unneeded precision statements that match the default when I run `rails db:migrate`

app/models/submission.rb
```diff
@@ -517,4 +519,10 @@ def report_if_internal_error
       }
     )
   end
+
+  def set_number
+    submissions = Submission.of_user(user).of_exercise(exercise)
+    submissions = submissions.where(course_id: course&.id)
```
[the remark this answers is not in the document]
chvp: It could be me misremembering as well. @bmesuere Do you recall?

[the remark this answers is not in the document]
jorg-vr: The code is already there, I will leave it as is. The numbers won't hurt anyone and could potentially be useful to some

chvp (approved):

app/models/submission.rb
```diff
@@ -517,4 +519,10 @@ def report_if_internal_error
...
+    self.number = submissions.length + 1
```
bmesuere, line 526: this should probably be count or size to prevent loading everything into memory.

bmesuere (approved):

PR #3494, opened 2022-03-22
Add dropdown filters to searchable tables

jorg-vr:
This pull request

- [ ] Tests were added
Part of 3458

### DIFF
--- a/app/views/layouts/_searchbar.html.erb
+++ b/app/views/layouts/_searchbar.html.erb
@@ -22,6 +22,61 @@
       </div>
     <% end %>
   </div>
+    <details class="text-muted" style="margin: 0.5em 0 0 1em; font-size: 14px">
+      <summary>More filters</summary>
+      <form>
+        <div class="dropdown m-1" style="display: inline-block">
+          <a class="btn btn-secondary dropdown-toggle" href="#" role="button" id="dropdownMenuLink" data-bs-toggle="dropdown" aria-expanded="false" style="font-size: 14px">
+            Programming language <i class="mdi mdi-chevron-down mdi-12"></i>
+          </a>
+
+          <ul class="dropdown-menu" aria-labelledby="dropdownMenuLink">
+            <li><a class="dropdown-item" href="#">Read activity <span class="badge rounded-pill bg-info float-end">54</span></a></li>
+            <li><a class="dropdown-item" href="#">Python <span class="badge rounded-pill bg-info float-end">45</span></a></li>
+            <li><a class="dropdown-item" href="#">HTML <span class="badge rounded-pill bg-info float-end">21</span></a></li>
+            <li><a class="dropdown-item" href="#">Java <span class="badge rounded-pill bg-info float-end">12</span></a></li>
+            <li><a class="dropdown-item disabled" href="#">Haskel <span class="badge rounded-pill bg-info float-end">0</span></a></li>
+          </ul>
+        </div>
+        <div class="dropdown m-1" style="display: inline-block">
+          <a class="btn btn-secondary dropdown-toggle" href="#" role="button" id="dropdownMenuLink" data-bs-toggle="dropdown" aria-expanded="false" style="font-size: 14px">
+            Language <i class="mdi mdi-chevron-down mdi-12"></i>
+          </a>
+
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/index.js
```diff
@@ -31,10 +32,15 @@ function setBaseUrl(_baseUrl) {
     window.dodona.index.doSearch();
 }
 
-function initFilterIndex(_baseUrl, eager, actions, doInitFilter, filterCollections, refreshElement = null) {
+function initFilterIndex(_baseUrl, eager, actions, doInitFilter, filterCollections, refreshElement = null, showDropdowns = true) {
```
chvp, line 35: Is this ever passed as false?

app/assets/javascripts/components/dropdown_filter.ts
```diff
@@ -0,0 +1,78 @@
+import { html, LitElement, TemplateResult } from "lit";
+import { customElement, property } from "lit/decorators.js";
+
+type Label = { id: string | number, name: string };
+
+@customElement("dodona-dropdown-filter")
+export class DropdownFilter extends LitElement {
+    @property({ type: Boolean })
+        multi: boolean;
```
chvp, line 9: This indentation looks kinda weird to me. It looks like it is seen as a continuation of the last line?

chvp (commented):
Some descriptions are not appropriate for the dropdown header, I think. E.g. "Talen waarin de beschrijving beschikbaar is".

app/assets/javascripts/components/dropdown_filter.ts
jorg-vr: problem is with eslint (Which sugested this indent)
But apparently there is a workaround here: https://github.com/typescript-eslint/typescript-eslint/issues/1824
I will add it

app/assets/javascripts/index.js
jorg-vr: No, I was unsure if we wanted the dropdown to appear on every searchbar which has search suggestions, but I can leave this option out for now as it is very easily added in the future

jorg-vr:
@chvp I agree that translation is not ideal for a dropdown header. I would update update the translation to keep it consistent with the searchbar. Or would you  prefer separate translation keys for the dropdown headers?

chvp:
> @chvp I agree that translation is not ideal for a dropdown header. I would update update the translation to keep it consistent with the searchbar. Or would you prefer separate translation keys for the dropdown headers?

Keeping it consistent seems best to me as well.

niknetniko (approved):

chvp (approved):

bmesuere (changes requested):
For future reference, the feedback I gave in person last week:
- [x] make it possible to disable individual dropdowns. e.g for exercises, labels (very long list) and judge (not useful) can be removed. For courses, I don't think the institution filter is useful since we already have the tabs for that.
- [ ] If only a single thing can be selected, I would use a singular noun as title. e.g. Programmeertaal instead of programmeertalen. "Taal van beschrijving" seems quite long to me?
- [ ] Since we use checkboxes for multi-select options, wouldn't it make sense to use radio buttons in the menu for the others instead of highlighting 

UI
- [ ] The dots are a nice indicator (the material design spec uses checkmarks, but the dots look nicer). I would add them to the left of the label instead of the right.
- [ ] The dropdowns have a drop shadow where the rest of the search interface looks quite "flat" so they feel a little out of place. I think this is also because the chips/labels have more generous padding where the dropdowns are quite tight and have a very small border radius.
- [ ] The left alignment feels a bit off. Maybe it looks better if it uses the same indentation as the fields above (so the search icon is in the middle of the token/search/filter area?

jorg-vr:
@bmesuere on the isue of the left margin:
In #3498 I changed the position of the search tokens to align with the search icon
image
This will be consistent with the alignment of the dropdowns in this pr:
image

bmesuere (approved):
Thanks for the changes! To me this is a nice improvement. Maybe a final suggestion: add a no wrap to the dropdown labels so they are displayed on a single line. Or was there a reason not to do this?

Can you rebase the other search-related PR after this is merged?

Issue #3458, opened 2022-03-03
Rework filtering and sorting


bmesuere:
- #1593
- #1713 
- #1714
- make filters more visible by (optionally) generating form fields

PR #3498, opened 2022-03-24
Refactor index.js to webcomponents

jorg-vr:
This pull request contains a major refactor of what used to be `index.js`.

It also replaces [uglifier](https://github.com/lautis/uglifier) by [terser](https://github.com/ahorek/terser-ruby) to minify js.
This was a requirement because uglifier failed to compile lit components that use [ref](https://lit.dev/docs/templates/directives/#ref), which this pr uses for the search field.  Terser is also more futureproofed as it supports ES6 and is more actively developed.

# Refactor of `index.js`
## Why?
1. it was outdated:
  - It was still written in js instead of typescript
  - It used no longer supported packages ([Typeahead](https://github.com/twitter/typeahead.js/)
  - it still relied heavily on jquery
2. It has grown out of the scope it was originally intended for, with code being added on top without proper refactor
4. State was being handled by html (Tags which were or were not present)
5. Multiple new features will be added that interact with this feature (see #3458)

## New structure
Everything is centered around the class `SearchQuery` defined in [search.ts](app/assets/javascripts/search.ts)
This class represents the state of the query parameters. It contains two maps of query parameters (one with single values and one with an array of values), which follow the subscriber scheme, allowing others to listen to changes in these maps;

The class already adds a first listener itself, that preforms a new search every time a query parameter changes.

Different webcomponents are created that interact with this state:

### [FilterButton](app/assets/javascripts/components/filter_button.ts)
This is a very basic component that replaces a clickable token (and sets a given query param to a given value)

The same file also contains `FilterIcon` which replaces the filter icons in tables (and sets the query param `filter` to a given value). This components uses the above component as it is a basic extension.

### [SearchActions](app/assets/javascripts/components/search_actions.ts)
This represents the optional dropdown to the right of the searchbar.
It contains searchoptions which directly interact with `SearchQuery`.
It also has searchactions which often use `SearchQuery` to get the current url params.

### [SearchField](app/assets/javascripts/components/search_field.ts)
This component represents the search field. It directly manipulates the filter parameter, but it also has search suggestions which interact with their own set of query parameters.
The search suggestion are effectively a custom implementation of the old typeahead module.

https://user-images.githubusercontent.com/21177904/165923720-3e8fd919-eae2-4768-8c47-d7b72d89068f.mp4

### [SearchTokens](app/assets/javascripts/components/search_token.ts)
These are the tokens displayed above the searchbar. These now act completely separate from the search field and only interact with the `SearchQuery` state.

### Search dropdowns 
These were develop in their own pr as webcomponents, and adjusted in this pr to use the searchQuery class. See #3494

## Other impacted places
`course.js` contained some special functionality to filter using tabs. I did not refactor this to components, but dit updete the code to use the `SearchQuery` state.

The page numbering for remote fetch (using will paginate defined in `app/helpers/application_helper.rb`) was not working correctly with our query parameters (even in existing code). Concretely it did not update the 'page' query param. I made an update to correct this.

A lot of rails views have been updated to use the new components.


app/assets/javascripts/components/search_actions.ts
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    getSearchActions(): Array<SearchAction> {
+        return this.actions.filter(isSearchAction);
+    }
+
+    // don't use shadow dom
+    createRenderRoot(): Element {
+        return this;
+    }
```
chvp, line 99: This seems to be present in (almost?) every one of the element classes. Would it make sense to make a generic DodonaElement, so that we don't have to repeat this all the time?

app/assets/javascripts/components/search_actions.ts
```diff
@@ -0,0 +1,167 @@
...
+    type: string,
+    text: string,
+    action: string,
+    js: string,
+    confirm: string,
+    icon: string
+};
+
+const isSearchOption = (opt): opt is SearchOption => (opt as SearchOption).search !== undefined;
+const isSearchAction= (action): action is SearchAction => (action as SearchAction).js !== undefined || (action as SearchAction).action !== undefined || (action as SearchAction).url !== undefined;
```
chvp, line 19: The formatting of these two lines is different?

app/views/activities/_activities_table.html.erb
```diff
@@ -67,15 +64,17 @@
           <% if policy(activity).info? %>
             <div class="text-muted ellipsis-overflow">
               <small>
-                <span class="clickable-token" title="<%= activity.repository.name %>: <%= activity.path %>" data-type="repositories" data-name="<%= activity.repository.name %>"><%= activity.repository.name %>: <%= activity.path %></span>
-              </small>
+                  <dodona-filter-button param="repository_id" value="<%= activity.repository.id %>" title="<%= activity.repository.name %>: <%= activity.path %>" >
+                      <%= activity.repository.name %>: <%= activity.path %>
+                  </dodona-filter-button>
+                </small>
```
chvp, line 70: Indentation seems off here?

chvp (changes requested):
- [x] Only some small code nits. I didn't do any functional testing yet, since the compile step of the assets seems to fail when deploying to Naos.

app/assets/javascripts/components/search_actions.ts
jorg-vr: I'll call it a ShadowlessLitElement or something (to explain the class better, and for some dodona elements the shadow dom can still be usefull)

niknetniko (changes requested):
A big improvement!

I left some minor comments here and there, and:

- [ ] Would it be possible to indicate the initial sort of the table somehow?
- [x] Sometimes when searching, the dropdown with suggestions is shown empty:
image

app/javascript/packs/application_pack.js
```diff
@@ -27,6 +27,13 @@ const bootstrap = { Alert, Button, Collapse, Dropdown, Modal, Popover, Tab, Tool
 window.bootstrap = bootstrap;
 
 import "polyfills.js";
+
+import "components/sort_button.ts";
+import "components/search_actions.ts";
+import "components/search_field.ts";
+import "components/search_token.ts";
+import "components/filter_button.ts";
```
bmesuere, line 35: Is there a reason why these are added here and not in the search pack?

bmesuere (changes requested):
This is awesome work 🔥 

- [x] In the screen recording, the "whitespace" of the tokens looks off, but when I run this locally it looks ok. Does the screen recording use an old version of the code and has this indeed been fixed or is it a rendering issue in your browser? For reference, the current design was taken from https://material.io/components/chips#specs

- [x] In the suggestion list, the "headers" are quite similar to the items. I have looked for material design examples but haven't really found any. Maybe it helps to play with the top and bottom margin to leave more space between the "blocks" and but the header a little closer to the items? If all else fails, the items could be indented a little like in the current implementation.

- [x] Sorting is nice, but I'm not sure about the icons. I agree with @niknetniko that the default sort order should be shown. In the material design spec, an arrow icon is only shown next to the column that is currently sorted. Clicking on a table header sorts them (and adds the arrow), hovering over the header shows a lighter arrow to indicate that sorting will be applied when you click the header. I'm not sure is a similar behaviour is possible here since I think you added the behaviour to the icon itself. If we keep it linked to the icons, maybe we could use the up and down arrows to indicate the currently sorted column (without the 3 stripes) because now the icons are very similar.

- [x] In the action menu, long items are split over 2 lines which is not usually done.

(A lot of queries are happening on these pages, but I think this was already the case previously. It might be time for another n+1 query hunt.)

jorg-vr:
@bmesuere and @niknetniko The initial sort is often the index, which is not present as a column. Thus hard to visualize correctly.

Ideas are welcome for this

app/helpers/application_helper.rb
```diff
@@ -233,7 +233,15 @@ class AjaxLinkRenderer < ::ActionView::Base::LinkRenderer
     protected
 
     def html_container(html)
```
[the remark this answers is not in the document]
jorg-vr: initPagination() needs to be called after the pagination buttons are added in the page. 
These are often dynamically included in the page and reload. I tried to write code that triggered when the buttons appeared on the page, but failed to do so.

Including this script part in the HTML partial of this container assures it is executed whenever the html partial is added to the page

niknetniko (approved):
:+1:  (apologies for the small nitpicks)

bmesuere (changes requested):
This looks very good 👍 

The headers in the autocomplete are more clear now and the left alignment of the tokens, search and selects is good. To me, the chips are a bit too close to each other (especially compared to the generous spacing between the selects below). The spec mentions 8px in between chips but I couldn't find the current value.

- [ ] fix spacing between chips

---

Given this refactor, how hard would it be to store the selected filters for the exercise search within a course (on the series edit page) in local storage and restore them if present? In this case, I think teachers mostly use the same settings to find exercises (certain programming language/repo/language) in a certain course so this would speed things up for them.

---
For the sorting, I'm not convinced of the current solution. Clicking the header instead of the icon is easier but by removing the icon it becomes less discoverable. In addition, the lack of showing the default sort seems suboptimal. Would it be possible to manually add the default sort order (in code) so we could show the correct icon in the respective column? This would also improve the discoverability: if an arrow is shown next to a column header, this hints that you can sort them.

When sorting for the first time, the addition of a sort icon increases the header height and makes the entire header jump. (Adding the default order would solve this)

Do we need the third "blank" toggle item? This seems to make ik more complicated.

Because the icon is in front of the title, the label jumps to the right when hovering over it. I can see 2 possible fixes: reserve some blank space in front that's only filled in on hover/sort (but then the left alignment of the titles with the content breaks) or add the icon to the right of the title (against the spec but this seems acceptable).

If I understood correctly, you added the sort here as a proof of concept? We might have to discuss where to add sorting and where not. For example, sorting all submissions (or even all submissions within a large course) can take a lot of time. Because of the size of this PR, is it an option to disable sorting for now, get this merged and solve this in a subsequent PR?

- [ ] move sorting to different pr

jorg-vr:
It would be very doable to to store selected filters in local storage. An easy solution would be to store the last used search url, and use it for the existing setbasurl method to initialize search params on a new page load without search params. 

I'll make an issue for this

bmesuere (approved):

bmesuere:
I also think this removes the last major dependency on jQuery!

app/assets/javascripts/components/search_field.ts
```diff
@@ -0,0 +1,149 @@
...
+import { html, TemplateResult } from "lit";
+import { createDelayer } from "util.js";
+import { unsafeHTML } from "lit/directives/unsafe-html.js";
+import { ref } from "lit/directives/ref.js";
+import { searchQuery } from "search";
+import { ShadowlessLitElement } from "components/shadowless_lit_element";
+import { FilterCollectionElement, Label } from "components/filter_collection_element";
+
+@customElement("dodona-search-field-suggestion")
+export abstract class SearchFieldSuggestion extends FilterCollectionElement {
```
chvp, line 11: Maybe I'm missing something, but this is an abstract class that is seemingly never inherited from? If this is intended, a comment explaining why this is the case would be good IMO.

app/assets/javascripts/components/search_token.ts
```diff
@@ -0,0 +1,47 @@
+import { html, TemplateResult } from "lit";
+import { customElement, property } from "lit/decorators.js";
+import { ShadowlessLitElement } from "components/shadowless_lit_element";
+import { FilterCollection, FilterCollectionElement, Label } from "components/filter_collection_element";
+
+@customElement("dodona-search-token")
+export abstract class SearchToken extends FilterCollectionElement {
```
chvp, line 7: Same here.

app/assets/stylesheets/components/dropdown.css.scss
```diff
@@ -2,3 +2,7 @@
 .dropdown-menu a:hover {
   font-size: $font-size-base;
 }
+
+.dropdown-item-text{
```
chvp, line 6:
```suggestion
.dropdown-item-text {
```

app/assets/stylesheets/components/search_field.css.scss
```diff
@@ -0,0 +1,14 @@
+.search-filter:focus + .show-search-dropdown, .show-search-dropdown:focus, .show-search-dropdown:active, .show-search-dropdown:target {
+  display: block;
+  max-height: 400px;
+  overflow-y: auto;
+
+  .dropdown-header{
```
chvp, line 6:
```suggestion
  .dropdown-header {
```

app/assets/stylesheets/components/search_field.css.scss
```diff
@@ -0,0 +1,14 @@
...
+    padding: 0.75rem 0 0.25rem 1rem;
+  }
+
+  .dropdown-item{
```
chvp, line 10:
```suggestion
  .dropdown-item {
```

chvp (changes requested):

chvp (changes requested):
On the version that is currently deployed to Naos, course member label edit does not look good, and does not autocomplete either.

chvp (approved):

PR #3516, opened 2022-04-09
Hide submissions tab at start

### LINKED ISSUE #2886: Easy fetch last submission in hand in area
Marleen Denert wonders if it would be a useful feature to add a "fetch latest submission"- button when starting an exercise in the hand in area. This would allow for students to continue where they left off without much hassle.

In the current situation you have to start from the series:
> Go to series:
> image
> Modify submission:
> image

Suggested situation: notify the user in the hand in area that there is a previous submission available and they can continue where they left off.
> image

BTWS2:
This pull request makes the difference clear between new and already made exercises.

The number of submission doesn't matter that much, so I didn't add a badge. It doesn't look as clean with a badge too. The "fetch latest submission" feature might be something for the Dodona extension/plugin (https://github.com/thepieterdc/dodona-plugin-vscode/issues/92).

https://user-images.githubusercontent.com/56451049/162568798-03528795-ecdf-4781-a19d-1d9209acd2e6.mp4

*Ignore the internal error and submission link icon in this video.*

Closes #2886.

### DIFF
--- a/app/assets/javascripts/exercise.js
+++ b/app/assets/javascripts/exercise.js
@@ -349,6 +349,7 @@ function initExerciseShow(exerciseId, programmingLanguage, loggedIn, editorShown
             url += `&course_id=${data.course_id}`;
         }
         $.get(url);
+        $("#activity-submission-link").removeClass("hidden")
         $("#activity-submission-link").tab("show");
     }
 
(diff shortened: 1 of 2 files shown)


bmesuere:
Not a fan of this. A better empty state for the submissions tab is a better solution.

PR #3530, opened 2022-04-12
Integration with Dolos for plagiarism detection

rien:
This pull request adds support for plagiarism detection by creating an export and uploading it to Dolos.

It is currently only allowed for users on an allowlist.

- [ ] Tests were added

Closes #.

### DIFF
--- a/app/controllers/series_controller.rb
+++ b/app/controllers/series_controller.rb
@@ -236,6 +236,23 @@ class SeriesController < ApplicationController
     Submission.rejudge_delayed(@submissions)
   end
 
+  def plagiarism_check
+    @data = { item: @series,
+              list: @series.exercises,
+              course: @series.course,
+              table_header_type: t('exports.download_submissions.exercise'),
+              is_series?: true }
+  end
+
+  def create_plagiarism_check
+    @exercises = Exercise.where(id: params[:exercises])
+    @series.plagiarism_check_delayed(@exercises, current_user)
+  end
+
+  def plagiarism_check_wait
+    @export = Export.find(params[:export_id])
+  end
+
   private
 
   # Use callbacks to share common setup or constraints between actions.
--- a/app/models/export.rb
+++ b/app/models/export.rb
@@ -22,6 +22,14 @@ class Export < ApplicationRecord
   default_scope { order(id: :desc) }
 
   def start(item, list, users, params)
+    make_archive(item, list, users, params)
+
+    delay(queue: 'cleaning', run_at: AUTOMATICALLY_DELETE_AFTER.from_now).destroy
+    notification = Notification.new(user: user, message: 'exports.index.ready_for_download')
+    update(status: :finished, notification: notification)
+  end
+
+  def make_archive(item, list, users, params)
     bundle = Zipper.new(
       item: item,
       users: users,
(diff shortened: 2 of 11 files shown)


rien:
I need some input how we want to implement this. The main question that should be answered are:
- Where/how do we start plagiarism detection? (in Dolos? in Dodona? where in Dodona?)
- Where/how do we show the results? (same as above)
- How do we want to implement this "under the hood"

Dolos works best when analyzing submissions for a single exercise.

For context, the Dolos API currently works push-based like this:
1. Send a `POST` request to `https://dolos.ugent.be/api/reports` with multipart form arguments  `zipfile`, `name` and (optional) `programming_language`. The format of the ZIP-file is the same as the exports Dodona creates and Dolos understands the `info.csv` contained in the exports.
3. Similar to how submissions are processed by Dodona, Dolos starts a job to run the analysis and will directly respond to the upload request with a `id` and `URL` where the Report instance and its status  (`queued` / `running` / `finished` / `failed` / `error`) can be fetched as JSON.
4. The initiator (Dodona) can poll this URL until the status is final (`finished` / `failed` / `error`), this should not take more than 60 seconds, but is usually less than 10 seconds.
5. If the status is `finished`, the initiator can bring the user to `https://dolos.ugent.be/#/share/{report_id}` to show the results. Any errors are also inlcuded in the report json representation.

As an alternative, Dolos could also work pull-based where it would access the Dodona API to pull in information and exports using an API key.

## Proposal

To start off, I suggest the following approach:

- On the submissions page for an exercise, add a button "Check for plagiarism"
- Clicking this button will start a job which performs the following tasks:
  - Collect submissions and store them in an export, including an `info.csv`
  - Send the export through a `POST` to the Dolos API to initiate the analysis
  - Store the Report ID or URL in Dodona (but don't fetch it)
- In addition, the user is sent to a page for all plagiarism reports for that course
- If the user is on that page, the browser will poll the Dolos API to find out the status of the report
- Once the report is `finished` (or failed / errored), the polling stops and the user can click through to go to the report.

Instead of a polling by the client browser, we could also opt to let Dodona poll periodically through a worker (not desired, could hog worker availability), or find a hybrid approach where Dodona would poll only once after 60 seconds to find out what the final state of the report is.

Dodona should also be able to cope with reports being deleted. If this is detected, we could replace the "go to report" button with a "run analysis again" which would restart the above process.

@jorg-vr @bmesuere @chvp What are your thoughts on this?

jorg-vr:
Thanks Rien for the detailed writeout
I like the proposal.

I wasn't a fan of the polling at first, but it makes more sense if I think about it more.
It allows to keep al state in Dolos. This would mean that on every load of the plagiarism page in Dodona, all Dolos links are refreshed to get to most resent status.
This allows future statuses such as 'removed' or 'outdated' to be added should we ever do such changes to Dolos.
The downside is that if a teacher uses this feature a lot, we would be making a lot of requests to Dolos every page load. 

A possible alternative is Dodona providing a callback to Dolos upon initiation of the process. This callback is called by Dolos when the status is final. This would then update the status in the Dodona database.
The client would then be polling the dodona servers as long as the status is not finished. But at least the status is saved in the Dodona backend for future calls. Which would limit the number of requests needed for such a list page

rien:
Actually, there is already a `removed` status in Dolos: users can delete their data on request, and we reserve the right to delete data if we would run out of storage space.

I also like the idea of a callback when the state changes. This would indeed limit the amount of requests overall and minimizes the coupling between Dodona and Dolos. This also provides a way for Dolos to communicate when a report is removed.

jorg-vr:
We have discussed and approved out a new approach, see https://github.com/dodona-edu/dodona/issues/5338
The changes in this pr are no longer relevant

PR #3533, opened 2022-04-15
Create dismissable announcements for all dodona users

### LINKED ISSUE #2515: Show transient messages to users
We want a way to easily show messages to all users (or all course admins/staff/...). Such message would be displayed until it is dismissed or until we disable it.

This could be used to announce new releases, scheduled downtime, etc.

We want to be able to enable/disable this using an interface on Dodona and not doing a deploy.

jorg-vr:
This pull request introduces Announcements to dodona. Announcements are markdown alerts that can be created by zeus users and will be displayed for all specified dodona users.

Currently filtering is possible by institution and user group.
Announcements can also be scheduled using a start and end time.

User can always dismiss an announcement once they have read it.

Zeus overview of all announcements
image

Zeus option to create new announcement
image

Zeus option to edit announcements
image

Different alert styles and markdown can be used depending on the context.
image
image
image
image
image

Zeus users get an extra line with information on who will get the announcement and the option to edit it
image

- [x] Tests were added

Closes #2515 .


niknetniko (commented):
(I did a first review of the functionality, haven't looked at the code yet)

Some thoughts:

- [ ] The dropdown in the form for announcements is not styled correctly:
image
- [ ] The text box in the form should indicate markdown is supported, as we do in other places
- [ ] I would add a confirm dialog to the delete button in the announcement itself
- [ ] Mark as read doesn't work on the index page for announcements.
- [ ] I think the index page (for admins) is a bit much, since all announcements are also fully styled in that page. Perhaps a more "table"-like overview would be better? E.g. the mark as read button should not be in the overview page.
- [ ] Perhaps we also want a way to edit an announcement (e.g. to fix spelling mistakes)? Unsure if editing an announcement should reset the read status. If we add a warning this is the case on the edit page, it might not be needed to reset the read status (e.g. for fixing a spelling mistake, those who have read it already don't need to again)
- [ ] Announcements are wider than other alerts:
image

bmesuere:
We also want a Dutch and English text

app/models/announcement.rb
```diff
@@ -0,0 +1,65 @@
...
+    where("(start_delivering_at < ? OR start_delivering_at IS NULL)
+				AND (stop_delivering_at > ? OR stop_delivering_at IS NULL)", Time.current, Time.current)
+  }
+
+  scope :unread_by, lambda { |current_user|
+    if current_user.present?
+      joins("LEFT JOIN announcement_views ON
+      announcement_views.announcement_id = announcements.id AND
+      announcement_views.user_id = #{sanitize_sql_for_conditions(current_user.id)}")
+        .where('announcement_views.announcement_id IS NULL AND announcement_views.user_id IS NULL')
```
niknetniko, line 36: I think at least this second part can be written without strings, e.g. something like `.where(announcement_views: {announcement_id: nil, user_id: nil})`.

app/policies/announcement_policy.rb
```diff
@@ -0,0 +1,54 @@
...
+        @scope = scope.or(all_scopes.where(user_group: :staff)) if user.staff?
+        scope.is_active
+      else
+        scope.where(institution_id: nil).where(user_group: :everyone).is_active
+      end
+    end
+  end
+
+  def index?
+    true
```
niknetniko, line 20: I'm not sure if the index should be public or not.

app/policies/announcement_policy.rb
```diff
@@ -0,0 +1,54 @@
+class AnnouncementPolicy < ApplicationPolicy
+  class Scope < Scope
+    def resolve
+      if user&.zeus?
+        scope
```
niknetniko, line 5: Do we want to show all announcements to Zeus, or only those that are active? (Since only we can handle it, we can just delete old announcements of course, so perhaps this is something we want).

app/views/announcements/_form.html.erb
```diff
@@ -0,0 +1,87 @@
...
+      <%= f.label :style %>
+    </div>
+    <div class="col-sm-9">
+      <%= f.select :style, Announcement.styles.keys.to_a, {}, {:class => 'form-select'} %>
+    </div>
+  </div>
+<% end %>
+
+<script>
+  dodona.initDatePicker('#stop-delivering-at-group');
```
niknetniko, line 85: For some reason, these are always in English, even if I am using Dodona in Dutch.

config/locales/models/nl.yml
```diff
@@ -185,3 +185,17 @@ nl:
         description: Beschrijving
         maximum: Maximum
         visible: Zichtbaar
+      announcement:
+        text_nl: Nederelands aankondiging
```
niknetniko, line 189:
```suggestion
        text_nl: Nederlandse aankondiging
```

config/locales/models/nl.yml
```diff
@@ -185,3 +185,17 @@ nl:
...
+        text_en: Engelse aankondiging
+        start_delivering_at: Start aankondiging op
+        stop_delivering_at: Stop aankondiging op
+        user_group: Gebruikers groep
```
niknetniko, line 193:
```suggestion
        user_group: Gebruikersgroep
```

config/locales/models/nl.yml
```diff
@@ -185,3 +185,17 @@ nl:
...
+        user_groups:
+          everyone: Iedereen
+          all_users: Alle gebruikers
+          students: Studenten
+          staff: Lesgevers
+          zeus: Zeus
+        institution_id: Onderwijsinstelling id
+        style: Bootstrap stijl
```
niknetniko, line 201:
```suggestion
        style: Bootstrap-stijl
```

config/locales/models/nl.yml
```diff
@@ -185,3 +185,17 @@ nl:
...
+        start_delivering_at: Start aankondiging op
+        stop_delivering_at: Stop aankondiging op
+        user_group: Gebruikers groep
+        user_groups:
+          everyone: Iedereen
+          all_users: Alle gebruikers
+          students: Studenten
+          staff: Lesgevers
+          zeus: Zeus
+        institution_id: Onderwijsinstelling id
```
niknetniko, line 200:
```suggestion
        institution_id: ID van onderwijsinstelling
```

config/locales/models/en.yml
```diff
@@ -184,3 +184,17 @@ en:
...
+        start_delivering_at: Start delivering at
+        stop_delivering_at: Stop delivering at
+        user_group: User group
+        user_groups:
+          everyone: Everyone
+          all_users: All users
+          students: Students
+          staff: Staff
+          zeus: Zeus
+        institution_id: Institution id
```
niknetniko, line 199:
```suggestion
        institution_id: Institution ID
```

config/locales/views/announcements/nl.yml
```diff
@@ -0,0 +1,28 @@
+nl:
+  announcements:
+    new:
+      title: Maak een nieuwe aankondiging
+    edit:
+      title: Bewerk aankondiging
```
niknetniko, line 6:
```suggestion
      title: Aankondiging bewerken
```

config/locales/views/announcements/nl.yml
```diff
@@ -0,0 +1,28 @@
+nl:
+  announcements:
+    new:
+      title: Maak een nieuwe aankondiging
```
niknetniko, line 4:
```suggestion
      title: Nieuwe aankondiging
```

config/locales/views/announcements/nl.yml
```diff
@@ -0,0 +1,28 @@
...
+    edit:
+      title: Bewerk aankondiging
+    form:
+      markdown_html: <a href="https://docs.dodona.be/en/references/exercise-description/#markdown" target="_blank">Markdown</a> formaat wordt ondersteund.
```
niknetniko, line 8:
```suggestion
      markdown_html: <a href="https://docs.dodona.be/nl/references/exercise-description/#markdown" target="_blank">Markdown</a> wordt ondersteund.
```

config/locales/views/announcements/en.yml
```diff
@@ -0,0 +1,28 @@
+en:
+  announcements:
+    new:
+      title: Create new announcement
+    edit:
+      title: Edit announcement
+    form:
+      markdown_html: <a href="https://docs.dodona.be/en/references/exercise-description/#markdown" target="_blank">Markdown</a> formatting is supported.
```
niknetniko, line 8:
```suggestion
      markdown_html: <a href="https://docs.dodona.be/en/references/exercise-description/#markdown" target="_blank">Markdown</a> is supported.
```

config/locales/views/announcements/nl.yml
```diff
@@ -0,0 +1,28 @@
...
+    index:
+      title: Aankondigingen
+      new: Maak een nieuwe aankondiging
```
niknetniko, line 11:
```suggestion
      new: Nieuwe aankondiging maken
```

config/locales/views/announcements/nl.yml
```diff
@@ -0,0 +1,28 @@
...
+      start: Start
+      stop: Einde
+      number_of_reads: "#Gelezen"
+      text: Tekst
+      waiting: Gepland
+      active: Actief
+      stopped: Beeindigd
```
niknetniko, line 18:
```suggestion
      stopped: Beëindigd
```

config/locales/views/announcements/nl.yml
```diff
@@ -0,0 +1,28 @@
...
+      ago: "%{readable_time} geleden"
+      from_now: "over %{readable_time}"
+      users_from_institution: "%{users} van onderwijsinstelling %{institution}"
+    announcement:
+      dismiss: sluit aankondiging
```
niknetniko, line 23:
```suggestion
      dismiss: aankondiging sluiten
```

config/locales/views/announcements/nl.yml
```diff
@@ -0,0 +1,28 @@
...
+      shown_to_institution: Deze aankondiging wordt getoond aan %{user_group} van de onderwijsinstelling %{institution}
+      shown_to: Deze aankondiging wordt getoond aan %{user_group}
+      start_delivering_at: vanaf %{start_delivering_at}
+      stop_delivering_at: tot %{stop_delivering_at}
+      destroy: Verwijder deze aankondiging permanent
```
niknetniko, line 28:
```suggestion
      destroy: Aankondiging permanent verwijderen
```

db/migrate/20220413130927_create_announcement_tables.rb
```diff
@@ -0,0 +1,25 @@
+class CreateAnnouncementTables < ActiveRecord::Migration[7.0]
+  def change
+    create_table :announcement_views do |t|
+      t.integer :user_id
```
niknetniko, line 4: You first need to create the announcements table, and then these should be foreign keys, e.g. `t.references :user, foreign_key: true, null: false, type: :integer`.

db/migrate/20220413130927_create_announcement_tables.rb
```diff
@@ -0,0 +1,25 @@
...
+      unique: true,
+      name: 'announcement_view_index'
+    )
+    create_table :announcements do |t|
+      t.text :text_nl, null: false
+      t.text :text_en, null: false
+      t.datetime :start_delivering_at
+      t.datetime :stop_delivering_at
+      t.integer :user_group, null: false
+      t.integer :institution_id
```
niknetniko, line 20: This should also be a foreign key.

db/migrate/20220413130927_create_announcement_tables.rb
```diff
@@ -0,0 +1,25 @@
+class CreateAnnouncementTables < ActiveRecord::Migration[7.0]
+  def change
+    create_table :announcement_views do |t|
+      t.integer :user_id
+      t.integer :announcement_id
+      t.timestamps
+    end
+    add_index(
```
niknetniko, line 8: I don't think it is necessary to add a name to the index, rails will pick one for us.

niknetniko (changes requested):
The index page is much nicer than the previous version :+1: 
I personally would prefer top-aligning the table cells:
image

app/policies/announcement_policy.rb
bmesuere: I would also limit this to zeus-only

bmesuere (changes requested):
- instead of showing a remove action in the announcement for zeus, I would link to the edit page instead
- on the edit page, make it possible to remove?
- on the index page, showing the formatted announcement clutters the page really quickly. Maybe show the first x chars of the raw markdown instead?
- edit should have the option to reset all read-states

app/policies/announcement_policy.rb
jorg-vr: Yes, I would prefer zeus to be able to see al announcements. Maybe to reuse old announcements etc

app/views/announcements/index.html.erb
```diff
@@ -0,0 +1,94 @@
...
+              <% end %>
+              <th><%= t ".text" %></th>
+              <th></th>
+            </tr>
+            </thead>
+            <tbody>
+            <% @announcements.each do |announcement| %>
+              <tr class="announcement">
+                <td class="status-col">
+                  <%= style_icon announcement.style %>
```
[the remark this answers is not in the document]
jorg-vr: I added the border but kept the icon with tooltip, as it is hard to add a tooltip to a border

app/assets/javascripts/components/datalist_input.ts
```diff
@@ -0,0 +1,37 @@
+import { customElement, property } from "lit/decorators.js";
+import { html, TemplateResult } from "lit";
+import { ShadowlessLitElement } from "components/shadowless_lit_element";
+import { ref, Ref, createRef } from "lit/directives/ref.js";
+
+@customElement("dodona-datalist-input")
```
bmesuere, line 6: Can you add a little comment here to describe the component? 

Maybe we should look for a better way to document these components, something catalogue-like. I'll look if something like that already exists.

app/controllers/application_controller.rb
```diff
@@ -32,6 +32,7 @@ class ApplicationController < ActionController::Base
   before_action :set_time_zone_offset
 
   before_action :set_notifications, if: :user_signed_in?
+  before_action :set_announcement
```
bmesuere, line 35: don't we want to skip this for remote_requests? (I'm not sure why we also don't skip setting notifications in that case, maybe @chvp knows?)

app/views/announcements/_form.html.erb
```diff
@@ -0,0 +1,107 @@
...
+      <%= f.label :user_group %>
+    </div>
+    <div class="col-sm-9">
+      <%= f.select :user_group, Announcement.user_groups.map {|key, value| [Announcement.human_enum_name(:user_group, key), key]}, {}, {:class => 'form-select'} %>
+    </div>
+  </div>
+
+  <div class="field form-group row">
+    <div class="col-sm-3 col-form-label">
+      <%= f.label :institution_id %>
```
bmesuere, line 67: The label should now be institution instead of institution id

app/views/announcements/_form.html.erb
```diff
@@ -0,0 +1,107 @@
...
+  </div>
+
+  <% if announcement.id.present? %>
+    <div class="field form-group row">
+      <div class="col-sm-3 col-form-label">
+        <label for="reset_announcement_views"><%= t ".reset_announcement_views" %></label>
+      </div>
+      <div class="col-sm-9">
+        <div class="form-check">
+          <input id="reset_announcement_views" class="form-check-input" type="checkbox" name="reset_announcement_views">
```
bmesuere, line 95: We usually put the (long) label to the right of the checkbox in such cases:
image

app/views/announcements/mark_as_read.js.erb
```diff
@@ -0,0 +1 @@
+location.reload();
```
bmesuere, line 1: Why is this? Shouldn't mark as read just remove/hide the notification? This way, current page state could be lost.

app/views/layouts/_drawer.html.erb
```diff
@@ -164,6 +164,18 @@
               <% end %>
             </li>
           <% end %>
+          <% if current_user.zeus? %>
```
bmesuere, line 167: This should probably be a policy check for index.

config/locales/models/en.yml
```diff
@@ -184,3 +184,17 @@ en:
         description: Description
         maximum: Maximum
         visible: Visible
+      announcement:
+        text_en: English announcement
+        text_nl: Dutch announcement
+        start_delivering_at: Start delivering at
+        stop_delivering_at: Stop delivering at
```
bmesuere, line 191:
```suggestion
        start_delivering_at: Start showing at
        stop_delivering_at: Stop showing at
```

config/locales/views/announcements/en.yml
```diff
@@ -0,0 +1,29 @@
+en:
+  announcements:
+    new:
+      title: Create new announcement
+    edit:
+      title: Edit announcement
+      destroy: Permanently remove this announcement
```
bmesuere, line 7:
```suggestion
      destroy: Remove permanently
```

config/locales/views/announcements/nl.yml
```diff
@@ -0,0 +1,29 @@
+nl:
+  announcements:
+    new:
+      title: Nieuwe aankondiging
+    edit:
+      title: Aankondiging bewerken
+      destroy: Aankondiging permanent verwijderen
```
bmesuere, line 7:
```suggestion
      destroy: Definitief verwijderen
```

bmesuere (changes requested):

app/views/announcements/index.html.erb
```diff
@@ -0,0 +1,94 @@
+<div class="row">
```
bmesuere, line 1: How about removing the "state" icon and displaying 3 separate tables for scheduled, active and old announcements? If a category is empty, you could not display the table/header.

app/controllers/application_controller.rb
jorg-vr: Disabled announcements for remote requests, let me know if I need to do notifications to

app/views/announcements/mark_as_read.js.erb
jorg-vr: It is a bit more complex, as there can be another announcement to be displayed once this one is dismissed. But I'll write it better, because a state loss would indeed be an unwanted side effect

app/views/announcements/edit.html.erb
```diff
@@ -0,0 +1,24 @@
...
+        <h2 class="card-title-text"><%= t ".title" %></h2>
+        <div class="card-title-fab">
+          <%= render 'fab_button', form: "edit_announcement_#{@announcement.id}", icon: 'send' %>
+        </div>
+      </div>
+      <div class="card-supporting-text mt-3">
+        <div class="col-sm-12">
+          <%= render 'form', announcement: @announcement %>
+          <% if policy(@announcement).destroy? %>
+            <hr/>
```
niknetniko, line 14: This button should go in the card actions, which takes care of adding the line.

```html
<div class="card-actions card-border">
   <button>
</div>
```

config/locales/views/announcements/nl.yml
```diff
@@ -0,0 +1,34 @@
+nl:
+  announcements:
+    new:
+      title: Nieuwe aankondiging
+    edit:
+      title: Aankondiging bewerken
+      destroy: Definitief verwijderen
+    form:
+      markdown_html: <a href="https://docs.dodona.be/nl/references/exercise-description/#markdown" target="_blank">Markdown</a> wordt ondersteund.
+      reset_announcement_views: "Toon deze aankondiging opnieuw aan gebruikers die hem al gelezen hebben"
```
niknetniko, line 10:
```suggestion
      reset_announcement_views: "Toon deze aankondiging opnieuw aan gebruikers die ze al gelezen hebben"
```

niknetniko (changes requested):
Besides a few nit picks here and there, I'm not 100% convinced of the delete button in the edit form:
image

I feel like this button is placed where normally the OK or save button is placed. When editing an announcement, I have almost pressed it twice, while I intended to save the form.

niknetniko (approved):

bmesuere (approved):
I think splitting the table in 3 sections is a lot clearer 👍 

A final suggestion: if I'm correct, the dates and institution aren't required fields. Maybe add a little help text below those fields stating what will happen if the field is left blank (will show up immediately, will keep displaying until disabled manually, will show to users of all institutions).

FYI: you can group the acceptance of suggested changes in a single commit by adding them to the "batch" (but I think this only works from the Files changed tab).

config/locales/views/announcements/nl.yml
```diff
@@ -10,6 +10,9 @@ nl:
       markdown_html: <a href="https://docs.dodona.be/nl/references/exercise-description/#markdown" target="_blank">Markdown</a> wordt ondersteund.
       reset_announcement_views: "Toon deze aankondiging opnieuw aan gebruikers die ze al gelezen hebben"
       reset: Reset
+      start_delivering_at_help: "Aankondiging zal onmiddelijk getoond worden bij lege input."
```
pdawyndt, line 13: onmiddellijk

pdawyndt: Of: Niets invullen om aankondiging onmiddellijk te tonen.

PR #3549, opened 2022-04-21
Improve UI of summary row on series status overview

jorg-vr:
This pull request improves the UI of the summary row on series status overview.

It adds a mid dot separator:
image

Tooltips:
image

And adds context to the number in the users collumn:
image

### DIFF
--- a/app/views/series/_scoresheet_table.html.erb
+++ b/app/views/series/_scoresheet_table.html.erb
@@ -10,7 +10,7 @@
     </thead>
     <tbody>
     <tr class="summary-row">
-      <td class="user-name ellipsis-overflow"><%= t('series.scoresheet.users', :count => users.length) %></td>
+      <td class="user-name ellipsis-overflow"><%= t('series.scoresheet.users', count: users.length) %></td>
       <% activities.each do |activity| %>
         <% if activity.exercise? %>
           <td class="status">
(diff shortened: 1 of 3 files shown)


app/views/series/_scoresheet_table.html.erb
```diff
@@ -10,15 +10,22 @@
     </thead>
     <tbody>
     <tr class="summary-row">
-      <td class="user-name ellipsis-overflow"><%= users.length %></td>
+      <td class="user-name ellipsis-overflow"><%= t('series.scoresheet.users', count: users.length) %></td>
       <% activities.each do |activity| %>
         <% if activity.exercise? %>
           <td class="status">
+            <% i = 0 %>
             <% statuses.each do |status| %>
```
chvp, line 18: You can use `each_with_index` to avoid defining this `i` manually.

chvp (changes requested):

bmesuere (approved):
❤️

app/views/series/_scoresheet_table.html.erb
jorg-vr: But not every status will have a count greater than zero, so this would require filtering first

Can do that, but this was simpler

niknetniko (approved):

chvp (approved):

PR #3553, opened 2022-04-25
Fix authenticity token verification not being skipped for media requests

chvp:


jorg-vr (approved):

niknetniko (approved):

chvp:
The fix works on Naos: https://naos.ugent.be/nl/activities/1813154454/#

PR #3562, opened 2022-04-25
Make sure inputServiceWorker.js can be served in production

winniederidder:
This pull request tries to fix the asset resolution problem for the input service worker by choosing the correct assets in any environment.

### DIFF
--- a/app/controllers/activities_controller.rb
+++ b/app/controllers/activities_controller.rb
@@ -2,6 +2,8 @@ class ActivitiesController < ApplicationController
   include SeriesHelper
   include SetLtiMessage
 
+  INPUT_SERVICE_WORKER = 'inputServiceWorker.js'.freeze
+
   before_action :set_activity, only: %i[show description edit update media info]
   before_action :set_course, only: %i[show edit update media info]
   before_action :set_series, only: %i[show edit update info]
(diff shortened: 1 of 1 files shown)


jorg-vr (approved):

app/controllers/activities_controller.rb
```diff
@@ -2,6 +2,8 @@ class ActivitiesController < ApplicationController
   include SeriesHelper
   include SetLtiMessage
 
+  INPUT_SERVICE_WORKER = 'inputServiceWorker.js'.freeze
```
chvp, line 5: Why move this to a constant if it's used once?

app/controllers/activities_controller.rb
```diff
@@ -204,7 +206,8 @@ def media
   # Asset has been preprocessed and built internally
   # Redirecting to the asset is not possible due to browser security policy
   def input_service_worker
-    send_file(Rails.application.assets['inputServiceWorker.js'].filename,
+    assets = Rails.application.assets || Rails.application.assets_manifest.assets
```
chvp, line 209: Can we switch these? Doing less work in production seems preferable to me.

winniederidder: Switching doesn't seem to work. The assets_manifest.assets is not nil in development, so the asset is not found. A check would have to be done to see if asset resolution was succesfull in the assets_manifest.assets, so I don't think we will gain anything in that case.

chvp (approved):

PR #3563, opened 2022-04-25
Correctly handle each asset type

winniederidder:
This pull request tries to fix the asset resolution problem for the input service worker by choosing the correct assets in any environment and obtaining the correct filename.

### DIFF
--- a/app/controllers/activities_controller.rb
+++ b/app/controllers/activities_controller.rb
@@ -206,6 +206,9 @@ class ActivitiesController < ApplicationController
   # Asset has been preprocessed and built internally
   # Redirecting to the asset is not possible due to browser security policy
   def input_service_worker
+    # Which assets are available depends on the mode of the environment
+    # The else-block is only reachable in production
+    # :nocov:
     filename = if Rails.application.assets
                  Rails.application.assets[INPUT_SERVICE_WORKER].filename
                else
@@ -214,6 +217,7 @@ class ActivitiesController < ApplicationController
                    Rails.application.assets_manifest.assets[INPUT_SERVICE_WORKER]
                  )
                end
+    # :nocov:
     send_file(filename,
               filename: INPUT_SERVICE_WORKER,
               type: 'text/javascript')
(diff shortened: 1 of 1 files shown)


app/controllers/activities_controller.rb
```diff
@@ -204,9 +206,13 @@ def media
   # Asset has been preprocessed and built internally
   # Redirecting to the asset is not possible due to browser security policy
   def input_service_worker
-    assets = Rails.application.assets || Rails.application.assets_manifest.assets
-    send_file(assets['inputServiceWorker.js'].filename,
-              filename: 'inputServiceWorker.js',
+    filename = if Rails.application.assets
+                 Rails.application.assets[INPUT_SERVICE_WORKER].filename
+               else
+                 Rails.application.assets_manifest.assets[INPUT_SERVICE_WORKER]
```
chvp, line 212: This returns something of the form `"inputServiceWorker-abadf00e34692d3841e1f735d8f83f1d9879fd0b95ee577ded766655f9d28661.js"`. Just `send_file`-ing that will not work.

chvp (changes requested):

app/controllers/activities_controller.rb
winniederidder: Would something like Rails.root.join("app", "assets", "inputServiceWorker-abadf00e34692d3841e1f735d8f83f1d9879fd0b95ee577ded766655f9d28661.js") work, or will this file also be in the build subdirectory in production?

chvp: It will be in `public/assets`, but that doesn't really feel like a clean solution to me.

jorg-vr: `Rails.application.assets_manifest.directory` contains the directory the file is in.

jorg-vr: https://stackoverflow.com/a/61552216

chvp (approved):

bmesuere (approved):

jorg-vr (approved):

niknetniko (approved):

PR #3577, opened 2022-04-30
Add Safari-specific CSP headers to allow web worker to be loaded

winniederidder:
This pull request updates the CSP used for Papyros to allow loading the WebWorkers in Safari.
Closes https://github.com/dodona-edu/papyros/issues/136.

### DIFF
--- a/app/controllers/activities_controller.rb
+++ b/app/controllers/activities_controller.rb
@@ -27,7 +27,8 @@ class ActivitiesController < ApplicationController
   content_security_policy only: %i[show] do |policy|
     policy.frame_src -> { ["'self'", sandbox_url] }
     policy.worker_src -> { ['blob:', "'self'"] }
-    # Safari ignores worker_src and falls back to child_src
+    # Safari doesn't support worker_src and falls back to child_src
+    # This should be fixed in Safari 15.5 https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/worker-src#browser_compatibility
     policy.child_src -> { ['blob:', "'self'"] }
     # Allow fetching Pyodide and related packages
     # The data: urls is specifically to allow fetching the Python dependencies via a bundled tar that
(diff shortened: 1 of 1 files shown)


chvp (approved):

chvp:
@pdawyndt Can you test this on Naos on your iPad?

pdawyndt:
Naos Version works on iPad, Chrome and Safari

pdawyndt:
Standalone Version no langer load on iPad (entire Page keeps loading)

pdawyndt:
Previous issue seems resolved.

winniederidder:
> Standalone Version no langer load on iPad (entire Page keeps loading)

When encountering such things, it is most helpful to include the logs from the DevTools section. The hanging usually means something failed to load, which could even be the Pyodide files, where the only solution is a page reload.
Good to know that it resolved automatically, that means it was likely a CDN hiccup as I mentioned above.

pdawyndt:
No DevTools on iPad

winniederidder:
> No DevTools on iPad
Should be possible according to Google, e.g.
https://www.hongkiat.com/blog/ios-debugging/#:~:text=First%2C%20launch%20Safari%20and%20enable,%2C%20go%20to%20Settings%20%3E%20Safari.

chvp:
Can we keep the conversation here to what is relevant to this PR?

app/controllers/activities_controller.rb
```diff
@@ -27,6 +27,8 @@ class ActivitiesController < ApplicationController
   content_security_policy only: %i[show] do |policy|
     policy.frame_src -> { ["'self'", sandbox_url] }
     policy.worker_src -> { ['blob:', "'self'"] }
+    # Safari ignores worker_src and falls back to child_src
```
bmesuere, line 30:
```suggestion
    # Safari doesn't support worker_src and falls back to child_src
    # This should be fixed in Safari 15.5 https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/worker-src#browser_compatibility
```

bmesuere (approved):
I left a suggestion to add the corresponding mdn article in the comment so we know when to remove this later.

PR #3579, opened 2022-05-02
Add sortbuttons to submissions table

jorg-vr:
# Sort buttons
Interacting with this new system I made a new component that represents a sort button that can be used in table headers.
All buttons interact with a single state class `SortQuery`, which make sure only one sort column and direction is active any time and which itself interacts with  `SearchQuery` to provide the relevant sort param.

These sort params are handled in the backend by scopes defined in the controller and the model.

This pr only contains such scopes for submissions as an example, leaving further additions of such sort scopes to other models for a future pr. (This pr is already overloaded as is).

- [x] Added tests for sorting

https://user-images.githubusercontent.com/21177904/165749398-acfb1ef3-788a-4f01-90d3-2a8e47b9b4af.mp4

Closes # .

### DIFF
--- a/app/assets/javascripts/components/sort_button.ts
+++ b/app/assets/javascripts/components/sort_button.ts
@@ -1,0 +1,148 @@
+import "search.ts";
+import { css, html, LitElement, TemplateResult } from "lit";
+import { customElement, property } from "lit/decorators.js";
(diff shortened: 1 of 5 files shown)


jorg-vr:
closed in favor of #3580

PR #3580, opened 2022-05-02
Add sort buttons to submissions tables within the context of a course

jorg-vr:
# Sort buttons
Interacting with this new search query system introduced in #3498,  I made a new component that represents a sort button that can be used in table headers.
All buttons interact with a single state class `SortQuery`, which make sure only one sort column and direction is active any time and which itself interacts with  `SearchQuery` to provide the relevant sort param.

These sort params are handled in the backend by scopes defined in the controller and the model.

This pr only contains such scopes for submissions as an example, leaving further additions of such sort scopes to other models for future prs.

- [x] Added tests for sorting

https://user-images.githubusercontent.com/21177904/166434201-a37dd44c-f6d2-46c3-939a-cc2fe0a82e51.mp4


jorg-vr:
changes requested in https://github.com/dodona-edu/dodona/pull/3498#pullrequestreview-958611954

> For the sorting, I'm not convinced of the current solution. Clicking the header instead of the icon is easier but by removing the icon it becomes less discoverable. In addition, the lack of showing the default sort seems suboptimal. Would it be possible to manually add the default sort order (in code) so we could show the correct icon in the respective column? This would also improve the discoverability: if an arrow is shown next to a column header, this hints that you can sort them.

> When sorting for the first time, the addition of a sort icon increases the header height and makes the entire header jump. (Adding the default order would solve this)

> Do we need the third "blank" toggle item? This seems to make ik more complicated.

> Because the icon is in front of the title, the label jumps to the right when hovering over it. I can see 2 possible fixes: reserve some blank space in front that's only filled in on hover/sort (but then the left alignment of the titles with the content breaks) or add the icon to the right of the title (against the spec but this seems acceptable).

> If I understood correctly, you added the sort here as a proof of concept? We might have to discuss where to add sorting and where not. For example, sorting all submissions (or even all submissions within a large course) can take a lot of time. Because of the size of this PR, is it an option to disable sorting for now, get this merged and solve this in a subsequent PR?

niknetniko (approved):

bmesuere (approved):
👍  Looks good!

Mini nitpick: you have swapped the up and down icons. I was unsure myself, but they felt wrong. After a an extensive google study, the consensus seems to be that the up arrow is used for ascending sort.

As an aside: I was wrong in saying that according to the spec the sort icon should the to the left of the title (and that we thus explicitly break the spec here because it looks better). Instead, the spec says that the sort icon should be at the opposite side of the alignment of the title, so since our titles are left aligned, the icon should be at the right side which we do.

app/models/submission.rb
```diff
@@ -98,6 +98,11 @@ class Submission < ApplicationRecord
     correct.group(:exercise_id, :user_id).least_recent
   }
 
+  scope :order_by_user, ->(direction) { includes(:user).reorder 'users.first_name': direction, 'users.last_name': direction, id: :desc }
+  scope :order_by_exercise, ->(direction) { includes(:exercise).reorder "activities.name_#{I18n.locale}": direction, id: :desc }
+  scope :order_by_created_at, ->(direction) { reorder created_at: direction, id: :desc }
+  scope :order_by_status, ->(direction) { reorder status: direction, id: :desc }
```
chvp, line 104: As far as I can tell, the directions here are never validated to be one of `ASC` or `DESC`. While not security-critical as far as I can tell, it would trigger errors that we would get emails from, I think.

PR #3589, opened 2022-05-02
Upgrade Papyros to version 0.3.x

winniederidder:
This pull request integrates the latest version of Papyros.
This includes the following:

- Interrupts are handled more gracefully.
- Dark mode styling is more consistent
- Output overflows can be downloaded
- Error handling as described in https://github.com/dodona-edu/papyros/issues/136
- Close button is visible again
- More reliable import installing
- More reliable registering of service workers across browsers/url variants


bmesuere (changes requested):
The output area grows too large instead of scrolling on Chrome:
image

bmesuere (dismissed):

bmesuere (changes requested):
Sorry, I missed this before. A toast is not the right way to show this error message. Papyros is visible on screen so it makes a lot more sense to show the error there. In addition, isn't this something we want to add to papyros itself instead of the Dodona integration?

winniederidder:
> Sorry, I missed this before. A toast is not the right way to show this error message. Papyros is visible on screen so it makes a lot more sense to show the error there. In addition, isn't this something we want to add to papyros itself instead of the Dodona integration?

Implemented this in https://github.com/dodona-edu/papyros/pull/155. When approved, I'll integrate it here.

bmesuere (approved):

chvp (approved):

PR #3600, opened 2022-05-06
Allow easy re-use of annotations

### LINKED ISSUE #2825: Allow re-use of annotations
From https://github.com/dodona-edu/dodona/issues/1852#issue-599457917

chvp:
This pull request adds the re-use of annotations. Annotations are saved separately for re-use, with an extra added title.

## Visuals:
When creating a new annotation, you now have more options
image

You can search for an existing comment
image

This automatically loads the saved comment text, but you can still edit it as needed
image

You can save each annotation for reuse, this automatically suggest the first five words as a keyword title
image

There is info on the annotation showing it is created using a saved annotation
image

A side bar shows al saved annotations for this exercise with some basic info
image

You can edit a saved annotation for later reuse from this sidebar
image

You can also save an existing annotation, there is a button for this on non saved annotations
image

Which opens a similar popup
image

There is also a page on which you can manage all your existing saved comments
image

## Technical
As the management of saved annotations all happens on the evaluation page, I chose to limit the api to a json rest api. (No html or js delivered except the index page)
All visualization in the frontend happens using webcomponents. 
As displaying the saved annotations and manipulating them happens in different components loaded separately on the page, I created an external reactive state which triggers component rerenders when the underlying model changes.
I tried to write this state system in such a way that it would be relevant for any restfull resource.

I also created a generic modal webcomponent mixin to fix working with modals as webcomponents.

For the listing I created a generic pagination component that mimics the behavior of will_paginate. I added pagination meta data to the request headers.

I also had to edit the webpack config for this, because even with the commons js pack, `PubSub` was loaded multiple times, resulting in publishers publishing to an object without subscribers. The single runtime optimization solves this.

This newly created pack is negligible in size (1.55 kb before minimization)
image

I have also removed https://github.com/dpogue/rails_server_timings
It is no longer actively maintained and causes the following bug: https://github.com/dpogue/rails_server_timings/issues/6
It was already disabled in all config files except the test config (And I do not see its added value in an automatic test)

- [x] Tests were added

The plan is to first allow this in a (very) closed beta, so I will wait with writing documentation until the results of that closed beta are clear.

Closes #2825.


app/controllers/saved_annotations_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+    annotation = Annotation.find(params[:from])
+    authorize annotation, :show?
+    @saved_annotation = SavedAnnotation.new(permitted_attributes(SavedAnnotation).merge({ user: current_user, course: annotation.course, exercise: annotation.submission.exercise }))
+    authorize @saved_annotation
+    respond_to do |format|
+      if @saved_annotation.save
+        annotation.update(saved_annotation: @saved_annotation)
+        format.json { render :show, status: :created, location: @saved_annotation }
+      else
+        format.json { render json: @saved_annotation.errors.full_messages, status: :unprocessable_entity }
```
niknetniko, line 29: Is is easy to add a test with wrong data?

niknetniko (changes requested):
Functionally:

1. Saving an annotation at creation time by filling in the title doesn't work. It is currently not obvious that this isn't intended.
2. When editing an annotation that is linked to a saved annotation, the title of the saved annotation remains, but the saved annotation is not updated. I think this can be confusing.
3. Could we modify this a bit to make the title optional? I can imagine it wouldn't always be necessary, if we can allow autocomplete based on the body (but this is probably a bigger change, so not sure if it is worth it)

4. Finally, what is the benefit of doing for example the overview page client side? For editing/saving the annotations, it is obvious, since everything there is already client side. But for the overview page, the benefit is less obvious to me.

app/assets/javascripts/state/PubSub.ts
```diff
@@ -0,0 +1,57 @@
+/**
+ * heavily based on https://github.com/hankchizljaw/vanilla-js-state-management/blob/master/src/js/lib/pubsub.js
```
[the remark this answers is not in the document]
jorg-vr: I did look for a library but did not find something that was both small and satisfied my requirements.

For example the linked package uses and ugly notation  where methods are called as strings. (eg `storeInstance.dispatch('saySomething', textElement.value.trim());` )

It also isn't actively maintained.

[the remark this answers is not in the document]
niknetniko: In that case, there is no issue :)

jorg-vr:
> Saving an annotation at creation time by filling in the title doesn't work. It is currently not obvious that this isn't intended.

- [ ] Could this be as simple as replacing the placeholder and title text by `Search saved annotation` or `Reuse saved annotation`. Or do you have a better suggestion to improve this?

> When editing an annotation that is linked to a saved annotation, the title of the saved annotation remains, but the saved annotation is not updated. I think this can be confusing.

- [ ] This is indeed confussing: I'll remove the link to the saved annotation if the text is eddited

> Could we modify this a bit to make the title optional? I can imagine it wouldn't always be necessary, if we can allow autocomplete based on the body (but this is probably a bigger change, so not sure if it is worth it)

- [ ] This would be a bit of extra work, but if desired I could make autocomplete also work based on body.
- [ ] I would not allow empty titles, but we could suggest a cropped first part of the annotation by default?


> Finally, what is the benefit of doing for example the overview page client side? For editing/saving the annotations, it is obvious, since everything there is already client side. But for the overview page, the benefit is less obvious to me.

I made the listing overview in the sidebar first. This worked better client side, together with the state system to keep it always consistent.

I then made the seperate listing page also client side because it allowed me to reuse all the code I had already written. (But here there is no obvious other benefit)

niknetniko:
> Could this be as simple as replacing the placeholder and title text by Search saved annotation or Reuse saved annotation. Or do you have a better suggestion to improve this?

I think this is a good solution.

> but we could suggest a cropped first part of the annotation by default?

Also seems like a good solution to me.

chvp:
> > When editing an annotation that is linked to a saved annotation, the title of the saved annotation remains, but the saved annotation is not updated. I think this can be confusing.
> 
>     * [ ]  This is indeed confussing: I'll remove the link to the saved annotation if the text is eddited

It is an explicit goal to keep this link. Users might want to load a saved annotation and edit it to account for variable name choice by users. Keeping the link will allow us to analyze these links (and the code the annotations were made near to) in the future.

chvp (commented):
Looks good!

Some small comments:
 - [x] The saved annotations management page is only accessible via the annotation edit menu.
 - [x] I think we should replace the save button with a link icon or similar if the annotation is already saved?
 - [ ] The idea is to release this in a very closed beta. The limiting code for this still needs to be written. I think limiting on course ids will probably be the easiest?

niknetniko (approved):

chvp:
This year's courses taught by us. So `773` and `1151`.

chvp (commented):
Code and UI changes look good to me

bmesuere (changes requested):
### General
- It is not clear for me from using this feature if comments are linked to a course/exercise/user. All are valid choices, but a little help text might clarify this.

### Manage saved comments
- What is the goal of this page? It is not clear to me why there is a link to this page from the "add comment" form. (and for example no link in the save comments section)

### Saved comments section
- I would move the section with saved comments to its own card instead of a section in the existing card.
- Add an empty state to the saved comments (or don't show it at all if there are none)
- Remove the remove button and add it to the edit modal?
- Text overflow is used (which is good) but the title attribute (to trigger the tooltip) with the full text is missing
- Is it an option to show a second line (in a smaller font) with some statistics like "used xx times"

### Edit saved comment
- The modal dismissed by clicking on the backdrop. It is not clear of the changes will be saved in that case
- Add a remove button
- I know we use this form style on a lot of pages on Dodona, but in a modal it seems like a waste of space to put the title left instead of on top
- In the info text, maybe mention that changes to the text will not be applied to the already used instances of this comment

### Adding/reusing comment
- The "linked" icon looks very much like a button and is positioned next to a button which is not good. Putting it in the header row (next to the comment not visible icon) would be better.
- I'm not sure about the "save" flow. I would expect it to be a checkbox in the form (which shows the title input after checking)
- Could be a browser thing, but the title of a comment is shown twice in the "search" field, once big and once before the `:`:
  image
- Not really sure about the search and reuse UI. Because of the alignment, its hierarchy to the actual form seems ambiguous. For one, I would try to add some kind of title like "reuse save comment" to make the intent of the search box more clear. In addition, I would expect it to be at the top of the text area because in a top to bottom flow, you would first select a saved comment, then tweak the text and then click save.

jorg-vr:
> - It is not clear for me from using this feature if comments are linked to a course/exercise/user. All are valid choices, but a little help text might clarify this.

They are linked to course, exercise and user. And all those filters are currently active when trying to reuse comments. But I can think of usecases for less strict limiting of all those options:
- a course tries to teach certain principles, so some comments are relevant for the whole course (The exercise restriction does not make sense here)
- an exercise is reused every year, or a teacher gives multiple parallel course with the same exercise (The course restriction does not make sense here)
- Multiple teachers are grading the same exercise and want to give the same types of feedback (The user limit does not make sense here)

In future versions we can explore different forms of filtering maybe, but for now this pr is already big enough to not diversify here. 

- [ ] I'll add a help text explaining that saved comments are filtered by course, exercise and user

>### Manage saved comments
>
> - What is the goal of this page? It is not clear to me why there is a link to this page from the "add comment" form. (and for example no link in the save comments section)

Saved comments are also available outside the evaluation module. (Eg. when answering questions on a random submission within a course) Here there is no sidebar available. I discussed this with @chvp and she suggested a link to a separate page as a solution.
This separate page is also available through the sidebar and allows you to manage all saved annotations across all courses.  (Similar to the other global listings pages) It thus has more general than the sidebar (And also contains a bit more info for this reason)
- [ ] I'll add a link to this from the saved comments sidebar

>### Saved comments section

All good ideas :) I'll work on it

- [ ] I would move the section with saved comments to its own card instead of a section in the existing card.
- [ ] Add an empty state to the saved comments (or don't show it at all if there are none)
- [ ] Remove the remove button and add it to the edit modal?
- [ ] Text overflow is used (which is good) but the title attribute (to trigger the tooltip) with the full text is missing
- [ ]  Is it an option to show a second line (in a smaller font) with some statistics like "used xx times"

>### Edit saved comment
>
> - The modal dismissed by clicking on the backdrop. It is not clear of the changes will be saved in that case
- [ ] Disable modal closing by clicking backdrop

> - I know we use this form style on a lot of pages on Dodona, but in a modal it seems like a waste of space to put the title left instead of on top
- [ ] Change form style
- [ ] In the info text, maybe mention that changes to the text will not be applied to the already used instances of this comment

> ### Adding/reusing comment

- [ ] The "linked" icon looks very much like a button and is positioned next to a button which is not good. Putting it in the header row (next to the comment not visible icon) would be better.

> - I'm not sure about the "save" flow. I would expect it to be a checkbox in the form (which shows the title input after checking)

I don't have enough experience evaluating to know if saving an annotation is something you do while you write the annotation, or as an after thought, when you go back to copy paste it because you have spotted the same mistake again. (The first case would suggest your flow, while the second would suggest the current flow) @chvp do you have an opinion (As you originally put the current button where it is)

> - Could be a browser thing, but the title of a comment is shown twice in the "search" field, once big and once before the ::

This is actually an OS thing. But this is not ideal
- [ ] I'll rewrite our datalist component to a custom dropdown instead of the browser native thing. 

> - Not really sure about the search and reuse UI. Because of the alignment, its hierarchy to the actual form seems ambiguous. For one, I would try to add some kind of title like "reuse save comment" to make the intent of the search box more clear. In addition, I would expect it to be at the top of the text area because in a top to bottom flow, you would first select a saved comment, then tweak the text and then click save.
- [ ] Good point. I'll try to make a more logical design

chvp:
> I don't have enough experience evaluating to know if saving an annotation is something you do while you write the annotation, or as an after thought, when you go back to copy paste it because you have spotted the same mistake again. (The first case would suggest your flow, while the second would suggest the current flow) @chvp do you have an opinion (As you originally put the current button where it is)

When implementing, I imagined myself going back to a previous instance of an annotation to save it there when wanting to re-use. I think part of the reason for the closed beta is so that we can discover the most intuitive interface. It might be the case that when actually using the functionality this is not the way I'll use it.

jorg-vr:
@bmesuere I think I handled all your comments.

I'll add some more tests later, but would prefer to have your review first. (So I don't need to change the tests multiple times because of functionality changes)

app/assets/javascripts/saved_annotation_beta.ts
```diff
@@ -0,0 +1,7 @@
+// REMOVE FILE AND USAGES AFTER CLOSED BETA
+
+const BETA_COURSES = new Set([10, 773, 1151]);
```
chvp, line 3:
```suggestion
const BETA_COURSES = new Set([10, 773, 1151, 1659]);
```

app/policies/saved_annotation_policy.rb
```diff
@@ -0,0 +1,61 @@
+class SavedAnnotationPolicy < ApplicationPolicy
+  # REMOVE AFTER CLOSED BETA
+  BETA_COURSES = [10, 773, 1151].freeze
```
chvp, line 3:
```suggestion
  BETA_COURSES = [10, 773, 1151, 1659].freeze
```

app/assets/javascripts/components/annotations/annotation_form.ts
```diff
@@ -0,0 +1,247 @@
...
+    get type(): string {
+        return this.questionMode ? "user_question" : "user_annotation";
+    }
+
+    get rows(): number {
+        return Math.max(3, this.annotationText.split("\n").length + 1);
+    }
+
+    handleSavedAnnotationInput(e: CustomEvent): void {
+        if (e.detail.text) {
```
chvp, line 75: Will this ever not be the case?

bmesuere (approved):

app/assets/javascripts/components/pagination.ts
```diff
@@ -0,0 +1,72 @@
...
+    render(): TemplateResult {
+        return this.total > 1 ? html`
+            <center>
+                <ul role="navigation" class="pagination">
+                    ${!this.small ? this.pageButton(this.current - 1, "←") : ""}
+                    ${this.pageButton(1)}
+                    ${this.total > 3 + 2 * this.width && this.current == 3 + this.width ? this.pageButton(2) : ""}
+                    ${this.total > 3 + 2 * this.width && this.current > 3 + this.width ? this.pageButton(undefined, "...") : ""}
+                    ${this.range.map(i => this.pageButton(i))}
+                    ${this.total > 3 + 2 * this.width && this.total - this.current > 2 + this.width ? this.pageButton(undefined, "...") : ""}
```
chvp, line 64:
```suggestion
                    ${this.total > 3 + 2 * this.width && this.total - this.current > 2 + this.width ? this.pageButton(undefined, "…") : ""}
```

app/assets/javascripts/components/pagination.ts
```diff
@@ -0,0 +1,72 @@
...
+
+
+    render(): TemplateResult {
+        return this.total > 1 ? html`
+            <center>
+                <ul role="navigation" class="pagination">
+                    ${!this.small ? this.pageButton(this.current - 1, "←") : ""}
+                    ${this.pageButton(1)}
+                    ${this.total > 3 + 2 * this.width && this.current == 3 + this.width ? this.pageButton(2) : ""}
+                    ${this.total > 3 + 2 * this.width && this.current > 3 + this.width ? this.pageButton(undefined, "...") : ""}
```
chvp, line 62:
```suggestion
                    ${this.total > 3 + 2 * this.width && this.current > 3 + this.width ? this.pageButton(undefined, "…") : ""}
```

app/assets/javascripts/i18n/translations.js
```diff
@@ -0,0 +1,3 @@
...
...} · %{state}","resolve":"Mark as answered","send":"Ask question","unresolve":"Re-open question","update":"Update question"},"users":"users","violin_desc":"This graph shows the distribution of the number of submissions. The larger the zone, the more students submitted that number of solutions. The dot indicates the average. When the average is larger than 20, the dot turns hollow.","violin_title":"Number of submissions per user","weekdays":{"long":{"fri":"Friday","mon":"Monday","sat":"Saturday","sun":"Sunday","thu":"Thursday","tue":"Tuesday","wed":"Wednesday"},"short":{"fri":"Fr","mon":"Mo","sat":"Sa","sun":"Su","thu":"Th","tue":"Tu","wed":"We"}},"with":"with"},"time":{"am":"am","formats":{"annotation":"%B %d, %Y %H:%M","announcement":"%B %d, %Y %H:%M","default":"%a, %d %b %Y %H:%M:%S %z","flatpickr_long":"F j Y H:i","flatpickr_short":"m/d/Y H:i","long":"%B %d, %Y %H:%M","plain_time":"%H:%M","question":"%B %d, %Y %H:%M:%S","read_state":"%B %d, %Y %H:%M:%S","short":"%d %b %H:%M","submission":"%B %d, %Y %H:%M:%S","us":"%m/%d/%Y %I:%M %p"},"pm":"pm","units":{"day":["day","days"],"hour":["hour","hours"],"min":["minute","minutes"],"sec":["second","seconds"],"week":["week","weeks"]}}}'));
```
chvp, line 2: This file has been removed on develop?

chvp:
It seems weird to me that one can save an annotation when re-using a saved annotation. (During annotation creation.)

jorg-vr:
@chvp i was unsure about this. Potential usecase (why I allowed it) is if you want to start from a saved annotation, edit it and save it again as a variant

chvp:
Makes sense. Definitely fine to leave in for the closed beta.

PR #3636, opened 2022-05-18
Hide progress on series submission export for students when series progress is set to be hidden

### LINKED ISSUE #3635: Hide group progress in series export
When the group progress is hidden in a series, students can still see the data when downloading an export of the series. Progress fields should also be hidden from the series export.

jorg-vr:
Closes #3635 .

### DIFF
--- a/app/policies/series_policy.rb
+++ b/app/policies/series_policy.rb
@@ -130,7 +130,7 @@ class SeriesPolicy < ApplicationPolicy
   end
 
   def show_progress?
-    record.instance_of?(Series) && (record&.progress_enabled || course_admin?)
+    record.instance_of?(Series) && (record.progress_enabled || course_admin?)
   end
 
   private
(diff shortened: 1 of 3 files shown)


bmesuere (approved):

app/policies/series_policy.rb
```diff
@@ -129,6 +129,10 @@ def timeseries?
     statistics?
   end
 
+  def show_progress?
+    record.instance_of?(Series) && (record&.progress_enabled || course_admin?)
```
chvp, line 133: Either remove `record.instance_of?(Series)` or remove the `&`. `nil` is of `NilClass`, so `record` can never be `nil` if the `record.instance_of?(Series)` check succeeds.

niknetniko (approved):

chvp (approved):

PR #3637, opened 2022-05-20
Only show submissions without course in the submission history of a submission without a course

### LINKED ISSUE #3544: Only show submissions without a course in the submission overview on the submission detail page
For exercises that have submissions prior to numbering of submissions was introduced, there seems to be an off-by-one error

image
image

jorg-vr:
Closes #3544

### DIFF
--- a/app/controllers/submissions_controller.rb
+++ b/app/controllers/submissions_controller.rb
@@ -78,7 +78,7 @@ class SubmissionsController < ApplicationController
               end
     @submissions = @submissions.of_exercise(@submission.exercise)
     @submissions = @submissions.of_user(@submission.user)
-    @submissions = @submissions.in_course(course) if course.present?
+    @submissions = @submissions.in_course(course)
 
     @submissions_time_stamps = []
     prev = nil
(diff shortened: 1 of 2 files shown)


chvp:
How does this interact with a page like https://dodona.ugent.be/nl/activities/1388161683/submissions/? When we re-use previous exercises for examinations/evaluations we use pages like that to check that students in the current course did not submit to a certain exercise. They could have done that both inside and outside of a course, so it is important that we can see both types of submissions.

jorg-vr:
This code does not change anything to that page.

The change in [app/controllers/submissions_controller.rb](https://github.com/dodona-edu/dodona/pull/3637/files#diff-75cb0b991f85cc8e446dbc4ae10045abf7cb717a8d96e925d414203bb2c994bb) only affects submission show

The change in [app/models/submission.rb](https://github.com/dodona-edu/dodona/pull/3637/files#diff-223454796fa90590392d48862ea8513975ea96984835d7260c1d588b39b2235a) makes the in_course filter support nil values
I highly doubt that this page does any filtering on course, and other pages that do use the filter already should not have provided nil values before as these would have resulted in errors. 

Thus no behavior other than on the submission show page

bmesuere (approved):

chvp (approved):

PR #3640, opened 2022-05-20
Add summary row to course scoresheets

### LINKED ISSUE #3624: Add total column to scoresheets
This looks great. I think it would make a lot of sense to also add a "total" column to these scoresheets?

_Originally posted by @bmesuere in https://github.com/dodona-edu/dodona/pull/3597#pullrequestreview-972348655_

jorg-vr:
This pull request adds a summary row to course scoresheets. The summary contains the count of the users and the sum of the solved exercise per user for each series.

image

### DIFF
--- a/app/assets/javascripts/components/histogram.ts
+++ b/app/assets/javascripts/components/histogram.ts
@@ -12,11 +12,12 @@ export class Histogram extends LitElement {
             align-items: flex-end;
             height: 100%;
             width: 100%;
+            border-bottom: 1px solid var(--dod-histo-axis, lightgray);
         }
         .bar {
             display: inline-block;
-            background: var(--bs-blue);
             width: 100%;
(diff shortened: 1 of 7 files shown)


bmesuere:
I replaced the number by a custom element histogram:
image

The element takes an array with numbers and will always take up 100% width and height.

I do get a few console warnings:
> search.source.js:3262 Multiple versions of Lit loaded. Loading multiple versions is not recommended.

Probably because this is in a separate pack then the other custom elements

> search.source.js:3262 Element dodona-search-field scheduled an update (generally because a property was set) after an update completed, causing a new update to be scheduled. This is inefficient and should be avoided unless the next update can only be scheduled as a side effect of the previous update.

I think this was already there and is related to the existing custom elements

bmesuere:
superseded by #3644

PR #3643, opened 2022-05-21
Prevent Papyros CSS from overwriting our own styles

### LINKED ISSUE #3638: Tailwind css manipulates default styles on exercise pages
In light mode, on an exercise detail page, links are displayed black:
image

while they should be orange:
image

This was caused by inline css from tailwind: 
```css
/*  !tailwindcss v3.0.24 | MIT License | https://tailwindcss.com */
...
/* Reset links to optimize for opt-in styling instead of opt-out. */
a {
  color: inherit;
  text-decoration: inherit;
}
```

I have not found where and why we import tailwind. 
The link colors are correct on the submission detail page.

winniederidder:
This pull request removes the impact Tailwind css had on other parts of the pages where it was included.
layout

In order to deploy to Naos, I also had to set strscan to 3.0.1, otherwise deployment failed.
Closes #3638 .

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
     "@babel/plugin-transform-runtime": "^7.17.10",
     "@babel/preset-env": "^7.17.10",
     "@babel/preset-typescript": "^7.16.7",
-    "@dodona/papyros": "^0.3.7-scoped",
+    "@dodona/papyros": "^0.3.8",
     "@popperjs/core": "^2.11.5",
     "@rails/activestorage": "^7.0.3",
     "@rails/ujs": "^7.0.3",
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
     "@babel/plugin-transform-runtime": "^7.17.10",
     "@babel/preset-env": "^7.17.10",
     "@babel/preset-typescript": "^7.16.7",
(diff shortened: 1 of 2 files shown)


bmesuere (commented):
Why was gemfile.lock edited?

winniederidder:
> In order to deploy to Naos, I also had to set strscan to 3.0.1, otherwise deployment failed.

As mentioned in the top comment:
In order to deploy to Naos, I also had to set strscan to 3.0.1, otherwise deployment failed.

bmesuere (approved):

winniederidder:
Separate issue for the Gemfile change has been openend: #3645

chvp:
Do we still need a "special" release now that the corresponding PR in papyros has been merged?

winniederidder:
> Do we still need a "special" release now that the corresponding PR in papyros has been merged?

I could indeed also make a release on Papyros and then let dependabot handle it. I used a PR in this case as it was a high-priority fix and I needed to test it on Dodona.

chvp:
A separate PR is definitely good, but I would still prefer using a real release if possible (which it seems like it is now).

jorg-vr (approved):

chvp (approved):

PR #3644, opened 2022-05-22
Add histograms to the course scoresheet

bmesuere:
This pull request adds histograms to the course scoresheet. It is based on #3640, but I created a separate PR to be able to discuss a few custom element conventions in more detail.

image

- I added JSdoc comments to the element to make it more easily discoverable what the properties, events and css-properties are. I used the convention as described in [https://github.com/runem/web-component-analyzer#-how-to-document-your-components-using-jsdoc](https://github.com/runem/web-component-analyzer#-how-to-document-your-components-using-jsdoc). I also tried to use [storybook](https://storybook.js.org/) to document/test the component, but I couldn't get it working.
- To be able to style the element (with shadow root), I used CSS custom properties with a fallback. The naming convention I used was `d-histogram-variable` (`d` for dodona, `histogram` for the element name, `var` for the specific property).
- I used `dodona-histogram` as element name, following convention of the existing elements. However, we could also opt to use `d` as prefix instead of `dodona`.

Open questions for which a separate issue could be created:
- We now use lit elements from 2 separate packs. This means that the library is loaded twice. In dev-mode, you get a warning for this, but I don't think this will break in production. This is however not optimal. As far as I know, there are 2 possible solutions:
  - Import all our custom elements in the main pack (or a separate pack that is loaded on all pages)
  - Tweak our webpack config to create a separate pack/chunk for libraries that are used in multiple individual packs. This used to be possible using the [commons chunk plugin](https://webpack.js.org/plugins/commons-chunk-plugin/) which is no longer available. It was replaced by the [split chunks plugin](https://webpack.js.org/plugins/split-chunks-plugin/) but I have no experience with that.
- Do we load all polyfills/compilation steps for lit [correctly](https://lit.dev/docs/tools/requirements/#building-for-legacy-browsers) and do we need to? For the september release we will bump the minimum supported browser version so it might not be needed to do everything listed on that page.


jorg-vr:
image

The histogram makes to many columns (There is a column for 4, while that value never appears)

I would also add a max width, as it looks really stretched out with few series

chvp (approved):
I agree that some form of max width would be nice for courses with few series. Otherwise looks good.

bmesuere:
I added a max-width + center: 
image

jorg-vr (approved):

bmesuere:
I rebased this on develop and fixed a bug that was still preset at the time of previous review: the histogram now also takes the user filter into account.

PR #3646, opened 2022-05-22
Update to ruby 3.1

chvp:


chvp:
Matching ansible PR: https://github.com/dodona-edu/dodona-ansible/pull/145

test/controllers/users_controller_test.rb
```diff
@@ -70,9 +70,9 @@ class UsersControllerTest < ActionDispatch::IntegrationTest
     token_user = create :user, institution: nil
     token = token_user.token
 
-    get token_sign_in_user_url token_user, token: token
+    get token_sign_in_user_url(token_user, token:)
 
-    assert_redirected_to root_path
+    assert_redirected_to(root_path)
```
jorg-vr, line 75: why are brackets required here and not on the line below?

jorg-vr (approved):
I am not sure of I am one 100% fan of the new syntax, but I'll get used to it :)

bmesuere:
> I am not sure of I am one 100% fan of the new syntax, but I'll get used to it :)

I agree, I'm more fond of explicit naming, but I guess rubocop has a good reason to do is this was.

bmesuere (approved):
I of course agree with bumping this. Can we coordinate this easily or do we need to wait till after the exams (and coordinate this with the yearly server upgrade)

chvp:
> I of course agree with bumping this. Can we coordinate this easily or do we need to wait till after the exams (and coordinate this with the yearly server upgrade)

Coordination is quite easy. We can pre-install the new ruby version on our servers while keeping things running with the old version. A passenger restart (which happens as part of the deployment process) then switches to the new version.

chvp:
I'm pretty sure the style is configurable. If we prefer writing the keyword/hash args explicitely, we can probably configure it that way. This is just the default. I personally don't have a strong opinion either way.

chvp:
I changed the style to disallow the hash shorthand syntax. Didn't like it either.

PR #3663, opened 2022-05-24
Merge users when they try to sign in with the same username into the same institution

### LINKED ISSUE #3639: Merge users when they try to sign in with the same username or email into the same institution
Currently when a user tries to sign in with a different login method to the same institution, they get an error because a user with their email or user name already exists.

As both login methods belong to the same institution, we can assume that if the same credentials are returned this actually is the same user.

It would thus be helpful for the users to be merged automatically on sign in.

jorg-vr:
This pr contains a refactor of the try login method. This refactor was mainly done to improve code readability.

There is also one functional change in the code
If no identity was found, we try to find an existing user within the same institution that matches the given username or email. (in the old functionality we only searched for an existing user by email). This change was requested by #3639

We also added the email address that was given to us by the provider to the email already used warning.

After this pr there are still some issues present for a future pr. I listed these here: #3668

Closes #3639 .


app/controllers/auth/omniauth_callbacks_controller.rb
```diff
@@ -65,27 +65,43 @@ def try_login!
...
-    identity, user = find_identity_and_user
+    identity = nil
+    # First try to find an existing identity
+    if auth_uid.present?
+      # Basic case
+      identity = find_identity_by_uid
+    else
+      # For providers without auth uid
+      user = find_user_in_institution
+      identity = find_identity_by_user(user) if user.present?
```
bmesuere, line 76: I discussed this with @chvp and we couldn't find such cases in the database.

I would leave the case in for now, but add logging (e.g. to our event log). If nothing is logged after a few months, we can remove the case and simplify the code.

app/controllers/auth/omniauth_callbacks_controller.rb
```diff
@@ -65,27 +65,44 @@ def try_login!
...
+      logger.info "User #{user.id} logged in without auth_id using identity #{identity.id}" if identity.present? && user.present?
```
bmesuere, line 77: Sorry, I wasn't clear in my previous review. I meant logging it using https://github.com/dodona-edu/dodona/blob/develop/app/models/event.rb, not the rails logger.

bmesuere (changes requested):

PR #3664, opened 2022-05-24
Add summary column to course scoresheets

jorg-vr:
This pull request adds a summary column to course scoresheets. This column can be sorted.
Related to #3642

- [ ] Tests were added

### DIFF
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -221,6 +221,14 @@ class CoursesController < ApplicationController
       @users = apply_scopes(scores[:users])
       @series = scores[:series]
       @hash = scores[:hash]
+
+      @total_activity_count = @series.sum{|s| s.activity_count}
+      @total_by_user = Hash.new(0)
+      @users.each do |u|
+        @series.each do |s|
+          @total_by_user[u.id] += @hash[[u.id, s.id]][:accepted]
+        end
+      end
     end
 
     respond_to do |format|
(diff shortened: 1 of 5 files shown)


app/models/user.rb
```diff
@@ -149,6 +149,22 @@ class User < ApplicationRecord
...
+    combined_sql = course.series.map do |series|
+      submissions = Submission.in_series(series)
+      submissions = submissions.before_deadline(series.deadline) if series.deadline.present?
+      submissions = submissions.group(:user_id, :exercise_id).most_recent.correct.select(:user_id)
+      read_states = ActivityReadState.in_series(series)
+      read_states = read_states.before_deadline(series.deadline) if series.deadline.present?
+      read_states = read_states.select(:user_id)
+      "(#{read_states.to_sql}) UNION ALL (#{submissions.to_sql})"
+    end
+    combined_sql = "(SELECT user_id, COUNT(*) AS count FROM (#{combined_sql.join(' UNION ALL ')}) AS c GROUP BY user_id)"
```
bmesuere, line 164: I'm still not fond of sql in our models, but I guess there's no other way :(

bmesuere (approved):
Do we always take into account the deadline for these scoresheets? (the query would be a lot simpler if we didn't)

app/models/user.rb
jorg-vr: Yes it always takes into account deadlines, which indeed made this query way more complex than I would have wanted.

I did search for a way to write this without plain sql, but the union all made it a bit to complex

jorg-vr:
I merged the histogram pr, and added a histogram to the user column

bmesuere (approved):

niknetniko (approved):

chvp (approved):

PR #3669, opened 2022-05-26
Fix CSV differ crashing on CSV with a newline in a field on the first row

chvp:


chvp:
Should not have a performance impact, since parse_line makes the given string a StringIO object and only reads what's required to parse a single line.

niknetniko (approved):

PR #3683, opened 2022-05-30
Allow scrolling the page

winniederidder:
This pull request allows scrolling the exercise page while Papyros is being shown. When focused, scroll events are captured by the relevant component in Papyros (i.e. the editor, or the offcanvas itself).

These changes were requested by a user to allow reading the assignment without having to close and re-open the coding tab.


bmesuere:
I'm not a fan of this solution since breaks the UI pattern we're using. A better solution would be to reduce friction between toggling the scratchpad or to use a side-by-side when the screen is large enough. This will require a lot more work.

pdawyndt:
Side-by-side is what is seen in most browser-based programming courses. An alternative would be to use Papyros in another browser tab.

PR #3684, opened 2022-05-30
Update Papyros

winniederidder:
This pull request integrates the latest version of Papyros: v1.0.0

### DIFF
--- a/app/assets/stylesheets/components/scratchpad.css.scss
+++ b/app/assets/stylesheets/components/scratchpad.css.scss
@@ -59,31 +59,26 @@
 }
 
 // Wrapper and its child should take up as much space as possible
-#scratchpad-editor-wrapper,
-#scratchpad-editor-wrapper>div:first-child {
-    height: 100%;
-    max-height: 100%;
-}
-
-.cm-editor {
-    // Inform editor of height preferences
-    height: 100%;
-}
-
-// More specific selector to override !important property from Papyros
-div .cm-gutters {
-    // Height 100% seems to not work on the gutters, so override the default
-    height: auto !important;
-    min-height: 100% !important;
+#scratchpad-editor-wrapper , #scratchpad-editor-wrapper > div:first-child {
+    display: flex;
+    flex-flow: column;
+    flex: 1 1 auto;
+    overflow-y: auto;
 }
 
 #scrathcpad-output-wrapper {
     margin-bottom: 10px;
+    flex: 0 1 auto;
 }
 
 // First child is the output area, limit its size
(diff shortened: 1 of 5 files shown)


app/assets/stylesheets/components/scratchpad.css.scss
```diff
@@ -59,31 +59,26 @@
...
-    // Inform editor of height preferences
-    height: 100%;
-}
-
-// More specific selector to override !important property from Papyros
-div .cm-gutters {
-    // Height 100% seems to not work on the gutters, so override the default
-    height: auto !important;
-    min-height: 100% !important;
+#scratchpad-editor-wrapper , #scratchpad-editor-wrapper > div:first-child {
```
chvp, line 62:
```suggestion
#scratchpad-editor-wrapper,
#scratchpad-editor-wrapper>div:first-child {
```
Style was made uniform in #3694, please follow it.

app/assets/stylesheets/components/scratchpad.css.scss
```diff
@@ -73,8 +73,8 @@
 }
 
 // First child is the output area, limit its size
-#scratchpad-output-wrapper>div:first-child {
-    max-height: 20vh;
+#scratchpad-output-wrapper > div:first-child {
```
chvp, line 76:
```suggestion
#scratchpad-output-wrapper>div:first-child {
```

bmesuere:
In chrome on mac, there is a layout jump when running code:

[Curling---Dodona.webm](https://user-images.githubusercontent.com/481872/176170353-cd2964d4-535b-4ae3-ac39-427c25440c14.webm)

bmesuere (changes requested):
see comment above

bmesuere:
The reason is that the "switch input" link at the bottom disappears while the code is running. @winniederidder this probably needs to be fixed in papyros itself and not in the integration here?

winniederidder:
> The reason is that the "switch input" link at the bottom disappears while the code is running. @winniederidder this probably needs to be fixed in papyros itself and not in the integration here?

Indeed, this would require a small change in InputManager.ts to prevent hiding the link. The problem however with switching during a run is that we now have visual elements indicating the used lines and such, which would get out of sync. I disabled switching the mode during a run for that reason, as requiring all input handlers to be updated/active all the time when sending input goes against the purpose of having separate modes active at a time. I'm not entirely sure what the best solution would be for that, maybe a simpler solution is to somehow fix the height to prevent the rendering issue?

bmesuere (approved):

bmesuere:
Thanks @winniederidder, we'll release this early next week.

PR #3688, opened 2022-05-31
Force office 365 sign in to prompt user selection before sign in

jorg-vr:
This pull request forces microsoft to prompt the user to select their preferred microsoft account when signing in to dodona. This solves issues with people signing in using an unexpected account.

https://user-images.githubusercontent.com/21177904/171203881-53875c76-7047-429c-9986-957121518db8.mp4

### DIFF
--- a/config/initializers/office365.rb
+++ b/config/initializers/office365.rb
@@ -10,7 +10,7 @@ module OmniAuth
       option :name, 'office365'
 
       option :client_options,
-             site: 'https://login.microsoftonline.com/',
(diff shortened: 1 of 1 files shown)


bmesuere (approved):

niknetniko (approved):

chvp:
Wouldn't it make more sense to set that option somewhere in [this file](https://github.com/dodona-edu/dodona/blob/develop/config/initializers/office365.rb)?

niknetniko (approved):

Issue #3590, opened 2022-05-03
Tracking: JS modernization


bmesuere:
The goal of this issue is to track the "older" parts of our JavaScript code. There are 3 separate parts:
- Getting rid of jQuery: this is a big library and not really needed on the modern web
- Getting rid of the old application.js file: this will unify our build setup to only use the application pack
- Migrate to typescript: better type checking

## jQuery

- [x] The bootstrap token-field and typeahead still depend on jQuery. This is only used in the course label edit. #4083 
- [x] Almost all views contain code to initialize something that is wrapped in `$(function() {})`. This can probably be replaced by a jQuery-less equivalent  #3692
- [x] jQuery in the rails views
- [x] A lot of old JS code still uses jQuery. I think there a limited number of uses:
  - Selecting elements: can easily be replaced by native JS
  - Fetching: can also be replaced by a fetch
  - Events: should also be possible to refactor without too many issues
  - Adding HTML to pages, mostly by converting strings to html elements. There is no easy replacement for this, but the template system from lit could be reused here.

## application.js

- [x] i18n: is only there because this is loaded first. Can probably move to the pack without any problems once the rest is gone https://github.com/dodona-edu/dodona/pull/3841
- [ ] ~ace: not really straightforward on how to move this. We now use a ruby gem for this that hooks into sprockets. The library has a lot of individual files for programming languages that are loaded on demand so not sure how bundling would work.~
- [x] tokenfield/typeahead: Already removed everywhere #4189
- [x] first-input-delay: No longer used to my knowledge. Can probably be removed here (and from the package file) #3591 
- [x] #4360

```[tasklist]
### TypeScript
- [x] translations.js
- [x] https://github.com/dodona-edu/dodona/pull/3921
- [x] https://github.com/dodona-edu/dodona/pull/4083
- [x] https://github.com/dodona-edu/dodona/pull/4209
- [x] https://github.com/dodona-edu/dodona/pull/3930
- [x] https://github.com/dodona-edu/dodona/pull/4267
- [x] https://github.com/dodona-edu/dodona/pull/3963
- [x]  https://github.com/dodona-edu/dodona/pull/4618 
- [x] https://github.com/dodona-edu/dodona/pull/3593
- [x] https://github.com/dodona-edu/dodona/pull/4080
- [x] https://github.com/dodona-edu/dodona/pull/4463
- [x] https://github.com/dodona-edu/dodona/pull/4298
[...]

PR #3692, opened 2022-06-02
Remove jquery usage for detecting if the dom is ready

jorg-vr:
This pull request replaces the jquery trick used to execute code after the dom is loaded by a custom function with the same functionality.

This is an effort to get rid of jquery in the long run.

There are two functional changes:
 - In [app/views/courses/_not_a_member_dialog.html.erb](https://github.com/dodona-edu/dodona/pull/3692/files#diff-805a50d951308596eeffdd452c7391fdea19a733096e5b38aaa3854621ed0042) the javascript will be excecuted  a bit earlier. See [this stackoverflow thread](https://stackoverflow.com/questions/3698200/window-onload-vs-document-ready). If this difference in execution time was intended I can edit the code.
 - In [app/views/series/_series.html.erb](https://github.com/dodona-edu/dodona/pull/3692/files#diff-5d219fef158522548179e1c76d36504491f8234244eb2800ef57be6ca50e70dd) the code will now be executed onReady. It used to be executed immediately, but due to the way the code was written, it does not look like this was the intended behaviour.

Part of #3590


app/helpers/renderers/feedback_code_renderer.rb
```diff
@@ -28,7 +28,7 @@ def add_code
       end
       @builder.script(type: 'application/javascript') do
         @builder << <<~HEREDOC
-          $(() => {
+          window.dodona.onReady(() => {
```
chvp, line 31: Just a question: is there a reason for the window prefix here? It seems to be the only place you added this?

chvp (approved):

app/helpers/renderers/feedback_code_renderer.rb
jorg-vr: it was to be consistent with the code on following lines (most of the others were found by search replace, so  I did not always make that distinction based on local context)

PR #3685, opened 2022-05-30
Migrate to Material 3

bmesuere:
This pull request migrates Dodona to the Material 3 spec.

- [ ] Buttons
  - [x] Normal buttons
  - [ ] Fabs
  - [ ] Icon buttons
  - [ ] Segmented buttons
  - [ ] Migrate all existing buttons
- [ ] Chips
- [ ] Cards

PR #3695, opened 2022-06-05
Implement Material 3 colors

bmesuere:
This pull request adapts the material 3 colors as described in [here](https://m3.material.io/styles/color/the-color-system/tokens).

It uses the script from [this script](https://github.com/dodona-edu/m3-colors) to generate all tones.

## Changes
- Replace the old colors file with a new one.
- Change how we work with the bootstrap variables. Instead of reusing them in our code, we force our values to bootstrap and don't reuse them ourselves.
- Removal of the development style. This wasn't used anymore and could be re-added later on if we have proper theme support.
- The secondary color amber was replaced by pink. We don't use it a lot yet.

## Advantages
- We follow the spec. This also makes sure everything has proper contrast. This was not the case in the old design.
- Greatly improved dark mode (instead of a home brewed one).
- Preparation for switching to dark mode using css variables. This will also make it possible to have additional themes.
- The variable files are a lot cleaner. The color file contains the raw color values, the theme file maps them to primary, secondary, info, etc. The theme-light and theme-dark files map them to semantic colors such as "background" or "on-background".

### Home page
image
image

### Exercise
image
image

### Course
image
image


jorg-vr:
image

Background color of flatpicker instances is gone

jorg-vr:
There is more padding between tokens from the same filter than between different filters: (Probably caused by #3685 )
image

Big fan of the orange being replaced by blue in a lot of places, makes the pages a lot calmer :)
As suggested earlier I would also use these blue background buttons in the drawer.

Success announcement color is very bright compared to the others (I do like the other announcement colors)
image

I like the new greys. The darkmode is still a bit lighter than most other tools I use (eg github, slack) but it looks nice :)

Still not a fan of the brown buttons in dark mode, as well as the green text on green background in announcements and course cards.
image

You should probably also look at the color schemes of the visualizations. For example the blues in the heatmap are still based on the old main blue color
image

Don't know how much we can/want to change this, but the code editor in dark mode doesn't blend in nicely
image
It is much nicer in the code preview blocks
image

jorg-vr:
Hompage in darkmode is not very readable (Although I only get to see this page in darkmode after a forced sign out, so not that important)
image

I prefer a blue indicator here(as in the old style), now it attracts way to much attention
image

Same here 
image
I prefer the navbar style
image
But that could be personal preference

The grey her still looks a bit blue to me, I would prefer something like the background color. But maybe I just need to get used to it
image

Code blocks in darkmode should have a secondary background color (As they have in light mode)
image

app/assets/stylesheets/base.css.scss
```diff
@@ -2,19 +2,25 @@
...
 @import "vendor";
 
-// also loads dark mode stuff
-@import "theme/theme-light.css.scss";
-
 @import "theme/rouge-light.css.scss";
 @import "theme/rouge-dark.css.scss";
-@if $theme == "dark" {
+
+@if $theme =="dark" {
```
jorg-vr, line 19:
```suggestion
@if $theme == "dark" {
```

app/assets/stylesheets/base.css.scss
```diff
@@ -2,19 +2,25 @@
 @import "../../../node_modules/bootstrap/scss/functions";
 
 // 2. Include any default variable overrides here
-@import "theme/colors.css.scss";
+@import "theme/m3-theme-light.css.scss";
+@import "theme/m3-theme-dark.css.scss";
+
+@if $theme =="dark" {
```
jorg-vr, line 8:
```suggestion
@if $theme == "dark" {
```

app/assets/stylesheets/base.css.scss
bmesuere: my formatter changes this back :(

app/assets/javascripts/heatmap.ts
```diff
@@ -71,9 +71,9 @@ function initHeatmap(url: string, oldestFirst: boolean, year: string | undefined
 
 function drawHeatmap(data: [dayjs.Dayjs, number][], oldestFirst: boolean, year: string | undefined): void {
     const darkMode = window.dodona.darkMode;
-    const emptyColor = darkMode ? "#37474F" : "white";
-    const lowColor = darkMode ? "#364953" : "#E3F2FD";
-    const highColor = darkMode ? "#039BE5" : "#0D47A1";
+    const emptyColor = darkMode ? "#001d36" : "#fcfcff";
+    const lowColor = darkMode ? "#00325a" : "#d0e4ff";
+    const highColor = darkMode ? "#9ccaff" : "#003d6c";
```
jorg-vr, line 76: This code is duplicated in visualisations/timeseries.ts and should also be edited there

jorg-vr (approved):

bmesuere:
I will merge this so we can continue without merge conflicts on other branches. Your comments are still welcome, @chvp and @niknetniko.

PR #3710, opened 2022-06-10
Create custom theme for ace-editor

jorg-vr:
This pull request

- [ ] Tests were added

Closes # .

### DIFF
--- a/app/assets/javascripts/application_old.js
+++ b/app/assets/javascripts/application_old.js
@@ -16,7 +16,6 @@
 
 //= require ace-rails-ap
 //= require ace/ext-language_tools
-//= require ace/theme-twilight
 
 //= require bootstrap-tokenfield
 //= require typeahead.js/dist/typeahead.bundle.js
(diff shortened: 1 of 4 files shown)


bmesuere (approved):
👍 look good! Nice to have a consistent theme.

The only thing that looks a bit strange to me is the active line indicator in ace. Maybe it's because the difference in bg color is quite prominent (and in my editor it's very subtle)?

chvp:
I feel like a lot of visual information is lost. Before:
screenshot_2022-06-27-230013
After:
screenshot_2022-06-27-230052

The `method`, `self` and `test` on line 7 all have the same color. Same for the `for`, `i`, `in` and `range` on line 10.

The rouge style has similar problems:
screenshot_2022-06-27-230303
screenshot_2022-06-27-230249

(Don't get me wrong, I'm definitely in favor of a consistent code styling, but I think it's important we don't lose visual clarity in the process.)

chvp (changes requested):
(Marking this as request changes because I forgot to do an actual review.)

jorg-vr:
Hmm something went wrong somewhere. This Is what I get when I test this branch (Both local and on mestra)

Old
image
New
image

Old
image
New
image

jorg-vr:
could it be that you used the wrong programming language? Which causes an incorrect parsing of the semantics

chvp (approved):

PR #3711, opened 2022-06-10
Update darkmode theme

jorg-vr:
After discussing it shortly at lunch, I played around with the color scheme for dark mode.

I tested keeping the banner the same across both dark an light mode. I actually already got used to the light header, so I no longer have a big preference for either option. But as others also mentioned it being to light I wanted to provide a visual with a darker header.

I reintroduced green as success color. I also tried generating different greens, but did not find a better set.

I ended up improving the announcements and course cards by changing container color lightness from 30/90 to 40/100.

New
image
Old
image

New
image
Old
image

New
image
Old
image


bmesuere:
see #3723, tweaks to the success color in light mode and de feedback table will follow in a separate PR

PR #3722, opened 2022-06-13
Make darkmode darker following material 3 spec

jorg-vr:
This pull request makes the dark surfaces of darkmode darker. This follows the darkmode material 3 spec that uses tone 10 as background.

Surfaces also use tone 10 in the spec, but elevated surfaces use an overlay with the primary color and low opacity. Recreating an overlay as specified was hard, so I simulated the same effect by using the color.mix sass function.

New
image
Old
image
Merged with #3711
image
Combined with #3711 and #3710 
image


bmesuere:
see #3723.

PR #3725, opened 2022-06-17
Color code thumbs up/down on hover in evaluation 

### LINKED ISSUE #2715: Switch order of thumbs up/down buttons in grading panel

BTWS2:
I closed #3484 accidentally.

My goal with this approach is to add a visual confirmation the teacher pressed the right button.

Changing the `color` property in both states wasn't possible without javascript. This is the proposal using pure CSS: small color change on hover and clicking shows more green/red for clear visual feedback. When the teacher hovers over the wrong button, a glimpse of the wrong color is shown.

he end of the animation is cut off at the end, because the button is updated when the loading is done (visible at the end of the screen recording of dark mode below). I made the animation shorter and used `ease-out` to mitigate this partly. I added a space before the brackets and added a newline at the end.

At first I chose a lighter color (`$danger-100`), but it resulted in even more harsh colors in dark mode (e.g. `$danger-100: $red-800 !global;`), so I used opacity instead.

Light mode:

https://user-images.githubusercontent.com/56451049/161397648-62fee4d5-b0c6-433c-83c6-1bed54ca3b30.mp4

Dark mode:

https://user-images.githubusercontent.com/56451049/161397654-fadb9aa3-5833-4d10-8022-500b1ab24b8f.mp4

Closes #2715.


chvp (approved):

niknetniko (approved):

bmesuere (approved):

bmesuere:
thanks, @BTWS2 !

PR #3748, opened 2022-06-27
Set browser color scheme to dark in dark mode

bmesuere:
This pull request instructs the browser to use a dark color scheme when Dodona is set to dark mode. This causes the browser to use a different default set of styles which fixes some unstyled elements:

Before:
image

After:
image


jorg-vr:
Can something similar be done for light mode? (Forcing browser light mode defaults)

Screenshot from 2022-06-28 09-03-08

jorg-vr (approved):

bmesuere:
It seems not possible to style the select options sheet.

PR #3753, opened 2022-06-29
Don't allow users who are not subscribed to a course to see hidden series

chvp:


test/controllers/series_controller_test.rb
```diff
@@ -283,7 +283,7 @@ def assert_show_and_overview(authorized, token: nil)
   test 'student should see hidden series with token' do
```
jorg-vr, line 283:
```suggestion
  test 'unsubscribed student should not see hidden series, even with token' do
```

jorg-vr (approved):

bmesuere (approved):

PR #3760, opened 2022-07-01
Only allow submissions within the same course to be considered within an evaluation

jorg-vr:
This pull request adds extra filters and checks to make sure no submission from outside the course are used in an evaluation.

This fixes a bug that has appeared in our test environment with dummy data.

- [x] Tests were added


chvp:
I checked in our production database, and this has never happened in practice.

bmesuere (approved):

niknetniko (approved):

chvp (approved):

PR #3762, opened 2022-07-04
Add announcement in database seed with login tokens

jorg-vr:
This pull request adds an announcement with the login tokens for the default users to the database seed. This only impact or development environments. It improves usability of those environments.

image

I'll fix the broken homepage design with announcements in a separate pr.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -467,14 +467,8 @@ if Rails.env.development?
 
   puts "Add announcements (#{Time.now - start})"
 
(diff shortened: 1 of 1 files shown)


chvp (approved):

bmesuere (changes requested):
Since the banner will be present on all pages, I would reduce the height by putting the 3 links inline instead of li's

bmesuere (approved):

PR #3764, opened 2022-07-04
Fix about page header layout when it contains alerts

jorg-vr:
This pull request fixes the about page header layout when it contains alerts. The blue background now moves with the header.

It will work smoothly with alerts up to 380px height. 
Before
image

After
image

### DIFF
--- a/app/assets/stylesheets/pages/home.css.scss
+++ b/app/assets/stylesheets/pages/home.css.scss
@@ -11,8 +11,9 @@
 .home-header {
   position: relative;
 }
+
 .home-header:before {
-  content:"";
+  content: "";
   background: $primary-30;
   position: absolute;
   width: 100vw;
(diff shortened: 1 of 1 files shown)


bmesuere (approved):

app/assets/stylesheets/pages/home.css.scss
```diff
@@ -8,6 +8,21 @@
   z-index: -1;
 }
 
+.home-header {
+  position: relative;
+}
```
chvp, line 13: Does this follow the new CSS styling? I would expect a newline beneath and a space after `content:`. I still think we should look into automatic linting of our SCSS if we want to keep a consistent style. Maybe it won't be the exact style of @bmesuere's VSCode, but that seems less important than using some enforced/enforceable style.

niknetniko (approved):

PR #3796, opened 2022-07-08
Replace logos by versions with transparent background

### LINKED ISSUE #3780: Make background of UGent logos transparent
In light mode, UGent logos on support page have a white background, which differs from the background of the surrounding.

BTWS2:
This pull request closes #3780.


jorg-vr:
Could you add some visuals of how this looks (in both light and dark mode)

PR #3752, opened 2022-06-29
Add a setting to courses to limit registration to institutional accounts

jorg-vr:
This pull request adds a setting to courses to limit registration to institutional accounts. (Thus excluding personal accounts). This is also the new default setting for courses.

image

What a personal account sees:
image

- [x] Tests were added

Part of #2597
Closes #3802

PR #3802, opened 2022-07-12
Add a setting to courses to allow registration with personal accounts

jorg-vr:
This pull request is a replacement of #3752. It introduces a single extra toggle option to allow personal users to register for a course.

image

It is not very clear to me how this separate setting should interact with `open_for_institution`. Right now I take a union: everyone in this institution or without an institution. But this is an odd combination which I do not expect to see in practice.

- [x] Tests were added

Part of https://github.com/dodona-edu/dodona/issues/2597

### DIFF
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -342,26 +342,14 @@ class CoursesController < ApplicationController
           success_method = method(:signup_succeeded_response)
         end
 
-        case @course.registration
-        when 'open_for_all'
+        if @course.open_for_user?(current_user)
           if try_to_subscribe_current_user status: status
             success_method.call(format)
           else
             subscription_failed_response format
           end
-        when 'open_for_institution'
-          if @course.institution == current_user.institution
-            if try_to_subscribe_current_user status: status
-              success_method.call(format)
-            else
-              subscription_failed_response format
-            end
-          else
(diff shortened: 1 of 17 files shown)


niknetniko (approved):
Code looks good :+1: 

I assume there are good reasons to have a toggle and not another level (as was done in #3752), but this does mean users can create potentially weird situations. For example, if registration is `open_for_institution`, but personal accounts are enabled, this is functionally the same as `open_for_all`, as users can trivially create a personal account. (This is not really actionable, just a thought)

bmesuere (changes requested):
@niknetniko @jorg-vr As far as I'm aware (but I could be wrong), the direct "secret" link could be used to bypass the general settings. This is something we don't want for personal accounts: a teacher sharing the direct course link doesn't want students with personal accounts.

The toggle should be a checkbox. According to the spec, toggles are used for instantaneous actions.

We have been calling this "personal accounts" internally, but that might not be the best name to use in the application. In a way, all accounts are personal since they are not group accounts. "private accounts" might be a better name and including gmail and microsoft in the toggle/checkbox text might also help with making this more clear.

jorg-vr:
@bmesuere the secret is only there to bypass who can see the course. It does not bypass registration.

The setting introduced in this pr only changes who can register for a course. If you are not allowed to register, this is also impossible with a secret. (I included this case in the tests)

I will edit the checkbox and the text

jorg-vr:
Closed in favor of https://github.com/dodona-edu/dodona/pull/3752

PR #3803, opened 2022-07-12
Force users to accept the privacy policy

jorg-vr:
This pull request forces users to accept the privacy policy before they can use dodona.

On most pages users are redirected to this prompt, unless they have already accepted it:
image

- [x] Tests were added
- [ ] Update privacy policy text (split of into separate pr #3855)

Open questions:
- Do we want to include a (short) version of the privacy policy in the prompt?
- Do we want to force the prompt upon all users or only upon personal users?

Part of #2597


jorg-vr:
The way this is implemented right now this check happens after sign in (Or after account creation)

While writing the new privacy policy, I have come to the realization that it should probably be part of the sign up process. This way no account would be created if the privacy policy is not accepted..

I'll update this

jorg-vr:
closed in favor of #3859

PR #3806, opened 2022-07-13
Allow personal accounts

jorg-vr:
This pull request allows signing in with personal accounts using google or office 365.

For office 365 we had to move to the new oauth api v2.0 to support personal accounts. The identifier we used in the old api is no longer available, for this reason this pr introduces a migration for office 365 sign-ins. 

image

Code to be run in rails console after deploy:
```ruby
# Mark existing office 365 identities as legacy
Identity.joins(:provider).where(provider: {type: "Provider::Office365"}).update_all(identifier_based_on_email: true)

# create personal providers
Provider::Office365.create identifier: '9188040d-6c67-4c5b-b112-36a304b66dad', institution: nil
Provider::GSuite.create identifier: nil, institution: nil
```

I also made sure that private users no longer create a toke. (Only users without identities)

- [x] Tests were added

Part of #2597 
Should only be deployed after https://github.com/dodona-edu/dodona/pull/3802 and https://github.com/dodona-edu/dodona/pull/3803


config/locales/views/auth/nl.yml
```diff
@@ -8,7 +8,7 @@ nl:
       institution-search: "Typ om jouw school te vinden"
       ilearn-help: "i-Learn gebruikers kunnen momenteel enkel inloggen via de i-Learn website."
       higher-education: "Hoger onderwijs"
-      other-education: "Andere onderwijsinstellingen"
+      other-education: "Generieke login methodes"
```
jorg-vr, line 11:
```suggestion
      other-education: "Generieke loginmethodes"
```

chvp (approved):
Looks good. I'm wondering if we should change the buttons on the sign in page to something more generic, since Google Workspace is Google's enterprise offering.

config/initializers/office365.rb
```diff
@@ -30,14 +30,14 @@ def authorize_params
...
+      uid { raw_info['oid'] }
 
       info do
         {
             username: username,
-            first_name: raw_info['given_name'],
-            last_name: raw_info['family_name'],
-            email: raw_info['upn'],
+            first_name: raw_info['name'].split(' ').first,
+            last_name: raw_info['name'].split(' ').drop(1).join(' '),
```
niknetniko, line 39: What happens if the name doesn't contain a space, e.g. if the user has no last/first name? (Unsure which guarantees Microsoft gives about this field, so perhaps this is not possible)

niknetniko (approved):

config/initializers/office365.rb
jorg-vr: Microsoft does not give any guarantees on this field. Good catch. I'll test this and make sure it works

jorg-vr: image

It all works without issues

bmesuere (changes requested):
I probably reviewed this during a coffee break previously, so here is the written down version.

I'm not a fan of combining the sign in buttons for school logins and private logins even though they take you to a common page and can be combined. The reason is that in the current solution it isn't clear what the buttons do.

A new institutional user having a google workspace account might think that the google button won't work for them because it simply says Google. In the current design, it clearly says "educational institutions" so that user knows those buttons work for them.

There is also a potential problem the other way around. Since we currently don't support gmail sign ins, a user that already knows Dodona might not infer that we support them now.

Because of this and the marketing value of this feature (this should be the 6.0 highlight), I propose adding separate buttons by adding a section between educational accounts and other institutional accounts with the microsoft and google buttons (and rename the google one back to google workspace in the first section). After a while, we could then potentially merge the private account and other account sections.

jorg-vr:
I updated the page

bmesuere (approved):

PR #3807, opened 2022-07-15
Use default text buttons for user selection in the evaluation wizard

### LINKED ISSUE #3750: Unclear buttons in evaluation wizard
The buttons to select a preset of users are no longer obviously a button:
test

(Maybe an oversight of #3685?)

jorg-vr:
This pull request uses the default buttons styles for user selection. The whole card is no longer clickable as this is inconsistent with our general ux/ui.

Left button in the image is hovered:
image
image

Closes #3750 .

### DIFF
--- a/app/assets/javascripts/evaluation.ts
+++ b/app/assets/javascripts/evaluation.ts
@@ -100,8 +100,8 @@ export function initEvaluationStepper(): void {
                 if (!running) {
                     running = true;
                     event.preventDefault();
-                    const button = option.querySelector(".button");
-                    const loader = option.querySelector(".loader");
+                    const button = option.querySelector(".btn");
+                    const loader = option.parentNode.querySelector(".loader");
                     button.classList.add("hidden");
                     loader.classList.remove("hidden");
                     const response = await fetch(option.getAttribute("href"), { method: "POST" });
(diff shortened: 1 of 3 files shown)


bmesuere (approved):

niknetniko (changes requested):
I'm not sure this is the full solution. The `Gebruikers selecteren` text used to be a text button, as is also indicated by the markup:

```erb
<div class="button btn-text"><%= t('.select_users') %></div>
```

If we would replace `.button`  with `.btn`, which seems clearer to me that it is clickable.

image

bmesuere:
I agree with @niknetniko. It should indeed be btn:

> All buttons on Dodona should have the `btn` base class. Depending on the preferred style, it should have exactly one subclass: `btn-elevated`, `btn-filled`, `btn-tonal`, `btn-outline` or `btn-text`.

niknetniko (changes requested):
Unsure why exactly, but the buttons now produce js errors for me:

image

Perhaps some js depended on the `button` class?

chvp:
I would still leave the entire component as the button/link.

jorg-vr:
On top of the text being a button?
So we would have a button inside a button? (Although both do the same)

I think that would be a bit odd.

I originally just made the whole card a button (With a shadow to visualize the effect) But I actually prefer the current solution for its consistency

niknetniko (approved):

PR #3850, opened 2022-08-01
Add TypeScript type check

### LINKED ISSUE #3497: Run TypeScript's type checks
While we do use TypeScript, it is compiled to JavaScript with Babel.
This means type checking is not performed.

We probably want to include `tsc --skipLibCheck` as part of the CI checks.

Running this command right now gives me this:

```
Found 19 errors in 6 files.

Errors  Files
     7  app/assets/javascripts/evaluation.ts:4
     1  app/assets/javascripts/feedback/actions.ts:33
     3  app/assets/javascripts/heatmap.ts:39
     4  app/assets/javascripts/institution.ts:6
     1  app/assets/javascripts/question_table.ts:2
     3  app/assets/javascripts/score_item.ts:84
```

It does seem like most of the errors are either false positives, or wrongly typed instead of being a real bug, so I would not consider this a high priority.

BramDevlaminck:
This pull request solves typing issues in `assets/javascripts`

- I added the `@types/bootstrap` dependency to solve typing issues created by bootstrap
- type checking using TypeScript was added to the linting GitHub action
- readme update to mention the new test

Closes #3497.


.github/workflows/lint.yml
```diff
@@ -41,6 +41,9 @@ jobs:
     - name: Lint
       run: |
         yarn lint --quiet
+    - name: TypeScript type check
+      run: |
+        yarn tsc --skipLibCheck --noEmit
```
chvp, line 46:
```suggestion
        yarn typeCheck
```

chvp (changes requested):

app/assets/javascripts/types/index.d.ts
```diff
@@ -17,3 +17,5 @@ declare module I18n {
 }
 
 declare var dodona;
+
+declare var bootstrap: { Alert: any, Button: any, Collapse: any, Dropdown: any, Modal: any, Popover: any, Tab: any, Tooltip: any };
```
jorg-vr, line 21: Shouldn't this be fixed by the new dependency  `"@types/bootstrap"`?
If  `"@types/bootstrap"` does not fix the bootstrap typing errors, why do we still need it?

chvp: Or at least, shouldn't the `any`s here be replaced with the relevant constructor types (if possible)?

BramDevlaminck: > Shouldn't this be fixed by the new dependency `"@types/bootstrap"`?
> If `"@types/bootstrap"` does not fix the bootstrap typing errors, why do we still need it?

It seems that the `@types/bootstrap` dependency is used in `app/assets/javascripts/score_item.ts` on line 84 and 87 to make the `.collapse` possible. 

image

I don't immediately see another way of solving that type issue without the dependency. There is no obvious type that the element can be cast to that has the required properties

BramDevlaminck: > Or at least, shouldn't the `any`s here be replaced with the relevant constructor types (if possible)?

I'll take a look if this is possible. I just took inspiration from the `Window`-interface defined at the top of that file
```ts
declare interface Window {
    dodona: any;
    bootstrap: { Alert: any, Button: any, Collapse: any, Dropdown: any, Modal: any, Popover: any, Tab: any, Tooltip: any };
    MathJax: MathJaxObject;
}
```

chvp: That's probably because we didn't have `@types/bootstrap` yet. Now that we have it, that should be updated as well IMO.

BramDevlaminck: As a *fix* for the UMD global access error that kept showing up I allowed UMD global access since that is how it was done. Please let me know if we don't want this.

The problem is that to not use the UMD global access, a `bootstrap`-variable needs to be defined in `index.d.ts`. However, this makes accessing the `bootstrap`-namespace in that file impossible, which implies I can't access the types from `@types/bootstrap`.

app/assets/javascripts/types/index.d.ts
```diff
@@ -1,6 +1,6 @@
 declare interface Window {
     dodona: any;
-    bootstrap: { Alert: any, Button: any, Collapse: any, Dropdown: any, Modal: any, Popover: any, Tab: any, Tooltip: any };
+    bootstrap: { Alert: bootstrap.Alert, Button: bootstrap.Button, Collapse: bootstrap.Collapse, Dropdown: bootstrap.Dropdown, Modal: bootstrap.Modal, Popover: bootstrap.Popover, Tab: bootstrap.Tab, Tooltip: bootstrap.Tooltip };
```
jorg-vr, line 3: could this be reduced to bootstrap: bootstrap?

BramDevlaminck: Sadly not since the bootstrap used in `bootstrap.Alert` (and so on,...) is a namespace and not a type.

app/assets/javascripts/types/index.d.ts
jorg-vr: I think this is an ok fix. 

I am not a big fan of bootstrap being a not imported global variable in general (Which is why this modifier is needed if I understand it correctly?). But fixing that is probably out of the scope of this pr

jorg-vr (approved):

app/assets/javascripts/types/index.d.ts
niknetniko: You should be able to write it as `bootstrap: typeof bootstrap;`.

niknetniko (approved):

chvp (approved):

PR #3853, opened 2022-08-02
Create bash script for releasing

jorg-vr:
This pull request adds a shell script that automates releasing new versions.

Script can be run as `bin/release [version]`

Tested it for release 5.6.5

Script folows the commands from https://github.com/dodona-edu/wiki/blob/main/technical/release.md which I already copy pasted on release
The only changes is that I added some sed commands to edit `config/initializers/00_version.rb`
I also made it run all tests on every release
And I made sure the scripts exits if any single command fails

### DIFF
--- a/bin/release
+++ b/bin/release
@@ -22,7 +22,7 @@ version=$1
 # get major, minor and patch as an array
 readarray -d . -t version_array< <(printf '%s' "$version")
 
-if [ "${#version_array[@]}" -ne 3 ];
+if [ "${#version_array[@]}" -ne 3 ]
 then
   >&2 echo "expected version to contain major minor and patch"
   >&2 echo "usage \`bin/release [version]\`"
--- a/bin/release
+++ b/bin/release
@@ -8,7 +8,7 @@ trap 'last_command=$current_command; current_command=$BASH_COMMAND' DEBUG
 # echo an error message before exiting
 trap 'echo "\"${last_command}\" command failed with exit code $?."' EXIT
 
-if [ "$#" -ne 1 ];
+if [ "$#" -ne 1 ]
 then
   >&2 echo "Expected exactly 1 argument"
   >&2 echo "usage \`bin/release [version]\`"
--- a/bin/release
+++ b/bin/release
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 # exit when any command fails
 set -e
 # keep track of the last executed command
@@ -6,10 +6,28 @@ trap 'last_command=$current_command; current_command=$BASH_COMMAND' DEBUG
 # echo an error message before exiting
 trap 'echo "\"${last_command}\" command failed with exit code $?."' EXIT
 
+if [ "$#" -ne 1 ];
+then
+  >&2 echo "Expected exactly 1 argument"
+  >&2 echo "usage \`bin/release [version]\`"
+  >&2 echo "version should be of the format MAJOR.MINOR.PATCH"
+  >&2 echo "example \`bin/release 5.6.4\`"
+  exit 1
+fi
+
 # first and only argument is version
 version=$1
 # get major, minor and patch as an array
-readarray -d . -t version_array<<<"$version."
+readarray -d . -t version_array< <(printf '%s' "$version")
+
+if [ "${#version_array[@]}" -ne 3 ];
+then
+  >&2 echo "expected version to contain major minor and patch"
+  >&2 echo "usage \`bin/release [version]\`"
+  >&2 echo "version should be of the format MAJOR.MINOR.PATCH"
+  >&2 echo "example \`bin/release 5.6.4\`"
+  exit 1
+fi
 
 # Make sure you have the latest commits
 git checkout master
--- a/bin/release
+++ b/bin/release
@@ -1,0 +1,54 @@
+#!/bin/bash
+# exit when any command fails
+set -e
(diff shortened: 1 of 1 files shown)


bin/release
```diff
@@ -0,0 +1,54 @@
+#!/bin/bash
```
chvp, line 1:
```suggestion
#!/usr/bin/env bash
```

bin/release
```diff
@@ -0,0 +1,54 @@
...
+# exit when any command fails
+set -e
+# keep track of the last executed command
+trap 'last_command=$current_command; current_command=$BASH_COMMAND' DEBUG
+# echo an error message before exiting
+trap 'echo "\"${last_command}\" command failed with exit code $?."' EXIT
+
+# first and only argument is version
+version=$1
```
chvp, line 10: I would add a check before this that `$#` is exactly equal to 1 (and print a help message and exit otherwise).

bin/release
```diff
@@ -0,0 +1,54 @@
...
+# get major, minor and patch as an array
+readarray -d . -t version_array<<<"$version."
```
chvp, line 12: Also maybe add a check here that `version_array`'s length is exactly equal to three?

chvp (changes requested):

chvp:
Shellcheck is not entirely happy:

```
[Line 5:](javascript:setPosition(5, 6))
trap 'last_command=$current_command; current_command=$BASH_COMMAND' DEBUG
     ^-- [SC2154](https://www.shellcheck.net/wiki/SC2154) (warning): current_command is referenced but not assigned.
 
[Line 7:](javascript:setPosition(7, 6))
trap 'echo "\"${last_command}\" command failed with exit code $?."' EXIT
     ^-- [SC2154](https://www.shellcheck.net/wiki/SC2154) (warning): last_command is referenced but not assigned.
```

bin/release
```diff
@@ -0,0 +1,74 @@
...
+# exit when any command fails
+set -e
+# keep track of the last executed command
+current_command=$BASH_COMMAND
+last_command=$current_command
+trap 'last_command=$current_command; current_command=$BASH_COMMAND' DEBUG
+# echo an error message before exiting
+trap 'echo "\"${last_command}\" command failed with exit code $?."' EXIT
+
+if [ "$#" -ne 1 ];
```
chvp, line 11:
```suggestion
if [ "$#" -ne 1 ]
```

bin/release
```diff
@@ -0,0 +1,74 @@
...
+git checkout master
+git pull
+git checkout develop
+git pull
+
+# Start release
+git flow release start "$version"
+sed -i "s/MAJOR = [0-9]\+/MAJOR = ${version_array[0]}/g" config/initializers/00_version.rb;
+sed -i "s/MINOR = [0-9]\+/MINOR = ${version_array[1]}/g" config/initializers/00_version.rb;
+sed -i "s/PATCH = [0-9]\+/PATCH = ${version_array[2]}/g" config/initializers/00_version.rb;
```
chvp, line 44:
```suggestion
sed -i "s/MAJOR = [0-9]\+/MAJOR = ${version_array[0]}/g" config/initializers/00_version.rb
sed -i "s/MINOR = [0-9]\+/MINOR = ${version_array[1]}/g" config/initializers/00_version.rb
sed -i "s/PATCH = [0-9]\+/PATCH = ${version_array[2]}/g" config/initializers/00_version.rb
```

bin/release
```diff
@@ -0,0 +1,74 @@
...
+  >&2 echo "example \`bin/release 5.6.4\`"
+  exit 1
+fi
+
+# first and only argument is version
+version=$1
+# get major, minor and patch as an array
+readarray -d . -t version_array< <(printf '%s' "$version")
+
+if [ "${#version_array[@]}" -ne 3 ];
```
chvp, line 25:
```suggestion
if [ "${#version_array[@]}" -ne 3 ]
```

bin/release
```diff
@@ -0,0 +1,74 @@
...
+bundle exec rails test:system
+yarn test
+
+# Finish release
+git flow release finish "$version"
+
+# Allow abort before push
+echo -n "Next step will push, press A to abort, any key to continue"
+read -r confirm
+if [ "$confirm" == "A" ] || [ "$confirm" == "a" ];
```
chvp, line 62:
```suggestion
if [ "$confirm" == "A" ] || [ "$confirm" == "a" ]
```

chvp (approved):

chvp (changes requested):
The script exits with `"cap production deploy" command failed with exit code 0.` which is kinda weird.

bin/release
```diff
@@ -0,0 +1,74 @@
+#!/usr/bin/env bash
+# exit when any command fails
+set -e
+# keep track of the last executed command
+current_command=$BASH_COMMAND
+last_command=$current_command
+trap 'last_command=$current_command; current_command=$BASH_COMMAND' DEBUG
+# echo an error message before exiting
+trap 'echo "\"${last_command}\" command failed with exit code $?."' EXIT
```
chvp, line 9:
```suggestion
trap '[ $? -eq 0 ] || echo "\"${last_command}\" command failed with exit code $?."' EXIT
```

chvp (approved):

bmesuere (approved):
During the coffee break, we should brainstorm about releasing using a github action

PR #3857, opened 2022-08-03
Fix the native browser back button when using pagination

### LINKED ISSUE #3594: Cannot go back to initial pagination with back button
Since #3498, the pagination is added to the browser history (and URL). However, the first pagination does not seem to update when the back button is pressed in the browser and the params are removed from the URL.

For example, assume I do the following in the browser:

1. https://naos.ugent.be/nl/activities (initial navigation, page 1 is displayed)
2. https://naos.ugent.be/nl/activities?page=2 (OK, content is updated to page 2)
3. https://naos.ugent.be/nl/activities?page=3 (OK, content is updated to page 3)
4. https://naos.ugent.be/nl/activities?page=4 (OK, content is updated to page 4)

If I then begin pressing the back button:

1. https://naos.ugent.be/nl/activities?page=3 (OK, content is updated back to page 3)
2. https://naos.ugent.be/nl/activities?page=2 (OK, content is updated back to page 2)
3. https://naos.ugent.be/nl/activities (nothing happens, content stays on page 2)

The address bar does update: a page refresh will fix it.

BramDevlaminck:
This pull request should default the empty page value to page 1

Closes #3594 .

### DIFF
--- a/app/assets/javascripts/course.js
+++ b/app/assets/javascripts/course.js
@@ -296,7 +296,10 @@ function initCoursesListing(firstTab) {
 
     function initCourseTabs(firstTab) {
         document.querySelectorAll("#course-tabs li a").forEach(tab => {
-            tab.addEventListener("click", () => selectTab(tab));
+            tab.addEventListener("click", event => {
+                event.preventDefault(); // used to prevent popstate event from firing
+                selectTab(tab);
+            });
         });
 
         // If the url hash is a valid tab, use that, otherwise use the given tab
--- a/app/assets/javascripts/search.ts
+++ b/app/assets/javascripts/search.ts
@@ -113,7 +113,7 @@ export class SearchQuery {
         this.queryParams.subscribe(k => this.paramChange(k));
         this.queryParams.subscribeByKey("refresh", (k, o, n) => this.refresh(n));
 
-        window.onpopstate = e => e.state && this.setBaseUrl();
+        window.onpopstate = () => this.setBaseUrl();
 
         this.setRefreshElement(refreshElement);
     }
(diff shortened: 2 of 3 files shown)


BramDevlaminck:
checking the value of `e.state` is useless since this is always `null` when going back to the original page.
This is what prevented the page from refreshing.

One would expect that `e.state == true` since this is what is pushed on line 153 but this is not what happens.
Following information comes from [this](https://www.codeguage.com/courses/js/events-popstate-event) site.

> When we go back to our initial entry (after pushState() has been executed and the URL has been changed), popstate gets fired and the state property of the event object becomes null. This is the default value for every single document, so there's nothing really interesting in it.
> 
> From this point on, if we go forward to the pushed entry, popstate would fire again, but this time the state property of the event object would be {x: 1}, since the state of this pushed entry was configured to be {x: 1} at the time of executing pushState().
> 
> Simple!
> 
> Remember one thing: whenever navigation to an entry x is made, the state property of the event's object would be the same object set for that entry x.
> 
> This means that if, for example, you call pushState(), and set its state parameter to {x: 1}, and then go back to the previous entry, the dispatched popstate event's state property won't hold the object {x: 1}.
> 
> This is because when you go back, the entry to which you're taken is not the one that was set using pushState(). But from this point, if you go forward to this pushed entry, popstate would fire and its state property would be set to {x: 1}, since the state was set for this specific entry.

jorg-vr (approved):
Did you test on multiple browsers? because firefox and chrome don't have the same behavior for the popstate event

see https://developer.mozilla.org/en-US/docs/Web/API/Window/popstate_event#browser_compatibility

BramDevlaminck:
> Did you test on multiple browsers? because firefox and chrome don't have the same behavior for the popstate event
> 
> see https://developer.mozilla.org/en-US/docs/Web/API/Window/popstate_event#browser_compatibility

I've tested this in Firefox, Chrome and Safari and everything seems to work as expected.

chvp (approved):

PR #3862, opened 2022-08-05
Obscure reviewers name in feedback evaluation

### LINKED ISSUE #2717: Obscure reviewers
When publishing evaluation feedback (grades and annotations) the name of the reviewer is visible for the students. Since the grades and annotations should (in theory) not depend on the reviewer, we might provide an option to obscure the name of the reviewer towards the students, as we always obscure the name of the student for the reviewer in demo mode.

BramDevlaminck:
This pull request hides the name of reviewers for students in an evaluation.

The student sees the following information when the reviewer name is obscured:

image

image

- [x] Tests were added

Closes #2717  .


app/assets/stylesheets/theme/m3-colors.css.scss
```diff
@@ -291,6 +291,7 @@ $brown-100: #ffffff;
 
 // changed to be more gray
 $gray-0: #000000;
+$gray-15: #2b2b2f;
```
jorg-vr, line 294: If you add a new color, put it a the correct line :) Is this color in line with the material design specification? (these colors are all auto generated)

app/views/annotations/_annotation.json.jbuilder
```diff
@@ -7,7 +7,12 @@ json.rendered_markdown markdown(annotation.annotation_text)
 json.submission_url submission_url(annotation.submission, format: :json)
 json.url annotation_url(annotation, format: :json)
 json.user do
-  json.name annotation.user.full_name
+  # if we are NOT an admin and the evaluation is anonymous and we are not the person that typed the annotation => hide name
+  if !current_user.a_course_admin? && !annotation.evaluation.nil? && annotation.evaluation.anonymous? && (current_user.id != annotation.user.id)
```
jorg-vr, line 11: I would make this a function in the annotation model

You can shorten `!annotation.evaluation.nil? && annotation.evaluation.anonymous?`
to `annotation.evaluation&.anonymous?`

test/controllers/evaluations_controller_test.rb
```diff
@@ -617,4 +617,20 @@ def expected_score_string(*args)
...
+      }
+    }
+    @series.evaluation.update(anonymous: false)
+
+    assert_not @series.evaluation.anonymous
+
+    post set_anonymous_evaluation_path(@series.evaluation, anonymous: true, format: :js)
+    @series.evaluation.reload
+    assert @series.evaluation.anonymous
+  end
```
jorg-vr, line 635: Could you also write a test to see if the name is effectively hidden when anonymous is active.

(And maybe two further tests that course admins, and own annotations still show the correct name)

jorg-vr (changes requested):
For me it was not immediately clear which of these two buttons was active: 

image


Overall good work on this one :)

app/assets/stylesheets/theme/m3-colors.css.scss
BramDevlaminck: Oops, I did not mean to put it out of order :sweat_smile: .

How can I best check if this color is in spec? I *stole* it from the user-table in the same screen. It's the same grey that is used to indicate the selected-rows in the user table. Maybe I'm wrong, but this seems to be some default grey used by the table because I could not find it back in the stylesheet.

Screenshot 2022-08-16 at 10 49 25

BramDevlaminck:
> For me it was not immediately clear which of these two buttons was active:
> 
> image
> 
> Overall good work on this one :)

I agree that this colour is not very clear. The problem is that when using 1 shade darker it seems really dark and that colour is not yet used in that screen

image

As described in my response to the new colour, I tried to reuse the already present colours in the UI.

app/assets/stylesheets/theme/m3-colors.css.scss
bmesuere: Isn't it possible to reuse a (different) existing color? These scales are auto-generated so I would advise against adding new lightness values.

bmesuere (changes requested):
The indication of the "hide reviewers" setting is too prominent in the overview page. It is in a lager font and it is at the top of the page. In my opinion, it should be the same size as the sentence below (or even smaller) and I would put it below the existing sentence. I also wouldn't use an icon there because we don't do is for the other settings. If possible, add a link to change the setting.

jorg-vr:
For the button colors: I would look at the color scheme of elevated buttons https://m3.material.io/components/buttons/specs#0eea2a85-b4d7-4c74-b08e-98410b9412c7 as this is closest to the current button

config/locales/js/nl.yml
```diff
@@ -137,6 +137,7 @@ nl:
       copy-to-clipboard: Kopieer code naar klembord
     user_annotation:
       meta: "%{user} · %{time}"
+      anonymous_message: 'Dit commentaar is deel van een evaluatie'
```
niknetniko, line 140:
```suggestion
      anonymous_message: 'Deze opmerking is deel van een evaluatie'
```

niknetniko (approved):

app/assets/javascripts/code_listing/user_annotation.ts
```diff
@@ -85,6 +88,11 @@ export class UserAnnotation extends Annotation {
     }
 
     protected get meta(): string {
+        // when the reviewer name is hidden, the created time is just a message that shows that the annotation is from during the review
+        if (this.anonymous) {
+            return this.createdAt;
```
bmesuere, line 93: I'm confused by this. Does "created at" contain the message? Why?

app/assets/javascripts/evaluation.ts
```diff
@@ -20,13 +20,14 @@ export function initCheckbox(row: HTMLTableRowElement): void {
     });
 }
 
+
```
bmesuere, line 23: Aren't the changes in this file an artefact of the previous implementation? Can't they be removed or is there a good reason to keep them?

app/models/annotation.rb
```diff
@@ -50,6 +50,11 @@ class Annotation < ApplicationRecord
     end.reduce(includes(submission: [:exercise], user: []), &:merge)
   }
 
+  def anonymous(current_user)
+    # if we are NOT an admin and the annotation is part of an evaluation and we are not who wrote the annotation => hide name
+    !current_user.a_course_admin? && !evaluation.nil? && (current_user.id != user.id)
+  end
```
bmesuere, line 56: @jorg-vr I think this should be moved to the policy, right?

app/views/annotations/_annotation.json.jbuilder
```diff
@@ -1,13 +1,28 @@
...
 end
+
+# hide timestamp depending on evaluation and current user
+if annotation.anonymous(current_user)
+  json.created_at t('js.user_annotation.anonymous_message')
+  json.anonymous true
+else
+  json.created_at annotation.created_at
+  json.anonymous false
+end
```
bmesuere, line 14: Ok, now I see where the created_at came from. This is a bit of a hack and I would prefer a cleaner solution.

bmesuere (changes requested):

app/assets/javascripts/evaluation.ts
BramDevlaminck: These changes are indeed artefacts from the previous implementation. I left them because when I had to add another choice-option I had to rename all these variables because the original names *assume* that it's the only choice that will need to be made in that screen. It seemed better to me to leave this name-change in case that anyone would need to add something later.

If you prefer, I'll just change it back to the original names.

app/models/annotation.rb
jorg-vr: It could move there, but I don't have a strong opinion on it.

The method name should end on a question mark though, as it returns a boolean

BramDevlaminck: What should I do?

jorg-vr: Move the relevant check to the annotation_policy (Which manages who can see or edit what about annotations)

and add a question mark to the method name

app/policies/annotation_policy.rb
```diff
@@ -46,4 +46,11 @@ def permitted_attributes_for_create
   def permitted_attributes_for_update
     %i[annotation_text]
   end
+
+  def anonymous?
+    # if we are NOT an admin and the annotation is part of an evaluation and we are not who wrote the annotation => hide name
+    # "user" is the person that receives the annotation
+    # "record.user" is the person that wrote the annotation
+    !user.a_course_admin? && !record.evaluation.nil? && (user.id != record.user.id)
```
jorg-vr, line 54: We probably want to check if the user is course admin of this course (Not just any course)

app/views/annotations/_annotation.json.jbuilder
```diff
@@ -3,16 +3,34 @@ if annotation.is_a?(Question)
   json.extract! annotation, :question_state
   json.newer_submission_url(annotation.newer_submission&.then { |s| submission_url(s) })
 end
+
+# add anonymous message if the annotation is anonymised
+json.test annotation.user.id
```
jorg-vr, line 8: this test line is probably no longer needed?

jorg-vr (changes requested):

app/policies/annotation_policy.rb
BramDevlaminck: yes! thank you. I did not see the difference

jorg-vr (approved):

app/views/annotations/_annotation.json.jbuilder
```diff
@@ -3,16 +3,33 @@ if annotation.is_a?(Question)
   json.extract! annotation, :question_state
   json.newer_submission_url(annotation.newer_submission&.then { |s| submission_url(s) })
 end
+
+# add anonymous message if the annotation is anonymised
+json.anonymous_message t('js.user_annotation.anonymous_message') if policy(annotation).anonymous?
+
```
bmesuere, line 9: Is there a reason why this message is passed from the API? I would expect that this would just be a boolean and that the JS-side fetches the message. You are already using the js i18n.

config/locales/js/nl.yml
```diff
@@ -137,6 +137,7 @@ nl:
       copy-to-clipboard: Kopieer code naar klembord
     user_annotation:
       meta: "%{user} · %{time}"
+      anonymous_message: 'Deze opmerking is deel van een evaluatie'
```
bmesuere, line 140:
```suggestion
      anonymous_message: 'Deze opmerking maakt deel uit van een evaluatie'
```

app/views/annotations/_annotation.json.jbuilder
```diff
@@ -3,16 +3,33 @@ if annotation.is_a?(Question)
...
 json.rendered_markdown markdown(annotation.annotation_text)
 json.submission_url submission_url(annotation.submission, format: :json)
 json.url annotation_url(annotation, format: :json)
 json.user do
-  json.name annotation.user.full_name
-  json.url user_url(annotation.user)
+  # hide reviewer name depending on evaluation and current user
+  if policy(annotation).anonymous?
+    json.name ''
+    json.url ''
```
bmesuere, line 17: I discussed this with @chvp. The cleanest thing to do in this case is not return the field at all in the json output instead of setting an empty string. The same is true for last_updated below.

bmesuere (changes requested):
Can you request a review from @chvp when you updated the PR?

app/views/annotations/_annotation.json.jbuilder
bmesuere: I would generalize it and make it a boolean named part_of_evaluation. The client can then decide on how to handle that.

BramDevlaminck: Isn't it more logical to name the boolean something like `is_anonymized`? The client will never receive the name if the feedback is part of an evaluation. The only choice they have is which message they show to indicate this happened.

app/assets/javascripts/code_listing/user_annotation.ts
```diff
@@ -52,8 +53,8 @@ export class UserAnnotation extends Annotation {
     public readonly released: boolean;
     public readonly evaluationId: number | null;
     public readonly url: string;
-    public readonly user: UserAnnotationUserData;
-    public readonly lastUpdatedBy: UserAnnotationUserData;
+    public readonly user: UserAnnotationUserData | null;
+    public readonly lastUpdatedBy: UserAnnotationUserData | null;
```
chvp, line 57: These type annotations don't match the generated JSON in the jbuilder file. In the jbuilder file you don't send the fields in these objects. Sending `null` values in the jbuilder file seems like the correct way to do this.

chvp (changes requested):

app/assets/javascripts/code_listing/user_annotation.ts
bmesuere: @chvp didn't you suggest to not send the fields when we discussed this?

chvp: I didn't realize the fields were in objects at the time. Not sending the objects at all also sounds good, but not as nice to represent in the typescript definition I think?

BramDevlaminck: I've changed it to make the types match. The field `name` in the `UserAnnotationUserData`-object can now be null and in the json I send the explicit `nil` value for the fields that should be hidden when the feedback is part of an evaluation.

chvp: That was not one of the options 😉. Either leave the object out or send a null value in its place.

chvp: And I think @bmesuere prefers leaving the object out.

BramDevlaminck: > That was not one of the options 😉. Either leave the object out or send a null value in its place.

Ah, okay, I understood your message wrong. I'll change it back to leaving it out, my bad

app/policies/annotation_policy.rb
```diff
@@ -46,4 +46,11 @@ def permitted_attributes_for_create
   def permitted_attributes_for_update
     %i[annotation_text]
   end
+
+  def anonymous?
+    # if we are NOT the admin and the annotation is part of an evaluation and we are not who wrote the annotation => hide name
+    # "user" is the person that receives the annotation
```
chvp, line 52:
```suggestion
    # "user" is the person that is viewing the annotation
```

app/views/annotations/_annotation.json.jbuilder
```diff
@@ -3,16 +3,24 @@ if annotation.is_a?(Question)
...
+unless policy(annotation).anonymous?
+  json.user do
+    json.name annotation.user.full_name
+    json.url user_url(annotation.user)
+  end
 end
-json.last_updated_by do
-  json.name annotation.last_updated_by.full_name
-  json.url user_url(annotation.last_updated_by)
+# hide reviewer name depending on evaluation and current user
```
chvp, line 18:
```suggestion
# hide name of last editor depending on evaluation and current user
```

app/assets/javascripts/code_listing/user_annotation.ts
```diff
@@ -85,8 +86,12 @@ export class UserAnnotation extends Annotation {
     }
 
     protected get meta(): string {
+        if (!this.permissions.can_see_annotator) {
+            return I18n.t("js.user_annotation.anonymous_message");
+        }
+
         const timestamp = I18n.l("time.formats.annotation", this.createdAt);
-        const user = this.user.name;
+        const user = this.user?.name;
```
chvp, line 94:
```suggestion
        const user = this.user!.name;
```
(Using the non null assertion makes more sense semantically to me.)

chvp (approved):

app/assets/javascripts/code_listing/user_annotation.ts
BramDevlaminck: I don't really like the usage of `!`. I see what you mean that it makes more sense since the name should always exist when we get there, but it feels like asking for trouble. 

Wouldn't it be a better option to use `?`, but place a comment next to the line that shows the developer that name should always exist?

niknetniko: > Wouldn't it be better to use `?`, but place a comment next to the line that shows the developer that name should always exist?

This seems to me to be exactly the purpose of `!`; to indicate that the developer is certain a value will always be there and won't be null/undefined.

bmesuere (approved):

PR #3881, opened 2022-08-12
Use local storage to save filter in edit series to add exercises

### LINKED ISSUE #3578: Store selected filters in local storage for exercise search within a course
> Given this refactor, how hard would it be to store the selected filters for the exercise search within a course (on the series edit page) in local storage and restore them if present? In this case, I think teachers mostly use the same settings to find exercises (certain programming language/repo/language) in a certain course so this would speed things up for them.

It would be very doable to to store selected filters in local storage. An easy solution would be to store the last used search url, and use it for the existing setbasurl method to initialize search params on a new page load without search params. 

I'll make an issue for this

_Originally posted by @jorg-vr in https://github.com/dodona-edu/dodona/issues/3498#issuecomment-1114590314_

BramDevlaminck:
This pull request uses the local storage to keep track of the filters used in edit series to add exercises.

Closes #3578 .


app/assets/javascripts/search.ts
```diff
@@ -55,15 +55,15 @@ export class SearchQuery {
     arrayQueryParams: QueryParameters<string[]> = new QueryParameters<string[]>();
     queryParams: QueryParameters<string> = new QueryParameters<string>();
```
jorg-vr, line 57: I am not a fan of how you are passing around `localStorageKey` in every method. I would just set it as a class param here. And then you can use it in the relevant methods

app/assets/javascripts/search.ts
```diff
@@ -72,12 +72,16 @@ export class SearchQuery {
...
-    setBaseUrl(baseUrl?: string): void {
+    setBaseUrl(baseUrl?: string, localStorageKey?: string): void {
         this.updateAddressBar = baseUrl === undefined || baseUrl === "";
         const _url = baseUrl || window.location.href;
         const url = new URL(_url.replace(/%5B%5D/g, "[]"), window.location.origin);
         this.baseUrl = url.href;
 
+        // update the listeners with the new localStorageKey
+        this.queryParams.listeners = this.queryParams.listeners.map(() => (k => this.paramChange(k, localStorageKey)));
+        this.arrayQueryParams.listeners = this.arrayQueryParams.listeners.map(() => (k => this.paramChange(k, localStorageKey)));
```
jorg-vr, line 83: This will have unintended consequences. There are other listeners (outside this class) that will be overwritten this way.

app/views/layouts/_searchbar.html.erb
```diff
@@ -171,8 +171,18 @@
...
+        // if there is a previous value stored in local storage, load that value to reuse those frequently used preferences
+        const localStorageKey = "<%= local_assigns.fetch :local_storage_key, "" %>"
+
+        dodona.searchQuery.setRefreshElement("<%= local_assigns.fetch :refresh_element, "" %>", localStorageKey);
+
+        const prevDefinedSearchParams = localStorage.getItem(localStorageKey);
+        let url = "<%= local_assigns.fetch :baseUrl, "" %>"
+        // if there are params from localStorage, concatenate them to the url to use these stored search params
+        if (prevDefinedSearchParams) {
+          url += "?" + prevDefinedSearchParams;
```
jorg-vr, line 183: I don't like this string concat way of building the query. For example the baseurl could already contain some query params. We have helper methods for this in util.js 

Is there a reason you choose not to store the full url in local storage? This would avoid this issue.

jorg-vr (changes requested):

app/views/layouts/_searchbar.html.erb
BramDevlaminck: I did not store the whole url because this includes the id of the specific series:
```
http://dodona.localhost:3000/nl/series/47/available_activities/?programming_language=python&page=1&filter=curl&format=js
```
If I set the whole url the results of series with id 47 would be given and my assumption is that this would give different results.

jorg-vr: good call, I did not think of that. It is indeed better to only store the actual filters in this case

BramDevlaminck: I'll try to find a way to check if there are already search params in the base url. I might find something [here](https://developer.mozilla.org/en-US/docs/Web/API/URL) that helps with this.

app/assets/javascripts/search.ts
```diff
@@ -115,7 +118,7 @@ export class SearchQuery {
 
         window.onpopstate = () => {
             if (this.updateAddressBar) {
-                this.setBaseUrl();
+                this.setBaseUrl(localStorageKey);
```
chvp, line 121: This will pass the `localStorageKey` to the `baseUrl` param?

app/assets/javascripts/search.ts
```diff
@@ -194,6 +197,12 @@ export class SearchQuery {
                     eval(data);
                 }
                 document.getElementById("progress-filter").style.visibility = "hidden";
+            }).then(() => {
```
chvp, line 200: Why do this in a third `then`? It doesn't relay on any `Promise`d results from the previous step?

app/views/layouts/_searchbar.html.erb
```diff
@@ -171,8 +171,28 @@
...
+          const searchParamsFromUrl = urlObj.searchParams;
+          for (const [key, value] of searchParamsObj) {
+            // only apply the filter value from localStorage if it is not already present in the base url
+            if (!searchParamsFromUrl.has(key)) {
+              searchParamsFromUrl.set(key, value);
+            }
+          }
+          urlObj.search = searchParamsFromUrl.toString();
+        }
+        dodona.searchQuery.setBaseUrl(urlObj.pathname + urlObj.search, localStorageKey);
```
chvp, line 195: This feels a bit hacky. It's also a lot of actual code to put in the HTML templates. I would expect most of this to happen in `search.ts` as well.

chvp: Also, I would not expect the actual base URL to change? Just the pre-filled parameters?

chvp (changes requested):

app/views/layouts/_searchbar.html.erb
jorg-vr: If you put this code in search.ts, there is no need to reuse setBaseUrl. This way you can avoid having to build a new URL and just reuse the updateParam methods directly

app/assets/javascripts/search.ts
BramDevlaminck: My idea was that I only wanted to update the value in the `localStorage` if the fetch of the updated URL succeeded.

app/assets/javascripts/search.ts
```diff
@@ -72,7 +73,9 @@ export class SearchQuery {
         }
     }
 
-    setBaseUrl(baseUrl?: string): void {
+    setBaseUrl(baseUrl?: string, localStorageKey?: string): void {
```
chvp, line 76: Setting the `baseUrl` and the `localStorageKey` seems pretty independent. Would it make sense to add a `setLocalStorageKey` method?

jorg-vr, line 76: Is there a reason this is still in the setBaseUrl method? or is this just legacy?
I would expect a seperate setLocalStorage key method

app/assets/javascripts/search.ts
```diff
@@ -105,8 +104,8 @@ export class SearchQuery {
         }));
     }
 
-    constructor(baseUrl?: string, refreshElement?: string) {
-        this.setBaseUrl(baseUrl);
+    constructor(baseUrl?: string, refreshElement?: string, localStorageKey?: string) {
+        this.setBaseUrl(baseUrl, localStorageKey);
```
chvp, line 108: Where is this constructor used? If I look at the other changes in this PR, it doesn't seem like this last param will ever be used?

jorg-vr (changes requested):
Code looks good :) just one minor remark

app/assets/javascripts/search.ts
```diff
@@ -115,7 +114,7 @@ export class SearchQuery {
 
         window.onpopstate = () => {
             if (this.updateAddressBar) {
-                this.setBaseUrl();
+                this.setBaseUrl(null, localStorageKey);
```
chvp, line 117: If `setLocalStorageKey` becomes a separate method then this does not need to be updated.

app/assets/javascripts/search.ts
BramDevlaminck: it would make sense to have a separate `setLocalStorageKey` method, but this would mean that when another page uses the `searchQuery`-object **after** that a page used a `localStorageKey` and the new page does not use the local storage to store previous params. The localStorageKey will still be set and the local storage from the previous page will keep being changed with the params from the new page.

To prevent this I just combined these 2 to make sure that all the other places automatically reset the localStorageKey (because the `localStorageKey-argument` is empty there).

A solution to not break everything but still have a separate `setLocalStorageKey`-function would be to
- always reset the `localStorageKey` to be `null` when `setBaseUrl` is called
- manually call `setLocalStorageKey` AFTER `setBaseURL` when the local storage is used

What do you think?

BramDevlaminck: Jorg had the same question, I answered it [here](https://github.com/dodona-edu/dodona/pull/3881#discussion_r949083662) and proposed a possible solution

app/assets/javascripts/search.ts
BramDevlaminck: It is indeed never used right now, I added it to give the option to do this immediately. I'll remove it since this can easily be set in another function

app/assets/javascripts/search.ts
jorg-vr: This should not be the case. If search query is used on another page, then all code in searchbar.html.erb should be rerun setting al relevant parameters. For example `refreshEllement`  is also different across all pages, so this would break if only `setbaseurl` was updated.

BramDevlaminck: ah okay. In that case it will indeed be a lot cleaner to separate these functions

app/assets/javascripts/search.ts
```diff
@@ -194,8 +195,50 @@ export class SearchQuery {
...
+    /**
+     * fetch params from localStorage using the localStorageKey if present and apply them to the current url
+     */
+    useLocalStorage() : void {
+        if (this.localStorageKey) {
+            const searchParamsStringFromStorage = localStorage.getItem(this.localStorageKey);
+            if (searchParamsStringFromStorage) {
+                const searchParamsFromStorage = new URLSearchParams(searchParamsStringFromStorage);
+                // don't overwrite currently set params with params from the localStorage
+                const currentSearchParams = new URL(this.baseUrl).searchParams;
```
jorg-vr, line 216: This part won't work as this.baseurl is set as url.href (Which does not contain any params)

All params are stored in `this.arrayQueryParams` and `this.queryParams`. So this is where you would look if one has already been set in the current state.

app/assets/javascripts/search.ts
```diff
@@ -87,13 +88,13 @@ export class SearchQuery {
...
-        for (const key of url.searchParams.keys()) {
-            if (key.endsWith("[]")) {
-                this.arrayQueryParams.updateParam(key.substring(0, key.length-2), url.searchParams.getAll(key));
-            } else {
-                this.queryParams.updateParam(key, url.searchParams.get(key));
-            }
-        }
+        this.initialiseParams(url.searchParams);
+        // apply parameters from local storage
+        this.useLocalStorage();
```
jorg-vr, line 93: I would expect this code to be in ` setLocalStorageKey` instead of `setBaseUrl`

jorg-vr (changes requested):

chvp (approved):

app/assets/javascripts/search.ts
```diff
@@ -194,8 +195,49 @@ export class SearchQuery {
...
+     * fetch params from localStorage using the localStorageKey if present and apply them to the current url
+     */
+    useLocalStorage() : void {
+        if (this.localStorageKey) {
+            const searchParamsStringFromStorage = localStorage.getItem(this.localStorageKey);
+            if (searchParamsStringFromStorage) {
+                const searchParamsFromStorage = new URLSearchParams(searchParamsStringFromStorage);
+                // don't overwrite currently set params with params from the localStorage
+                searchParamsFromStorage.forEach((_value: string, key:string) => {
+                    if (this.queryParams.params.get(key) !== undefined || this.arrayQueryParams.params.get(key) !== undefined) {
```
jorg-vr, line 217: array query params in the url will contain `[]` at the end, while they are stored without `[]` in the `arrayQueryParams` object.

see `intialiseParams` for example code

`this.arrayQueryParams.params.get(key) !== undefined` will thus not have the desired effect

jorg-vr (changes requested):

jorg-vr (approved):

bmesuere (approved):
👍  nice work

PR #3899, opened 2022-08-18
Fix the wrapping of very long identifiers on the profile page

jorg-vr:
image

I also fixed users being shown as token users in the dev environment.


bmesuere (approved):

chvp (changes requested):
I would also add a title attribute to get a tooltip with the full text if we are going to cut it off sometimes.

PR #3901, opened 2022-08-19
Give users a friendlier message about why they cannot register for a course

jorg-vr:
This pull request reuses the hidden icon title text in the callout, to give a better explanation of why they cannot register.

image

image
image


bmesuere (approved):
👌

chvp (approved):

PR #3902, opened 2022-08-19
Automatically filter on courses for which you can register when relevant

jorg-vr:
This pull request filters the courses list on courses for which the user can register. It also adds some extra icons to visualize whether you are already registered.

The filter is optional. It it automatically activated when accessing the courses list from the home page. (Either 'more courses' or 'view courses') But it can be disabled through the search options.

image
image

- [x] test where added

### DIFF
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -16,6 +16,10 @@ class CoursesController < ApplicationController
     scope.by_course_labels(value, controller.params[:id])
   end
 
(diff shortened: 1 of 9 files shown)


chvp (approved):

bmesuere (changes requested):
- On naos, I can only see 5 courses in "all courses" as zeus user
- Is it possible to only enable it on the all courses page? Right now, even the "my courses" list is empty which is unwanted behaviour.

bmesuere (approved):

PR #3918, opened 2022-08-22
Clear all grades button

### LINKED ISSUE #2716: Clear grades

BramDevlaminck:
This pull request adds a button to clear all score items in the feedback panel
The change to the top level 👍 and 👎 is not (yet) implemented since we weren't sure if we wanted this.

screenshots:

image

image

Closes #2716 (partially) .

### DIFF
--- a/app/assets/javascripts/feedback/actions.ts
+++ b/app/assets/javascripts/feedback/actions.ts
@@ -268,7 +268,12 @@ export default class FeedbackActions {
                     f.markBusy();
                     f.data = "";
                 });
-                const values = this.scoreForms.map(f => f.getDataForNested(true));
+                const values = this.scoreForms
+                    .map(f => f.getDataForNested())
+                    .map(record => {
+                        record.score = null;
+                        return record;
+                    });
                 this.update({
                     // eslint-disable-next-line camelcase
                     scores_attributes: values
--- a/app/assets/javascripts/feedback/actions.ts
+++ b/app/assets/javascripts/feedback/actions.ts
@@ -102,11 +102,11 @@ export default class FeedbackActions {
         this.scoreForms.forEach(s => s.disableInputs());
     }
 
-    update(data: Record<string, unknown>, clearScores: boolean): Promise<void> {
+    update(data: Record<string, unknown>): Promise<void> {
         this.disableInputs();
         return fetch(this.options.currentURL, {
             method: "PATCH",
-            body: JSON.stringify({ feedback: data, clear_scores: clearScores }),
+            body: JSON.stringify({ feedback: data }),
             headers: {
                 "Content-Type": "application/json",
                 "Accept": "text/javascript"
@@ -187,7 +187,7 @@ export default class FeedbackActions {
             if (autoMark && hasAutoMark) {
                 await this.update({
                     completed: true,
-                }, false);
+                });
             }
             if (skipCompleted) {
                 window.location.href = this.options.nextUnseenURL;
@@ -239,7 +239,7 @@ export default class FeedbackActions {
             await this.update({
                 // eslint-disable-next-line camelcase
                 scores_attributes: values
-            }, false);
+            });
         });
         this.allScoresMaxButton?.addEventListener("click", async e => {
             e.preventDefault();
@@ -252,7 +252,7 @@ export default class FeedbackActions {
             await this.update({
                 // eslint-disable-next-line camelcase
                 scores_attributes: values
-            }, false);
+            });
         });
         this.deleteAllButton?.addEventListener("click", async e => {
             e.preventDefault();
@@ -262,7 +262,7 @@ export default class FeedbackActions {
                     f.markBusy();
                     f.data = "";
                 });
-                await this.update({}, true);
+                await this.update(null);
             }
         });
     }
--- a/app/assets/javascripts/feedback/actions.ts
+++ b/app/assets/javascripts/feedback/actions.ts
@@ -262,7 +262,11 @@ export default class FeedbackActions {
                     f.markBusy();
                     f.data = "";
                 });
-                await this.update(null);
+                const values = this.scoreForms.map(f => f.getDataForNested(true));
+                await this.update({
+                    // eslint-disable-next-line camelcase
+                    scores_attributes: values
+                });
             }
         });
     }
--- a/app/assets/javascripts/feedback/actions.ts
+++ b/app/assets/javascripts/feedback/actions.ts
@@ -31,6 +31,7 @@ export default class FeedbackActions {
     private readonly allScoresMaxButton: HTMLButtonElement | null;
     private readonly scoreSumElement: HTMLInputElement | null;
     private readonly completedIcon: HTMLButtonElement | null;
+    private readonly deleteAllButton: HTMLButtonElement | null;
 
     private scoreForms: ScoreForm[];
     // ID's of the score forms that are updating.
@@ -50,7 +51,8 @@ export default class FeedbackActions {
         this.skipCompletedCheckBox = document.getElementById("skip-completed") as HTMLInputElement;
 
         this.scoreSumElement = document.getElementById("score-sum") as HTMLInputElement;
-        this.completedIcon = document.getElementById("completed-button") as HTMLButtonElement;
+        this.completedIcon = document.getElementById("completed-checkmark") as HTMLButtonElement;
+        this.deleteAllButton = document.getElementById("delete-all-button") as HTMLButtonElement;
 
         // Score forms
         this.scoreForms = [];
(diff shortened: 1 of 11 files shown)


app/assets/javascripts/feedback/actions.ts
```diff
@@ -158,16 +160,11 @@ export default class FeedbackActions {
         // Only update the total if we have a total.
         if (this.scoreSumElement) {
             this.scoreSumElement.value = newTotal;
+            this.deleteAllButton.disabled = this.scoreSumElement.value === "-"; // '-' is the placeholder if no score items are filled in yet
```
chvp, line 163: Using string matching for this doesn't feel very clean. Especially using a string from the HTML.

app/assets/javascripts/feedback/actions.ts
```diff
@@ -100,11 +102,11 @@ export default class FeedbackActions {
         this.scoreForms.forEach(s => s.disableInputs());
     }
 
-    update(data: Record<string, unknown>): Promise<void> {
+    update(data: Record<string, unknown>, clearScores: boolean): Promise<void> {
         this.disableInputs();
         return fetch(this.options.currentURL, {
             method: "PATCH",
-            body: JSON.stringify({ feedback: data }),
+            body: JSON.stringify({ feedback: data, clear_scores: clearScores }),
```
chvp, line 109: I think it would be cleaner to make the data an object matching the score item to a `null` value. Not sure if this would work on the rails side, but if it works that would be preferable IMO.

app/controllers/feedbacks_controller.rb
```diff
@@ -64,21 +64,41 @@ def edit
...
+      @feedback = Feedback.find(params[:id])
+      @score_map = @feedback.scores.index_by(&:score_item_id)
 
-    respond_to do |format|
-      if updated
+      # render the view
+      respond_to do |format|
         format.html { redirect_to evaluation_feedback_path(@feedback.evaluation, @feedback) }
-        format.json { render :show, status: :ok, location: @feedback }
+        format.json { render :show, status: :no_content, location: @feedback }
```
chvp, line 81: Rendering `show` with a status `no content` feels weird.

chvp (changes requested):

app/assets/javascripts/feedback/actions.ts
```diff
@@ -258,5 +254,16 @@ export default class FeedbackActions {
...
         });
+        this.deleteAllButton?.addEventListener("click", async e => {
+            e.preventDefault();
+            if (window.confirm(I18n.t("js.score.confirm"))) {
+                this.disableInputs();
+                this.scoreForms.forEach(f => {
+                    f.markBusy();
+                    f.data = "";
+                });
+                await this.update(null);
```
chvp, line 265: This is not what I meant with making things `null`. Just like the all correct and the all wrong button, I would make the actual (individual) score items sent to the backend `null`.

chvp (changes requested):

app/assets/javascripts/feedback/actions.ts
```diff
@@ -158,16 +160,11 @@ export default class FeedbackActions {
...
 
         if (this.completedIcon) {
-            if (completed) {
-                this.completedIcon.classList.add("mdi-check");
-                this.completedIcon.classList.remove("mdi-circle-slice-3");
-            } else {
-                this.completedIcon.classList.add("mdi-circle-slice-3");
-                this.completedIcon.classList.remove("mdi-check");
-            }
+            this.completedIcon.hidden = !completed;
```
bmesuere, line 167: Is there a reason why the partial icon was removed?

app/assets/javascripts/feedback/actions.ts
```diff
@@ -258,5 +254,20 @@ export default class FeedbackActions {
...
+        this.deleteAllButton?.addEventListener("click", async e => {
+            e.preventDefault();
+            if (window.confirm(I18n.t("js.score.confirm"))) {
+                this.disableInputs();
+                this.scoreForms.forEach(f => {
+                    f.markBusy();
+                    f.data = "";
+                });
+                const values = this.scoreForms.map(f => f.getDataForNested(true));
+                await this.update({
```
bmesuere, line 266: I'm probably missing something obvious, but is there a reason to add await here (and thus make it async)?

app/assets/javascripts/feedback/score.ts
```diff
@@ -119,16 +119,18 @@ export default class ScoreForm {
         return this.input.value;
     }
 
-    public getDataForNested(): Record<string, unknown> {
+    public getDataForNested(setNullScore=false): Record<string, unknown> {
+        const newScore = setNullScore? null: this.input.value;
```
bmesuere, line 123:
```suggestion
        const newScore = setNullScore ? null : this.input.value;
```

app/controllers/feedbacks_controller.rb
```diff
@@ -64,21 +64,39 @@ def edit
   end
 
   def update
```
bmesuere, line 66: @chvp or @jorg-vr might not agree, but it seems there is almost no common code in the 2 cases of the if. Doesn't it then make more sense to add a separate action in the controller?

app/assets/javascripts/feedback/actions.ts
BramDevlaminck: Not specifically. I just did something similar to the checkmarks that were there in the sub-scores. Do you prefer the partial icon to be there when the feedback is not yet complete?

I'm just not sure if it adds unnecessary complexity or not. The icon itself does not add a lot of value imo.

app/assets/javascripts/feedback/actions.ts
BramDevlaminck: The `await` is there since `this.update` is an async function (returns the `fetch`)

bmesuere: yes, I know, but adding await won't make a difference, right? There is no code below in the function and the function is inline so nothing is waiting on it.

BramDevlaminck: I think so yeah. But not using the `await` keyword gives a warning. The "solution" to prevent this warning is to just place `.then()`. But as you said, it shouldn't make a difference. (the same should be true for the other buttons above the delete all button).

app/assets/javascripts/feedback/actions.ts
chvp: There is a reason though? The delete button was put in its place?

app/controllers/feedbacks_controller.rb
```diff
@@ -64,21 +64,39 @@ def edit
...
-    attrs = permitted_attributes(@feedback)
-    attrs['scores_attributes'].each { |s| s['last_updated_by_id'] = current_user.id } if attrs['scores_attributes'].present?
+    # the set scores are nil if we want to delete all the currently set scores
+    if params[:feedback][:scores_attributes]&.all? { |item| item[:score].nil? }
```
chvp, line 68: I would hope that this is not required due to the `accepts_nested_attributes_for` in `app/models/feedback.rb`. If it is required, I agree with @bmesuere to move this to a separate action in the controller. (Or maybe check in that case if `accepts_nested_attributes_for` has some other mechanism for deletion?)

chvp (changes requested):

app/assets/javascripts/feedback/actions.ts
bmesuere: Yes, but the checkmark was simply moved to a new position so it seems logical to also move the partial icon to the new place?

app/controllers/feedbacks_controller.rb
BramDevlaminck: What exactly is required? If you mean that the scores have to be nil? It's not really required in general  (but it is for this implementation). It's also not forced by `accepts_nested_attributes_for`. It's just how I identify if the request sent by the client is to delete the scores or if it is to modify the scores.

I debated using the `destroy` function as a separate action, but that did not really seem appropriate since the feedback itself is not getting deleted. Only the scores in the feedback.

app/assets/javascripts/feedback/score.ts
```diff
@@ -119,16 +119,18 @@ export default class ScoreForm {
         return this.input.value;
     }
 
-    public getDataForNested(): Record<string, unknown> {
+    public getDataForNested(setNullScore=false): Record<string, unknown> {
```
jorg-vr, line 122: I do not like this `setNullScore` parameter to the getter method. Just do an extra `map` call after calling this method when needed instead editing this method

app/controllers/feedbacks_controller.rb
jorg-vr: I agree that this should probably be a separate method. 
Restfull I would expect the endpoint to be a DELETE request on feedback/{ID}/scores
(And not a DELETE on feedback/{id} as you discussen in your response to @chvp )

As the functionality is `@feedback.scores.each(&:destroy)`

app/controllers/feedbacks_controller.rb
jorg-vr: If you do keep it as a single method, I would not expect this check. Instead I would expect the method to be able to handle any combination of `nil` and filled out scores (Destroying those which are nil). This way more code would be shared.

But as pointed out above a separate method is probably easier and thus preferable.

jorg-vr (changes requested):

app/assets/javascripts/feedback/actions.ts
BramDevlaminck: I've updated the description with a new screenshot that contains the partial icon when not all the score items are filled in

jorg-vr:
could you fix the failing tests?

jorg-vr (approved):

bmesuere (changes requested):
Looks good 👍 
A last minor request, could you add tooltips (by setting the title attribute) to the partial and complete icon on the total row, and to the saved icon on the other rows?

chvp (approved):

bmesuere (approved):

PR #3921, opened 2022-08-23
Course.js: remove jquery and convert to typescript

### LINKED ISSUE #3923: Refactor course.js to TypeScript and remove jQuery
- [x] Remove all the jquery specific code and replace with vanilla js code
- [x] convert to typescript

BramDevlaminck:
This pull request removes jquery code out of `course.js`

progress on  #3590, closes #3923  .


app/assets/javascripts/course.js
```diff
@@ -219,72 +237,72 @@
...
-            $(this)
-                .closest(".panel")
-                .find(".answer")
-                .html($(this).data("answer"));
+        document.getElementById("new-course").addEventListener("click", function () {
+            choosePanel.classList.add("hidden");
+            formPanel.querySelector(".step-circle").innerHTML = "2";
+            this.closest(".panel")
+                .querySelector(".answer")
+                .innerHTML = this.getAttribute("data-answer");
```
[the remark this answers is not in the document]
BramDevlaminck: Solved by using `.textContent` instead of `.innerHTML`

app/assets/javascripts/course.js
```diff
@@ -219,72 +237,72 @@
...
-                .html($(this).data("answer"));
+        document.getElementById("copy-course").addEventListener("click", function () {
+            choosePanel.classList.remove("hidden");
+            choosePanel.querySelector(".panel-collapse").classList.add("show");
+            choosePanel.querySelector("input[type=\"radio\"]").checked = false;
+            formPanel.classList.add("hidden");
+            formPanel.querySelector(".step-circle").innerHTML = "3";
+            this.closest(".panel")
+                .querySelector(".answer")
+                .innerHTML = this.getAttribute("data-answer");
```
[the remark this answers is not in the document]
BramDevlaminck: Solved by using `.textContent` instead of `.innerHTML`

app/assets/javascripts/course.ts
```diff
@@ -0,0 +1,378 @@
...
+
+function initCourseMembers(): void {
+    function init(): void {
+        initUserTabs();
+        initLabelsEditModal();
+    }
+
+    function initUserTabs(): void {
+        const userTabs = document.getElementById("user-tabs");
+        if (userTabs !== null) {
```
niknetniko, line 19: Maybe this would be clearer by reversing the `if` and returning early, but I'm unsure what others think about this.

Something like this:

```typescript
if (userTabs === null) {
    return;
}
```

app/assets/javascripts/course.ts
```diff
@@ -0,0 +1,378 @@
...
+            const baseUrl = userTabs.getAttribute("data-baseurl");
```
niknetniko, line 20:
```suggestion
            const baseUrl = userTabs.dataset.baseurl;
```
A similar change can be done for other data attributes.

app/assets/javascripts/course.ts
```diff
@@ -0,0 +1,378 @@
...
+
+            // Select tab and load users
+            const selectTab = (tab): HTMLElement => {
+                const kebab = document.getElementById("kebab-menu");
+                const status = tab.getAttribute("data-status");
```
niknetniko, line 25:
```suggestion
                const status = tab.dataset.status;
```

app/assets/javascripts/course.ts
```diff
@@ -0,0 +1,378 @@
...
+                const kebabItems = kebab.querySelectorAll<HTMLElement>("li a.action");
+                let anyShown = false;
+                for (const item of Array.from(kebabItems)) {
+                    const dataType = item.getAttribute("data-type");
```
niknetniko, line 29:
```suggestion
                    const dataType = item.dataset.type;
```

app/assets/javascripts/course.ts
```diff
@@ -0,0 +1,378 @@
...
+        if (userTabs !== null) {
+            const baseUrl = userTabs.getAttribute("data-baseurl");
+
+            // Select tab and load users
+            const selectTab = (tab): HTMLElement => {
+                const kebab = document.getElementById("kebab-menu");
+                const status = tab.getAttribute("data-status");
+                const kebabItems = kebab.querySelectorAll<HTMLElement>("li a.action");
+                let anyShown = false;
+                for (const item of Array.from(kebabItems)) {
```
niknetniko, line 28:
```suggestion
                for (const item of kebabItems) {
```
You should be able to loop over a `NodeList` with a `for...of` directly.

app/assets/javascripts/course.ts
```diff
@@ -0,0 +1,378 @@
...
+                loadUsers(baseUrl, status);
+                document.querySelector("#user-tabs li.active").classList.remove("active");
+                tab.parentNode.classList.add("active");
+            };
+
+            // Switch to clicked tab
+            document.querySelectorAll("#user-tabs li a")
+                .forEach(el => {
+                    el.addEventListener("click", function () {
+                        selectTab(this);
```
niknetniko, line 56: What is the `this` referring to here? Since arrow functions don't have their own `this`. If I am correct, this would refer to the context of the `initUserTabs` function, but I'm not sure this is what we want.

app/assets/javascripts/course.ts
```diff
@@ -0,0 +1,378 @@
...
+        }
+    }
+
+    function initLabelsEditModal(): void {
+        document.getElementById("labelsUploadButton").addEventListener("click", () => {
+            const modal = document.getElementById("labelsUploadModal");
+            const input = document.getElementById("newCsvFileInput") as HTMLInputElement;
+            const formData = new FormData();
+            formData.append("file", input.files[0]);
+            fetch(`/courses/${modal.getAttribute("data-course_id")}/members/upload_labels_csv`, {
```
niknetniko, line 76:
```suggestion
            fetch(`/courses/${modal.dataset.course_id}/members/upload_labels_csv`, {
```

app/assets/javascripts/course.ts
```diff
@@ -0,0 +1,378 @@
...
+    private card: HTMLElement;
+    private url: string;
+    private table_wrapper: HTMLElement;
+    private skeleton: HTMLElement;
+    private loaded: boolean;
+    private loading: boolean;
+    private top: number;
+    private bottom: number;
+
+    getTop(): number {
```
niknetniko, line 108: Perhaps we can use the getter syntax here?

app/assets/javascripts/course.ts
```diff
@@ -0,0 +1,378 @@
...
+
+    constructor(card) {
+        this.id = +card.id.split("series-card-")[1];
+
+        this.reselect(card);
+    }
+
+    reselect(cardSelector): void {
+        this.card = cardSelector;
+        this.url = this.card.getAttribute("data-series-url");
```
niknetniko, line 129:
```suggestion
        this.url = this.card.dataset.seriesUrl;
```

app/assets/javascripts/course.ts
```diff
@@ -0,0 +1,378 @@
...
+        return Array.from(cards, card => new Series(card));
+    }
+
+    constructor(card) {
+        this.id = +card.id.split("series-card-")[1];
+
+        this.reselect(card);
+    }
+
+    reselect(cardSelector): void {
```
niknetniko, line 127: This param is no longer the selector, but the element I think (see the call on line 150)

app/assets/javascripts/course.ts
```diff
@@ -0,0 +1,378 @@
...
+
+function initCourseNew(): void {
+    function init(): void {
+        initPanelLogic();
+        window.dodona.courseFormLoaded = courseFormLoaded;
+        window.dodona.copyCoursesLoaded = copyCoursesLoaded;
+
+        // Bootstrap's automatic collapsing of other elements in the parent breaks
+        // when doing manual shows and hides, so we have to do this.
+        typeCollapseElement.addEventListener("show.bs.collapse", function () {
```
niknetniko, line 256: This can be an arrow function I think.

app/assets/javascripts/course.ts
```diff
@@ -0,0 +1,378 @@
...
+            chooseCollapse.hide();
+            formCollapse.hide();
+        });
+        chooseCollapseElement.addEventListener("show.bs.collapse", function () {
```
niknetniko, line 260: This can be an arrow function I think.

app/assets/javascripts/course.ts
```diff
@@ -0,0 +1,378 @@
...
+            typeCollapse.hide();
+            formCollapse.hide();
+        });
+        formCollapseElement.addEventListener("show.bs.collapse", function () {
```
niknetniko, line 264: This can be an arrow function I think.

app/assets/javascripts/course.ts
```diff
@@ -0,0 +1,378 @@
...
+        });
+    }
+
+    function copyCoursesLoaded(): void {
+        document.querySelectorAll("[data-course_id]").forEach(el => {
+            el.addEventListener("click", function () {
+                this.querySelector("input[type=\"radio\"]").checked = true;
+                this.closest(".panel")
+                    .querySelector(".answer")
+                    .textContent = this.getAttribute("data-answer");
```
niknetniko, line 314:
```suggestion
                    .textContent = this.dataset.answer";
```

app/assets/javascripts/course.ts
```diff
@@ -0,0 +1,378 @@
...
+                fetch(`/courses/new.js?copy_options[base_id]=${this.getAttribute("data-course_id")}`)
+                    .then(req => req.text())
+                    .then(resp => eval(resp));
+            });
+        });
+
+        document.querySelectorAll(".copy-course-row .nested-link").forEach(el => {
+            el.addEventListener("click", function (e) {
```
niknetniko, line 322: This can be an arrow function I think.

app/assets/javascripts/course_membership.js
```diff
@@ -5,7 +5,7 @@ function initCourseMemberLabelsEdit(labels) {
         identify: d => d.id,
         datumTokenizer: d => {
             const result = Bloodhound.tokenizers.whitespace(d.name);
-            $.each(result, (i, val) => {
```
niknetniko, line 8: It doesn't feel like these changes belong on this PR, as there are still jQuery usages in this file (and also since the titles says it's about courses.js)

niknetniko (changes requested):

app/assets/javascripts/course.ts
BramDevlaminck: I thought so too, but the typescript linter thought otherwise.

The solution was to add `DOM.iterable` to the libs in tsconfig. (as suggested in the comment on [this](https://stackoverflow.com/questions/65364900/ts-lint-error-with-document-queryselectorall) stackoverflow post)

app/assets/javascripts/course.ts
BramDevlaminck: ah this is indeed a mistake, this `this` had to refer to the selected tab (in this case `el`)

app/assets/javascripts/course.ts
niknetniko: The problem is that the `target` in our `tsconfig.js` is `es5`. `NodeList` was added in `es6`, so Typescript isn't happy. Since we use Babel I think we could set the target to `es6`, but I'm not completely sure on this.

bmesuere: I think we could. Even without babel, es6 should be supported by the browsers we cover.

BramDevlaminck: > The problem is that the `target` in our `tsconfig.js` is `es5`. `NodeList` was added in `es6`, so Typescript isn't happy. Since we use Babel I think we could set the target to `es6`, but I'm not completely sure on this.

Why is it a problem that our target is `es5`? I don't immediately find anything online about it and even when we use `es6`as target the linter warning stays without adding the `DOM.iterable` lib

niknetniko: I tried https://stackoverflow.com/questions/51723962/typescript-nodelistofelement-is-not-an-array-type-or-a-string-type locally and for me it seemed to work

BramDevlaminck: Hmm, it does not work for me (tried both in VSCode and Webstorm to double-check).

chvp: It also doesn't work for me when setting the target to `es6` and removing `DOM.iterable` from the `lib` list. But I don't see any issue with adding `DOM.iterable` to `lib`?

chvp (approved):

app/assets/javascripts/course.ts
niknetniko: You are right that setting the target itself doesn't work (oops, I modified more locally apparently).

I think the best solution is probably removing `lib` and setting `"target": "es2021"`. Setting the [`target`](https://www.typescriptlang.org/tsconfig#target) will also set a default for [`lib`](https://www.typescriptlang.org/tsconfig#lib), unless you specific it yourself, which is what we did. Since we already include `es2021` libs, we might as well set the target to `es2021` and use the default value.

niknetniko (approved):

bmesuere (approved):
A minor nitpick: I think template strings (between backticks) would look cleaner for the selectors where you now have to escape the double quotes, e.g., in `type=\"radio\"`.

BramDevlaminck:
> A minor nitpick: I think template strings (between backticks) would look cleaner for the selectors where you now have to escape the double quotes, e.g., in `type=\"radio\"`.

I've applied this and also changed the `.eslintrc` to allow this. With the previous `.eslintrc` eslint would not allow this since it forced you to use double quotes on every string

PR #3926, opened 2022-08-25
Improve buttons in choose submission screen for evaluation

### LINKED ISSUE #3795: Reduce size of buttons on change submission table
The table that is used to change the selected submission of a student in an evaluation has buttons to the right of the table whose size could be reduce a bit.

image

BramDevlaminck:
This pull request changes the buttons that are used to select the submission and adds an indication to the currently selected submission.

image
image

Closes #3795  .

### DIFF
--- a/app/assets/stylesheets/components/table.css.scss
+++ b/app/assets/stylesheets/components/table.css.scss
@@ -56,7 +56,7 @@
 
 .table-resource .actions {
   text-align: center;
-  width: 1px;
+  width: 1px; // by default the size of the element in the col is used as width if width is smaller than that element. This width is just here to force everything to the right
 }
 
 tr.gu-mirror {
--- a/app/assets/stylesheets/components/table.css.scss
+++ b/app/assets/stylesheets/components/table.css.scss
@@ -59,9 +59,11 @@
 
   &.submissions-table {
     text-align: center;
-    width: 1px; // by default the size of the element in the col is used as width if width is smaller than that element. This width is just here to force everything to the right
-  }
 
+    // by default the size of the element in the col is used as width if width is smaller than that element.
+    // This width is just here to force everything to the right
+    width: 1px;
+  }
 }
 
 tr.gu-mirror {
--- a/app/assets/stylesheets/components/table.css.scss
+++ b/app/assets/stylesheets/components/table.css.scss
@@ -55,8 +55,13 @@
 }
 
 .table-resource .actions {
-  text-align: center;
-  width: 1px; // by default the size of the element in the col is used as width if width is smaller than that element. This width is just here to force everything to the right
+  text-align: right;
+
+  &.submissions-table {
+    text-align: center;
+    width: 1px; // by default the size of the element in the col is used as width if width is smaller than that element. This width is just here to force everything to the right
+  }
+
 }
 
 tr.gu-mirror {
--- a/app/assets/stylesheets/components/table.css.scss
+++ b/app/assets/stylesheets/components/table.css.scss
@@ -55,7 +55,8 @@
 }
 
 .table-resource .actions {
-  text-align: right;
+  text-align: center;
+  width: 1px;
 }
 
 tr.gu-mirror {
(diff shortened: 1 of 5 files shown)


chvp (commented):
Maybe it's just my imagination but the "Selected" in the English screenshot looks very weirdly aligned. LGTM otherwise.

config/locales/views/submissions/nl.yml
```diff
@@ -14,6 +14,8 @@ nl:
       result: Resultaat
       summary: Samenvatting
       annotated: Deze oplossing heeft opmerkingen.
+      select: Selecteer
```
niknetniko, line 17:
```suggestion
      select: Selecteren
```

niknetniko (approved):

BramDevlaminck:
> Maybe it's just my imagination but the "Selected" in the English screenshot looks very weirdly aligned. LGTM otherwise.

It indeed seemed a bit off-centre. To make it fit, I've shifted the button a bit. Is this new version better?
image
image

app/views/feedbacks/_submissions_table.html.erb
```diff
@@ -28,10 +28,17 @@
             <%= submission.summary %>
           </td>
           <td class="actions">
-            <% unless @feedback.submission&.id == submission.id %>
-              <%= link_to feedback_path(@feedback, feedback: { submission_id: submission.id }), class: "btn btn-icon btn-icon-filled bg-primary", title: t(".update-submission"), confirm: t('.confirm'), method: :patch do %>
-                <i class="mdi mdi-check"></i>
-              <% end %>
+            <% if @feedback.submission&.id == submission.id %>
+              <button class="btn btn-text outline-btn-alignment" disabled>
```
jorg-vr, line 32: You should not be using a btn-text to simulate a disabled btn-outline

Disabled btn-outline has its own spec:
image

See: https://m3.material.io/components/buttons/specs

If this style is not yet implemented for outlined buttons it should be, instead of misusing btn-text

jorg-vr (changes requested):

app/views/feedbacks/_submissions_table.html.erb
BramDevlaminck: This would look something like this:
image

It just looks weird to me with the 1 button that is bigger. That's why I just used btn-text (since that has no outline).
I just used the text since it seemed a good indicator to show what was selected and still had the same style as the buttons.

What do you think of this?

jorg-vr: Can't this be solved by some kind of centered alignment of the buttons  (Instead of pulled to the right as they are right now)

BramDevlaminck: Centring the elements won't work, since the buttons start to "float" somewhere in the row.

image

There also does not seem to be an obvious way to right-align the buttons and center them at the same time.

I could try some hacky way by adding an extra class to the active buttons that adds more padding to the right, but that's not really a solution to this problem. It's only hiding the real problem a bit.

jorg-vr: just give the actions column a predefined width. eg:
```css
.table-resource .actions {
  text-align: center;
  width: 150px;
}
```

This is much less hacky than custom paddings

image

BramDevlaminck: I've updated the description of the pr to contain the new screenshots

app/assets/stylesheets/components/table.css.scss
```diff
@@ -55,7 +55,8 @@
 }
 
 .table-resource .actions {
-  text-align: right;
+  text-align: center;
+  width: 1px; // by default the size of the element in the col is used as width if width is smaller than that element. This width is just here to force everything to the right
```
jorg-vr, line 59: As you made this edit globally, did you check all other tables for issues created by this change?

If you encounter issues, just make it more specific with the `.submissions-table` selector

jorg-vr (approved):

app/assets/stylesheets/components/table.css.scss
BramDevlaminck: i've checked this on multiple places and it all seemed to work. But I'll make it more specific just to be sure since it's possible there is a table somewhere where this will break. I should've done that in the first place

PR #3929, opened 2022-08-25
No more teal in token colors

bmesuere:
This pull request swaps out teal for red in the search tokens. To me, teal doesn't match with the other colors we used here, red blends in better. 

Other opinions are of course welcome!

image

image


chvp (approved):
I agree that it looks nicer. :+1:

PR #3961, opened 2022-09-01
Fix empty series in course crash

BramDevlaminck:
This pull request solves the crash when a course has an empty series

the fix was changing:
```ts
this.skeleton = this.table_wrapper.querySelector(SKELETON_TABLE_SELECTOR);
```
to
```ts
this.skeleton = this.table_wrapper?.querySelector(SKELETON_TABLE_SELECTOR);
```
since `table_wrapper` was null when a series is empty


app/assets/javascripts/course.ts
```diff
@@ -129,7 +129,7 @@ class Series {
         this.card = card;
         this.url = this.card.dataset.seriesUrl;
         this.table_wrapper = this.card.querySelector(TABLE_WRAPPER_SELECTOR);
-        this.skeleton = this.table_wrapper.querySelector(SKELETON_TABLE_SELECTOR);
```
niknetniko: While this is a correct fix, the types of the instance variables should also reflect that `table_wrapper` and `skeleton` can be `null`.

<hr>

However, an even better improvement is probably to remove `card`,  `table_wrapper` and `skeleton` from the instance variables and just use them locally in the function, like so:

```javascript
reselect(card: HTMLElement): void {
    this.url = card.dataset.seriesUrl;
    const tableWrapper = card.querySelector(TABLE_WRAPPER_SELECTOR);
    const skeleton = tableWrapper?.querySelector(SKELETON_TABLE_SELECTOR);
    this.loaded = skeleton === null;
    this.loading = false;
    this._top = card.getBoundingClientRect().top + window.scrollY;
    this._bottom = this.top + card.getBoundingClientRect().height;
}
```
(Not necessarily needed for this PR, if we want to keep this one smaller)

niknetniko (changes requested):

app/assets/javascripts/course.ts
BramDevlaminck: I'll change them to local values, since this indeed is better and the rewrite is minimal

chvp (changes requested):
As mentioned during the coffee break, I think that while this fix works, it isn't *really* correct. Because with this fix, the series will be marked as not loaded, even though the content of the series is all there.

chvp (approved):

niknetniko (approved):

jorg-vr (approved):

PR #3963, opened 2022-09-02
favorite_course_buttons.js: remove jquery and use typescript

BramDevlaminck:
This pull request removes jquery code from `favorite_course_button.js` and migrates js to ts.

progress on  #3590 .

### DIFF
--- a/app/assets/javascripts/favorite_course_buttons.ts
+++ b/app/assets/javascripts/favorite_course_buttons.ts
@@ -1,12 +1,12 @@
 import { Toast } from "./toast";
 
-function initFavoriteButtons() {
-    function init() {
+function initFavoriteButtons(): void {
+    function init(): void {
         document.querySelectorAll(".favorite-button")
             .forEach(btn => btn.addEventListener("click", toggleFavorite));
     }
 
-    function toggleFavorite() {
+    function toggleFavorite(): void {
         const element = this;
         if (element.classList.contains("favorited")) {
             unfavoriteCourse(element);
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/favorite_course_buttons.ts
```diff
@@ -0,0 +1,99 @@
...
+        if (this.classList.contains("favorited")) {
+            unfavoriteCourse(this);
+        } else {
+            favoriteCourse(this);
+        }
+    }
+
+    function favoriteCourse(element: HTMLElement): void {
+        const courseId = element.dataset.course_id;
+        fetch(`/courses/${courseId}/favorite.js`, {
```
niknetniko, line 19: Here it is also worth using the `fetch` from `utils.js`.

app/assets/javascripts/favorite_course_buttons.ts
```diff
@@ -0,0 +1,99 @@
...
+            "method": "POST",
+            "headers": {
+                "x-csrf-token": document.querySelector(`meta[name="csrf-token"]`).getAttribute("content"),
+            },
+        }).then(response => {
+            if (response.ok) {
```
niknetniko, line 25: We might want to create a wrapper that automatically fails the promise if the return status is not `ok`, since we need this in a bunch of places here and in other PR's (this is probably not something for this PR, but just a note for another time)

niknetniko (changes requested):

app/assets/javascripts/favorite_course_buttons.ts
BramDevlaminck: That would indeed be better since it's ugly that the `catch` gets the error when there is a network issue and the `else` in the `then` for server issues,... It would also make the rewrite easier since it would follow the `.done` and `.fail` from ajax

niknetniko: Yep

jorg-vr (approved):

niknetniko (approved):

PR #4023, opened 2022-09-20
Replace sign_in.js by webcomponent

jorg-vr:
This pull request replaces sign_in.j by a webcomponent. This is part of our js modernization.

I also added tabcompletion and higlighting matches to the datalist component. Which has a wider impact.

Old
image
Screenshot from 2022-09-20 14-19-51

New
image
image
image

Part of #3590


chvp (approved):

app/assets/javascripts/util.js
```diff
@@ -173,6 +173,13 @@ const ready = new Promise(resolve => {
     }
 });
 
+function htmlEncode(str) {
+    return String(str)
+        .replace(/&/g, "&amp;")
+        .replace(/</g, "&lt;")
+        .replace(/>/g, "&gt;")
+        .replace(/"/g, "&quot;");
+}
```
bmesuere, line 182: I'm a bit hesitant about adding this function because it's incomplete. (and in combination with the unsafeHTML function where it's called). Is there an alternative?

jorg-vr: I don't see a way without the unsafe html that also marks the search results. And as the institution names are user input we need some kind of html encoding.

This method is an often suggested solution:
- https://stackoverflow.com/questions/6234773/can-i-escape-html-special-chars-in-javascript
- https://stackoverflow.com/questions/24816/escaping-html-strings-with-jquery

Some indeed change more keys. I will add those. They are more relevant when the string is used within html tags. But for a general util method it should cover all cases.

There is weirdly no standard javascript method for this. JQuery and lodash provide solutions, but we dont want to include those. 


This might be a better alternative:
`function escapeHtml(html){
  var text = document.createTextNode(html);
  var p = document.createElement('p');
  p.appendChild(text);
  return p.innerHTML;
}`

bmesuere: Soon™️ https://developer.mozilla.org/en-US/docs/Web/API/HTML_Sanitizer_API

bmesuere (approved):

PR #4029, opened 2022-09-26
Fix ufora LTI authentication without revert

jorg-vr:
This pull request reintroduces #3476 after its removal in #4028 

It provides a better fix by skipping the token validation for LTI sign in, since no such token is present after the redirect from ufora.

This pr has been tested on the uforatest server.


chvp (approved):

chvp:
The revert of the revert doesn't show for me in the files changed tab. Are you sure it is included?

jorg-vr:
@chvp  I updated the branch, the files are shown now

niknetniko (approved):

bmesuere (approved):

Issue #2127, opened 2020-07-31
Migrate to active record session store


bmesuere:
The LTI implementation uses redirect URLs with (large) JWT tokens in it. These URLs are stored in the session while signing in. We currently use the default rails cookie store for session data which has a limit of 4KB and due to the JWT tokens, this is not enough.

In theory, the max cookie size can be increased, but it is better not to use cookies larger than 4KB in size. An alternative is to use the database for storing the session data. This used to be part of rails, but was moved to a separate gem: https://github.com/rails/activerecord-session_store

PR #4041, opened 2022-09-29
Store session information in the database instead of a cookie

jorg-vr:
This pull request fixes the cookie overflow bug that was caused by #4002

To fix this we store the session info in the database instead of a cookie as was already suggested in #2127

- [x]  Script to trim session db on ansible: https://github.com/dodona-edu/dodona-ansible/pull/151
> To avoid your sessions table expanding without limit as it will store expired and potentially sensitive session data, it is strongly recommended in production environments to schedule the `db:sessions:trim` rake task to run daily. Running `bin/rake db:sessions:trim` will delete all sessions that have not been updated in the last 30 days.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -140,6 +140,9 @@ gem 'ddtrace', '~> 1.4.1'
 # Make sure filesystem changes only happen at the end of a transaction
 gem 'after_commit_everywhere', '~> 1.2.2'
 
+# Store session data in the database
+gem 'activerecord-session_store', '~> 2.0.0'
+
 group :development, :test do
   # Use mocha for stubbing and mocking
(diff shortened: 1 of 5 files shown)


chvp (approved):
Running the rake task periodically is something for the dodona-ansible repo, IMO. A cron entry or a systemd timer can be added there.

bmesuere:
Since the reason to add this is an edge case for which the use of the session is questionable and that this change has a potential performance impact (DB read/write on every request) we will not proceed with this.

Instead, we will store the temporary data in memcached with a 5 minutes auto expiry using a random key. We will then store the random key in the session instead.

PR #4048, opened 2022-10-03
Fix annotations not shown bug for submissions outside a course

jorg-vr:
This pull request fixes a bug where annotations were not shown in the code tab for submissions with course_id nil.

### DIFF
--- a/app/helpers/renderers/feedback_code_renderer.rb
+++ b/app/helpers/renderers/feedback_code_renderer.rb
@@ -92,7 +92,7 @@ class FeedbackCodeRenderer
     @builder.script(type: 'application/javascript') do
       @builder << <<~HEREDOC
         window.MathJax.startup.promise.then(() => {
-          window.dodona.codeListing = new window.dodona.codeListingClass(#{submission.id}, #{submission.course_id}, #{submission.exercise_id}, #{user.id}, #{@code.to_json}, #{@code.lines.length}, #{user_is_student});
+          window.dodona.codeListing = new window.dodona.codeListingClass(#{submission.id}, #{submission.course_id || 'undefined'}, #{submission.exercise_id}, #{user.id}, #{@code.to_json}, #{@code.lines.length}, #{user_is_student});
           window.dodona.codeListing.addMachineAnnotations(#{messages.to_json});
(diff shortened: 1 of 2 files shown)


app/helpers/renderers/feedback_code_renderer.rb
```diff
@@ -92,7 +92,7 @@ def add_messages(submission, messages, user)
     @builder.script(type: 'application/javascript') do
       @builder << <<~HEREDOC
         window.MathJax.startup.promise.then(() => {
-          window.dodona.codeListing = new window.dodona.codeListingClass(#{submission.id}, #{submission.course_id}, #{submission.exercise_id}, #{user.id}, #{@code.to_json}, #{@code.lines.length}, #{user_is_student});
+          window.dodona.codeListing = new window.dodona.codeListingClass(#{submission.id}, #{submission.course_id || 'undefined'}, #{submission.exercise_id}, #{user.id}, #{@code.to_json}, #{@code.lines.length}, #{user_is_student});
```
chvp, line 95: Maybe use `submission.course_id.to_json`?

chvp (approved):

app/helpers/renderers/feedback_code_renderer.rb
```diff
@@ -92,7 +92,7 @@ def add_messages(submission, messages, user)
     @builder.script(type: 'application/javascript') do
       @builder << <<~HEREDOC
         window.MathJax.startup.promise.then(() => {
-          window.dodona.codeListing = new window.dodona.codeListingClass(#{submission.id}, #{submission.course_id}, #{submission.exercise_id}, #{user.id}, #{@code.to_json}, #{@code.lines.length}, #{user_is_student});
+          window.dodona.codeListing = new window.dodona.codeListingClass(#{submission.id}, #{submission.course_id.to_json}, #{submission.exercise_id}, #{user.id}, #{@code.to_json}, #{@code.lines.length}, #{user_is_student});
```
niknetniko, line 95: We should probably also update the types in `code_listing.ts` that the course id can now be `null`  instead of `undefined`  (although looking at that code, the types don't indicate it can be `undefined` at all, so perhaps this should be done in another PR).

PR #4061, opened 2022-10-04
Correctly set scores and completed states when a new user is added to an evaluation

### LINKED ISSUE #3983: Users added to evaluation after initial creation don't get 0 on score items with no submissions
The feedback is marked as completed, leaving it in an even more confusing state.

jorg-vr:
When a user is added to an evaluation after score items have been added, the scores for blank submissions should also be set to zero.

Before this pr this only happened when score items were added after the users.

- [x] tests are added

Closes #3983

### DIFF
--- a/app/models/feedback.rb
+++ b/app/models/feedback.rb
@@ -22,6 +22,9 @@ class Feedback < ApplicationRecord
   has_many :scores, dependent: :destroy
   has_many :score_items, through: :evaluation_exercise
 
+  # Should know the current user to store who last updated the score
+  attr_accessor :current_user
+
   delegate :user, to: :evaluation_user
   delegate :exercise, to: :evaluation_exercise
 
@@ -29,6 +32,7 @@ class Feedback < ApplicationRecord
   before_save :reset_feedback_after_submission_update
   before_create :generate_id
   before_create :determine_submission
+  after_create :uncomplete_or_set_blank_to_zero
   before_destroy :destroy_related_annotations
 
   validate :submission_user_exercise_correct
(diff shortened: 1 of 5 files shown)


chvp (approved):
Maybe add soms tests?

app/models/feedback.rb
```diff
@@ -119,4 +120,14 @@ def submission_user_exercise_correct
     errors.add(:submission, 'exercise should be the same as in the evaluation') if submission.present? && submission.exercise_id != exercise.id
     errors.add(:submission, 'course should be the same as in the evaluation') if submission.present? && submission.course_id != evaluation.series.course_id
   end
+
+  def uncomplete_or_set_blank_to_zero
```
niknetniko, line 124: In which scenario should the feedback be uncompleted? I'm just wondering when it happens that we create a new feedback that is already completed.

test/controllers/evaluations_controller_test.rb
```diff
@@ -617,4 +617,47 @@ def expected_score_string(*args)
     assert_no_match expected_score_string(s2), response.body
     assert_no_match expected_score_string(feedback.score, feedback.maximum_score), response.body
   end
+
+  test 'Scores are set if a user is added' do
```
niknetniko, line 621: This test feels more like it belongs in the model tests.

app/models/feedback.rb
```diff
@@ -29,6 +29,7 @@ class Feedback < ApplicationRecord
   before_save :reset_feedback_after_submission_update
   before_create :generate_id
   before_create :determine_submission
+  after_create :uncomplete_or_set_blank_to_zero
```
niknetniko, line 32: Thinking about it, should we also set blank submissions to zero if the submission changes? (The incomplete state is normally already unset by deleting the scores)

We currently don't, and this is a very unlikely scenario, so perhaps not needed.

niknetniko (changes requested):

app/models/feedback.rb
jorg-vr: complete appears to be the default. So a blank submission is completed automatically

app/models/feedback.rb
jorg-vr: It is impossible to select no submission when changing a feedback. so submission.blank? will never be true after an update

app/models/feedback.rb
niknetniko: Ah some code must set it then, I was assumed it was uncompleted by default, since

```
#  completed              :boolean          default(FALSE), not null
```

but I haven't checked all callbacks

jorg-vr: Tests work without the uncomplete, so I'll remove it. I am unsure where blanks where set to complete, but it did happen automatically

niknetniko (approved):

bmesuere (approved):

PR #4069, opened 2022-10-06
Store auth hash in cache instead of session to fix cookie overflow

jorg-vr:
This pull request fixes a cookie overflow bug that was cause by the auth hash being stored in the session. (#4002)
To fix this we opted to store the auth hash in mem_cache, with an expiry of 5 minutes. The session now contains a unique key to that hash object.

Replaces  #4041

### DIFF
--- a/test/controllers/auth/omniauth_callbacks_controller_test.rb
+++ b/test/controllers/auth/omniauth_callbacks_controller_test.rb
@@ -3,15 +3,13 @@ require 'test_helper'
 class OmniauthCallbacksControllerTest < ActionDispatch::IntegrationTest
   include ActiveJob::TestHelper
 
-  def setup
-    super
+  setup do
     # Caching is used in some login flows and should be active for the tests
     ActionController::Base.perform_caching = true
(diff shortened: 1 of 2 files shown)


bmesuere (approved):

test/controllers/auth/omniauth_callbacks_controller_test.rb
```diff
@@ -3,6 +3,17 @@
 class OmniauthCallbacksControllerTest < ActionDispatch::IntegrationTest
   include ActiveJob::TestHelper
 
+  setup do
```
niknetniko, line 6: Is there a reason to enable caching for all tests, instead of using the `with_cache` helper (from `testhelpers/cache_helper.rb`) to enable caching in the tests where it is necessary?

jorg-vr: 1. I didn't know this helper already existed. Good to know now :)
2. All tests in this file are on the login flow. Caching is required in some of the login flows but not all. I thought it would be better if the tests were agnostic of the fact that their specific flow uses caching or not.

niknetniko (approved):

Issue #3299, opened 2022-01-14
Enable additional host names


bmesuere:
This issue tracks the steps that need to be taken to switch from dodona.ugent.be to dodona.be. This will be a multi step process.

## Step 1: Enable naos.dodona.be
**Code changes**
- [x] Disable the redirect from naos.dodona.be to naos.ugent.be
- [x] Enable additional host for naos in rails #3292 
- [x] Set a relative asset host to host the assets from the same domain as the main request #3293
- [x] Provide different SAML metadata depending on the hostname of the request #3293
- [x] Use correct host name for OIDC requests #3293
- [x] Use correct host name for LTI requests #3293
- [x] Enable additional host for dodona in rails #3293
- [x] Enable /media requests on alternative host names #3293

**Things to check**
- [x] UJS seems to use the default host for remote links
- [x] Check LTI (ufora and ilearn)
- [x] Check UGent sign in
- [x] Check O365 sign in
- [x] Check GSuite sign in
- [x] Check Smartschool sign in
- [x] Check Elixir sign in
- [x] Check Vlaanderen sign in  
- [x] Check Surf sign in
- [x] Line 42 of the routes file

**Administrative changes**
- [x] Enable Cloudflare proxy for naos
- [x] Additional redirect URLs were added to Google sign in
- [x] Additional redirect URLs were added to azure
- [x] Register additional SAML service providers in Belnet for the extra hostnames

## Step 2: Enable dodona.be
**Code changes**
- [x] Disable the redirect from dodona.be to dodona.ugent.be

**Things to check**
- [x] UJS seems to use the default host for remote links
- [ ] Check LTI  (ufora and ilearn)
- [x] Check UGent sign in
- [x] Check O365 sign in
- [x] Check GSuite sign in
- [x] Check Smartschool sign in
- [x] Check Elixir sign in
- [ ] Check Vlaanderen sign in  
- [ ]  Check Surf sign in
- [ ] Line 42 of the routes file
- [ ] Check if github webhooks are still working

**Administrative changes**
- [ ] Contact the IDPs we use to add the new hostnames/issuers to the whitelist (can be done at a later time)

## Step 3: Redirect all traffic to the new hostname
**Code changes**
- [x] Redirect dodona.ugent.be to dodona.be

## Step 4: Cleanup
**Code changes**
- [x] Re enable the asset host: https://github.com/dodona-edu/dodona/pull/3293#discussion_r797404955
[...]

PR #4070, opened 2022-10-06
Fix wrong hostname used when using rails `object_url` helper

jorg-vr:
This pull request replaces the default host by the current host for the rails `object_url` helper methods. Otherwise multiple links are broken when multiple hostnames are in use.

Examples of broken links: handing in a submission or deleting a course.

part of #3299

### DIFF
--- a/app/views/layouts/frame.html.erb
+++ b/app/views/layouts/frame.html.erb
@@ -8,7 +8,7 @@
   <meta http-equiv="X-UA-Compatible" content="IE=edge">
   <meta name="viewport" content="width=device-width, initial-scale=1">
   <%= yield :meta_tags %>
-  <meta property="og:image" content="<%= root_url(locale: nil) %>icon.png">
+  <meta property="og:image" content="<%= root_path(locale: nil) %>icon.png">
   <link href="https://fonts.googleapis.com/css2?family=Roboto:ital,wght@0,300;0,400;0,500;0,700;1,400&display=swap" rel="stylesheet">
   <link href="https://cdn.jsdelivr.net/npm/@mdi/font@5.x/css/materialdesignicons.min.css" rel="stylesheet" type='text/css'>
   <% if params[:dark].nil? %>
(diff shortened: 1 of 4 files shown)


chvp:
This fix is kinda weird to me? Shouldn't the default host just be edited to the new one?

bmesuere:
@chvp No, we want both hostnames to be working during the transition period. We therefore have to use the current host.

chvp:
Ah, I missed that bit of information.

jorg-vr:
An update on the sandbox domain.

It was introduced by @rien in https://github.com/dodona-edu/dodona/pull/1234
This pr also introduced the setting of the default host for urls. @rien if you still remember why this was required, let me know.

The only place the sandbox domain is used is to render descriptions. And the only real use of an object_url I found was the default meta path for an og:image that links to the dodona icon (And this path works for both the sandbox and the default host)

The urls in the text description are also absolutised for footnotes and imags, but these already explicitly used the `@request.original_url`, thus the sandbox domain url. 

I tested it on descriptions with multiple links and images and everything still seems to work: https://naos.ugent.be/nl/courses/62/activities/910319224/#

app/views/layouts/frame.html.erb
```diff
@@ -8,7 +8,7 @@
   <meta http-equiv="X-UA-Compatible" content="IE=edge">
   <meta name="viewport" content="width=device-width, initial-scale=1">
   <%= yield :meta_tags %>
-  <meta property="og:image" content="<%= root_url(locale: nil) %>icon.png">
+  <meta property="og:image" content="<%= root_path(locale: nil) %>icon.png">
```
chvp, line 11: Did you test that this change works? Because https://ogp.me/ mentions that it should be an URL to an image.

jorg-vr: I'll leave it as is in that case

chvp (approved):

rien:
@jorg-vr I'm not 100% sure anymore, but I can think of some reasons:
- setting the default URL was required for links back to Dodona
- there are some situations where there current URL is not available (in mailers or workers, anything running "delayed")
- for the tests

PR #4080, opened 2022-10-10
punchcard.js: convert to typescript

freyavs:
This pull request converts `punchcard.js` to `punchcard.ts`

It also formats the x-axis of the punchcard chart as hours, like this:

It is progress on #3590


jorg-vr (approved):

app/assets/javascripts/punchcard.ts
```diff
@@ -50,18 +51,22 @@ function initPunchcard(url) {
     const xAxis = d3.axisBottom(x)
         .ticks(24)
         .tickSize(0)
-        .tickFormat((d, i) => labelsX[i])
+        .tickFormat((hour: d3.NumberValue) => {
+            const now = new Date();
+            now.setHours(hour.valueOf(), 0, 0, 0);
+            return d3.timeFormat("%H:%M")(now);
+        })
```
bmesuere, line 58: Is there a reason for this rather complex function?

niknetniko (changes requested):

app/assets/javascripts/punchcard.ts
freyavs: While going over the x values, it sets them to hours on the x-as. 

I could also just append `":00"` here (to the original x-axis elements 0-23). Or set x as `scaleTime ` instead of `scaleLinear`, but the hours in the data are strings 0-23 so that brings other (small) complications in another part of the code. 

I'm not sure which is the cleanest, or if you have another suggestion?

bmesuere: maybe something like this?
```javascript
`${hour < 10 ? '0' : ''}${hour}:00`
```

chvp (approved):

app/assets/javascripts/punchcard.ts
freyavs: Alright, I'll apply this

niknetniko (approved):

PR #4083, opened 2022-10-17
course_membership.js: replace with webcomponent

freyavs:
This pull request removes course_membership.js and replaces the functionality with web components (using the already existing `d-datalist-input`), with the goal of removing jquery and converting to typescript.

The view has slightly changed from this:

To this:

It is progress on https://github.com/dodona-edu/dodona/issues/3590


app/assets/javascripts/components/course_labels_search_bar.ts
```diff
@@ -0,0 +1,111 @@
+import { customElement, property } from "lit/decorators.js";
+import { html, TemplateResult } from "lit";
+import { ShadowlessLitElement } from "components/shadowless_lit_element";
+import { Option } from "components/datalist_input";
+import { ready } from "util.js";
+import "components/datalist_input";
+/**
+ * This component represents a list of d-label-token
```
jorg-vr, line 8: This comment does not look correct

app/assets/javascripts/components/course_labels_search_bar.ts
```diff
@@ -0,0 +1,111 @@
...
+ * @prop {string[]} Labels
+ */
+@customElement("d-course-label-tokens")
+export class CourseLabelTokens extends ShadowlessLitElement {
+    @property({ type: Array })
+    labels: string[];
+
+    processClick(e: Event, label: string): void {
+        this.labels.splice(this.labels.indexOf(label), 1);
+        this.requestUpdate();
```
jorg-vr, line 21: Could you double check if this is required? I would expect the request update to be triggered by the change to labels in the line above. (But maybe lit has problems with auto detecting changes in lists)

app/assets/javascripts/components/course_labels_search_bar.ts
```diff
@@ -0,0 +1,111 @@
...
+            <input type="hidden" name="course_membership[course_labels]" .value="${this.labels.join(",")}">
+        `;
+    }
+}
+
+/**
+ * @element d-course-labels-search-bar
+ *
+ * @prop {{id: number, name: string}[]} Labels
+ * @prop {string[]} SelectedLabels
```
jorg-vr, line 46: Please describe the function of the element and all properties in the comments

app/assets/javascripts/components/datalist_input.ts
```diff
@@ -121,6 +119,9 @@ export class DatalistInput extends watchMixin(ShadowlessLitElement) {
             this.value = this.filtered_options[0].value;
             this.filter = this.filtered_options[0].label;
         }
+        if (e.key === "Enter") {
+            e.preventDefault();
```
jorg-vr, line 123: Maybe add a comment here why this is done

jorg-vr (changes requested):
This looks good :) 

I left some minor remarks but these are mostly about comments

One visual issue I noticed, could you add a bit of margin to the top or bottom of the labels? they don't look great when there is more than one row
image

app/assets/javascripts/components/course_labels_search_bar.ts
freyavs: Yes so without `requestUpdate` this doesn't rerender. I tried fixing it by copying the list, doing the changes on that list, and then assigning this new list to the `this.labels` (it shoud only autodetects when you assign a new value instead of changing the current list). 

This works, but then when I start to type again, the list gets "overwritten" by the searchbar component as the deletion hasn't been registered there (so when deleting a label it just appears again when you start typing to add another label). And something weird.. when I call `requestUpdate` after the fix of assigning a new list instead of just changing the existing one, it still doesn't change the list in the searchbar component, which causes the same problem. 

Should I keep it like this then?

jorg-vr: yes keep it as is :)

jorg-vr (approved):

bmesuere (commented):
The code looks good, but I have a minor UX remark. If I'm correct, you can type something that is not present in the list and add it as well, but the UI isn't super clear on that because of the double focus on "search" both in the placeholder text and the info below.

Maybe searching doesn't even have to be mentioned because it's not needed?

niknetniko (approved):

freyavs:
> The code looks good, but I have a minor UX remark. If I'm correct, you can type something that is not present in the list and add it as well, but the UI isn't super clear on that because of the double focus on "search" both in the placeholder text and the info below.
> 
> Maybe searching doesn't even have to be mentioned because it's not needed?

I agree. Would this be better, or still not clear enough?
info:  "Add an existing userlabel or create a new one"
placeholder: just something like "start typing..."

bmesuere:
> > The code looks good, but I have a minor UX remark. If I'm correct, you can type something that is not present in the list and add it as well, but the UI isn't super clear on that because of the double focus on "search" both in the placeholder text and the info below.
> > Maybe searching doesn't even have to be mentioned because it's not needed?
> 
> I agree. Would this be better, or still not clear enough? info: "Add an existing userlabel or create a new one" placeholder: just something like "start typing..."

I agree with the info text, but would just use "Label" as placeholder text. This best matches the material design spec.

bmesuere (approved):

PR #4093, opened 2022-10-20
Create github action for deploying the database backup to naos

### LINKED ISSUE #2866: Automate DB backup restore to naos

jorg-vr:
This pull request adds a github action to deploy the latest db backup from dodona on our staging environment naos.

Related pr on ansible: https://github.com/dodona-edu/dodona-ansible/pull/152

Related issue #2866 .

### DIFF
--- a/.github/workflows/deploy-db-backup-to-naos.yml
+++ b/.github/workflows/deploy-db-backup-to-naos.yml
@@ -4,7 +4,7 @@ on:
   workflow_dispatch:
 
 concurrency:
-  group: deploy-db-backup-to-naos
+  group: deploy-naos
   cancel-in-progress: true
 
 jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy-db-backup-to-naos.yml
```diff
@@ -0,0 +1,24 @@
+name: Deploy database backup to Naos
+
+on:
+  workflow_dispatch:
+
+concurrency:
+  group: deploy-db-backup-to-naos
```
chvp, line 7: Maybe this should use the same group as the naos deploy? I can't imagine we want to deploy a branch and the DB backup at the same time.

chvp (approved):

bmesuere (approved):
Not sure if it is easily doable, but I would generate new names for non-zeus users  when importing.

PR #4113, opened 2022-10-28
Enhance/match insititutions

jorg-vr:
This pull request

- [ ] Tests were added

Closes # .

### DIFF
--- a/app/controllers/institutions_controller.rb
+++ b/app/controllers/institutions_controller.rb
@@ -5,9 +5,18 @@ class InstitutionsController < ApplicationController
     scope.by_name(value)
   end
 
+  has_scope :order_by, using: %i[column direction], type: :hash do |_controller, scope, value|
+    column, direction = value
+    if %w[ASC DESC].include?(direction) && %w[name short_name users courses most_similar].include?(column)
+      scope.send "order_by_#{column}", direction
+    else
+      scope
+    end
+  end
+
   def index
     authorize Institution
-    @institutions = apply_scopes(policy_scope(Institution)).all.order(generated_name: :desc, name: :asc)
+    @institutions = apply_scopes(policy_scope(Institution)).all
                                                            .includes(:courses, :users, :providers)
                                                            .paginate(page: parse_pagination_param(params[:page]))
     @title = I18n.t('institutions.index.title')
--- a/app/models/institution.rb
+++ b/app/models/institution.rb
@@ -73,7 +73,24 @@ class Institution < ApplicationRecord
           matrix[j.id][i.id] = matrix[i.id][j.id]
         end
       end
+      matrix
     end
+  end
+
+  def self.sorted_most_similar_institutions
+    Institution.all
+               .map { |i| [i, i.most_similar_institution] }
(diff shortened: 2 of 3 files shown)


chvp (changes requested):
There's a lot of manual SQL here. Which is a bit unfortunate, since the ActiveRecord DSL makes it hard to write invalid SQL.

Testing on mestra got me an SQL error on the following page: https://mestra.ugent.be/nl/institutions/?order_by%5Bcolumn%5D=most_similar&order_by%5Bdirection%5D=ASC&page=1

jorg-vr:
@chvp Some of my plain sql was mariadb specific, I fixed that.

I also tried to use less plain sql, but Activerecord is not made for these complex queries.
Eg:
```sql
SELECT count(u.email) AS count, u.institution_id, u2.institution_id AS other_institution_id
FROM users u INNER JOIN users u2 ON u.email = u2.email
WHERE u.institution_id != u2.institution_id AND u.institution_id IS NOT NULL AND u2.institution_id IS NOT NULL AND u.email IS NOT NULL
GROUP BY u.institution_id, u2.institution_id
```
translates to
```ruby
User.select('count(users.email) AS count, users.institution_id, u2.institution_id AS other_institution_id')
    .joins('INNER JOIN users u2 ON users.email = u2.email')
    .where('users.institution_id != u2.institution_id AND users.institution_id IS NOT NULL AND u2.institution_id IS NOT NULL AND u2.email IS NOT NULL')
    .group('users.institution_id, u2.institution_id')
```
Which is not a lot better... 

I actually prefer the 100% sql query to the 80% plain sql, 20% rails combo which I find less readable.

jorg-vr:
Ignore my previous comment. I have found an equally fast rails native solution, by improving how I match usernames/emails/domains. (Used to be `O(n^2)` with `n` number of users, now it is `O(n log(n))` )

bmesuere (approved):
Is there a reason why you picked such a short cache time?

jorg-vr:
@bmesuere when the queries were really slow I had the cache time on one day, but now they are quite reasonable, so they don't cause that much delay for a first load.

In this case the time I want to cover is one user exploring the institutions page (On which I don't think they will often spend more than five minutes)

Then maybe they want to come back to the page later (for example after merging two institutions) to see the new situation. I think a caching time of more then five minutes could be annoying for that use-case.

app/controllers/institutions_controller.rb
```diff
@@ -5,11 +5,20 @@ class InstitutionsController < ApplicationController
...
+      scope
+    end
+  end
+
   def index
     authorize Institution
-    @institutions = apply_scopes(policy_scope(Institution)).all.order(generated_name: :desc, name: :asc)
-                                                           .includes(:courses, :users, :providers)
-                                                           .paginate(page: parse_pagination_param(params[:page]))
+    @institutions = apply_scopes(policy_scope(Institution.order_by_name(:asc))).all
```
chvp, line 19: It was already there previously, but I think this `all` is superfluous.

chvp (approved):

PR #4153, opened 2022-11-10
Add `unstarted_by` scope to exercise model

chvp:
Every once in a while, I have to look up exercises that are not solved by a group of users yet. Every time I have to re-invent the query to do so. This makes it so I won't have to re-invent the query.

### DIFF
--- a/app/models/exercise.rb
+++ b/app/models/exercise.rb
@@ -40,7 +40,7 @@ class Exercise < Activity
   before_save :check_memory_limit
 
   # Only used manually from the console
-  scope :unsolved_by, ->(users) { where.not(id: Submission.where(user_id: users).select(:exercise_id)) }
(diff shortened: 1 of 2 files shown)


app/models/exercise.rb
```diff
@@ -39,6 +39,8 @@ class Exercise < Activity
 
   before_save :check_memory_limit
 
+  scope :unsolved_by, ->(users) { where.not(id: Submission.where(user_id: users).select(:exercise_id)) }
```
jorg-vr, line 42: could you add a comment that this is used for manual command line checks?
I tend to remove code if I notice it has no usages anywhere in the code base

niknetniko (commented):
Do we want a test for this?

chvp:
While adding a test I realised that `unstarted_by` is a better name, so I changed it to that.

niknetniko (approved):

PR #4163, opened 2022-11-14
Use OIDC for elixir login

jorg-vr:
This pull request fixes logging in to Dodona using Elixir. The old method through SAML had issues when trying to log in multiple times, which I was unable to solve.

As they also support open id connect, we decided to try the connection this way.

The implementation is pretty similar to that of surf, so I tried merging them, but in the end left it as is because some minor details were still different. The overlap is also mostly setup and boilerplate, the actual shared logic is in the openid_connect gem.

All users logging in through this method will appear in a specific elixir institution. Once in production this new institution should be merged with the existing elixir institution (ID = 42). This should enable user to keep their existing account. (If they manage to have a succefull login through the old SAML system or by emailing us, only 26 users are impacted)


lib/Surf/strategy.rb
```diff
@@ -1,11 +1,13 @@
 require 'openid_connect'
 require 'openid_connect/response_object'
 
-# This strategy enables Surf and Elixir to be used by Dodona.
-# It adds a small change to the default open id connect, introducing the user institution
+# This strategy enables Surf to be used by Dodona.
+# It adds a smal change to the default open id connect, introducing the user institution
```
pdawyndt, line 5: small

chvp (approved):
What will happen to the current Elixir users?

jorg-vr:
@chvp these will still exist. On first sign in with the new provider they will be asked to confirm it is them and asked to sign in again using the old SAML provider.

If this works => they from now on they can sign in using the new provider
If this fails => They will get an error and will hopefully contact us (We can manually link them once we have their new credentials logged) There are only 26 users so a manual fix is doable

bmesuere (approved):

PR #4167, opened 2022-11-16
Fix table height for not fully loaded series

### LINKED ISSUE #3962: Series navigation should always bring selected series to top of viewport
In [this course](https://dodona.ugent.be/en/courses/773/) with multiple series, this is the view I get after clicking the "exam (group 2)" series in the series nagivation in the right margin:

image

Two issues:

1. The series is not on top of the viewport, even though there's enough content below it to make this happen.
2. Another series is highlighted in the series navigation.

This no longer happens after series skeletons have been replaced with the full rendering of the series. A height difference between skeletons and full renderings of series might cause this issue (@chvp witnesses having observed this issue upon page loading).

jorg-vr:
This pull request fixes the automatic scroll on course pages. I only replaced the slow fields of the table by a placeholder, instead of all the fields, allowing a search by exercise name as requested.

Below you can find the results of a speed comparison. For fair comparison I cleared my sql cache before each request. I tested on https://dodona.ugent.be/nl/courses/359/ which is a rather large course.
- Old: 2421ms/1798ms/1887ms
- Load course with all series: 14446ms/13961ms
- Load course with only progress bars hidden: 2445ms/2345ms/2298ms/2513ms
- New: 1875ms/1900ms/1991ms

Old:
image

New:
image

Closes #3962.

Replaced by #4169

### DIFF
--- a/app/views/activities/_series_activities_table.html.erb
+++ b/app/views/activities/_series_activities_table.html.erb
@@ -46,7 +46,7 @@
               <% end %>
               <%= render partial: 'activities/user_status_icon', locals: {activity: activity, series: local_assigns[:series], user: user} %>
             <% else %>
-              TBD
+              <i class="mdi mdi-check unloaded-skeleton-field skeleton-icon"></i>
             <% end %>
           <% end %>
         </td>
@@ -109,7 +109,7 @@
                 <% end %>
               <% end %>
             <% else %>
-              TBD
+              <div class="unloaded-skeleton-field skeleton-progress"></div>
             <% end %>
           </td>
         <% end %>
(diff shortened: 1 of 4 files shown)


chvp (approved):
<3

app/views/activities/_series_activities_table.html.erb
```diff
@@ -70,60 +76,72 @@
...
-                    tried: activity.users_tried(course: @course),
-                    correct: activity.users_correct(course: @course),
-                    info_tried: 'activities.index.progress_chart_info_tried',
-                    info_correct: 'activities.index.progress_chart_info_correct'
-                  }
-                %>
-              <% else %>
-                <%= content_tag :span, title: t('activities.activity.count_tooltip'), class: "text-end text-muted" do %>
-                  <%= activity.users_correct %>/<%= activity.users_tried %>
+            <%if loaded %>
```
chvp, line 79:
```suggestion
            <% if loaded %>
```

niknetniko (approved):

Issue #2882, opened 2021-06-24
Tracking: rework home page


bmesuere:
Right now, the home page contains
- A list of course cards for the courses you're subscribed to (grouped by AY, potentially with favorited on top)
- General stats like the number of submissions and solved exercises
- A list of upcoming deadlines over all courses
- A list of recent exercises over all courses

The list of courses is the most prominent, but might not be ideal for all users:
- Some users have a lot of courses (e.g., @pdawyndt has around 60), many of them not super relevant to show on the home page.
- Quite a lot of users only have a single course. Here, a lot of space is wasted and the home page doesn't show a lot of extra things.

In addition, we have a lot more data. The course cards now show some key numbers (submissions and users for teachers, submissions and solved exercises for students), but are they the most relevant? For teachers, maybe submissions in the last week is more relevant?

We added a lot of new features to Dodona over the years. Some of them could be added to the home page more prominently:
- Notifications
- (Open) questions (especially for teachers)
- Ongoing grading sessions (for teachers)
- Grading results (for students)

---

https://github.com/dodona-edu/dodona/issues/2882#tasklist-block-298a9224-f5c7-43b3-a284-2f864c52ab7e

### To Discuss
- for teacher course cards, some "live" statistics might be useful. Before we add those, we will experiment with adding better statistics on the course page #4340
- in the right sidebar, add some kind of activity feed? We should first think of what could be included here before we go in with it. Might be combined with the item below
- in the right sidebar certain cards that only show up when they need to be. We could probably rework the current notification implementation
  - for teachers: a card with open questions
  - for teachers: a card with registration approvals
  - for teachers: pending evaluations?
  - exports
  - for students: evaluation results
  - for students, questions that were answered
- in the right sidebar, a place to put "non urgent" messages, such as new dodona versions or feature highlights #4172
- if we would have a lot of content for the sidebar, we could double its size for users with only a single course 
- a variant of the current "statistics" Maybe a little graph? a streak indicator?

bmesuere:
The homepage currently is quite static and could do more to surface relevant information. Especially for users that are only subscribed to a single course, the home page is relatively empty. This could be improved by adding a row of "action cards" to the top of the home page:

- The leftmost card is "jump back in" and is shown in blue. This shows a recent exercise that wasn't yet correctly solved. If there are no such recent exercises, the card could lead to a recent incomplete series. Clicking on the card brings the user to the respective exercise, potentially with their last submission already loaded in the editor.
[...]

PR #4185, opened 2022-11-21
Home page redesign: put most important content on top

jorg-vr:
This pull request

- [ ] Tests were added

Part of #2882

### DIFF
--- a/app/views/pages/home.html.erb
+++ b/app/views/pages/home.html.erb
@@ -1,6 +1,7 @@
 <%= content_for :javascripts do %>
   <%= javascript_include_tag 'favorite_course_buttons' %>
 <% end %>
+<%= render partial: "pages/important_homepage_links" %>
 <div class="row">
   <div class="col-12 col-md-6 col-lg-8">
     <% if @subscribed_courses.empty? %>
--- a/app/views/pages/_important_homepage_links.html.erb
+++ b/app/views/pages/_important_homepage_links.html.erb
@@ -1,0 +1,48 @@
+<div class="row">
+  <div class="col-3">
+    <h4>Jump back in</h4>
+    <div class="card">
+      <div class="card-title">
+        <h2 class="card-title-text course-name ellipsis-overflow d-block">Echo function print</h2>
+        <h3 class="card-subtitle-text ellipsis-overflow">Object oriented programming</h3>
+        <h3 class="card-subtitle-text ellipsis-overflow mt-1">
+          Series 5: Loops
+        </h3>
+      </div>
+    </div>
+  </div>
+  <div class="col-3">
+    <h4>Upcoming deadlines</h4>
+    <div class="card" style="background: #A5D6A7">
+      <div class="card-title">
+        <h2 class="card-title-text course-name ellipsis-overflow d-block" style="color: rgb(0 0 0 / 75%)">Series 2: Variables</h2>
+        <h3 class="card-subtitle-text ellipsis-overflow" style="color: rgb(0 0 0 / 60%)">Object oriented programming</h3>
+        <h3 class="card-subtitle-text ellipsis-overflow mt-1" style="color: rgb(0 0 0 / 60%)">
+          2 hours left
+        </h3>
+      </div>
+    </div>
+  </div>
+  <div class="col-3 align-self-end">
+    <div class="card" style="background: #EF9A9A">
+      <div class="card-title">
+        <h2 class="card-title-text course-name ellipsis-overflow d-block" style="color: rgb(0 0 0 / 75%)">Series 5: Loops</h2>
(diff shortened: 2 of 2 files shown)


bmesuere (changes requested):
Sorry for the brevity, but GitHub ate my draft so I have to write everything again 😢 

---
- [x] Deploy on mestra and signing in as staff triggers an error
image

Unrelated: our normal user in the dummy data has no submissions and no courses

---

- [ ] As Zeus, line wrapping messes up the layout:
image

---

- [x] Jump back in cards might seem a bit random. It would be good to add an explanation of why the card is shown. I would limit the course/series to a single line (e.g. separated with a `>`) as a subheader (maybe a bit smaller like the academic year in the course cards) and then add a line of text as supporting text. The same could be done with the series cards to be a bit more explicit "You solved all exercises, good job".

image

---

- [x] Still not sold on the appearance of the cards. They are both similar (shadow) and dissimilar (less shadow) to the course cards. Note that according to the spec, a card can be clickable as a whole, so no need to style as a button.

I would like to see a mockup where the top cards are in filled and outline variant using the header color as bg and border (lighter and darker variant respectively).

Alternatively, they could be style as normal cards, similar to the course cards, but include a "continue" button in the bottom right.

---

- [x] Could we wait with removing the sidebar cards that were removed until we test drive this for a release? The remaining cards seems a bit lonely otherwise.

---

- [ ] In a next iteration, we could improve on these cards for course admins. We could for example show the class progress in the progress bar. In that case, I would add the "staff" icon to the top right of the card similar to the course cards.

---

- [x] Just an idea, for the series cards, could we use the "series status" icon we use on the course page? This would improve the recognizability that this is about a series and the state would be displayed in a similar way. (We could also show the seasonal variations here).

bmesuere (changes requested):
I really like this change 👍 
Two minor UI remarks:
- I feel that the gap between title and subtitle is a bit too large. This is most prominent in the exercise card where the info text is closer to the subtitle than the title. This is probably because you used a `<br>` in combination with the bigger line height of the header tag.
- Is it possible to add a slight shadow on hover?

config/locales/views/pages/nl.yml
```diff
@@ -27,6 +28,23 @@ nl:
...
       favorite: "Favoriete cursus"
+    important_homepage_links:
+      jump_back_in: Ga verder
+      upcoming_deadlines: Aankomende deadlines
+    clickable_homepage_cards:
+      course:
+        jump_back_in_info: "Werk verder aan deze cursus"
+      activity:
+        jump_back_in_info: "Werk verder aan deze oefening"
+        jump_back_in_info_success: "De vorige oefening is correct opgelost. Start met deze oefening."
```
bmesuere, line 39:
```suggestion
        jump_back_in_info_success: "De vorige oefening was correct, start met deze oefening."
```

bmesuere (changes requested):
The UI changes look good 👍 

There's still a small issue with the wording of the progress bar tooltip: in a series of 4 exercises where 2 are correct and 2 unstarted, it now says: 2 started, 2 correct. This is confusing because the 2 started ones are the two that are correct. Also, the order should follow the order of the bar itself: correct, incorrect, not yet started.

I have no remarks about the code.

niknetniko (approved):
Besides some minor spacing nitpicks, it all looks good to me :+1:

jorg-vr:
I updated the tooltip to 
image

config/locales/views/pages/nl.yml
```diff
@@ -27,6 +28,26 @@ nl:
...
+        jump_back_in_info: "Je hebt de vorige reeks afgewerkt, start met deze reeks."
+        progress_chart_info_correct:
+          one: "Eén oefening correct opgelost."
+          other: "%{count} oefeningen correct opgelost."
+        progress_chart_info_wrong:
+          one: "Eén oefening incorrect."
+          other: "%{count} oefeningen incorrect."
+        progress_chart_info_not_started:
+          one: "Eén oefening niet begonnen."
+          other: "%{count} oefeningen niet begonnen."
```
bmesuere, line 50:
```suggestion
          one: "Eén oefening nog niet begonnen."
          other: "%{count} oefeningen nog niet begonnen."
```

bmesuere (approved):

bmesuere:
(maybe wait to merge this one till after the exercise filter PRs are merged and released as 6.3?)

app/views/pages/home.html.erb
```diff
@@ -10,7 +11,11 @@
       <% end %>
     <% else %>
       <div class="row favorites-row">
-        <% unless @favorite_courses.empty? %>
+
+        <div class="page-subtitle first">
+          <h3><%= t('.courses') %></h3>
+        </div>
+            <% unless @favorite_courses.empty? %>
```
chvp, line 18:
```suggestion
        <% unless @favorite_courses.empty? %>
```

chvp (approved):

PR #4186, opened 2022-11-22
Show teacher rights notification on public contact page

jorg-vr:
This pull request shows the teacher rights notification also when not signed in. Hopefully this increases the amount of users that find the form on their own.

I removed the policy check, as that only checked if a user is present


chvp:
I think the new is there because otherwise we saw people using the contact form to request rights when they already had them.

chvp (approved):

niknetniko (approved):

bmesuere (approved):

PR #4187, opened 2022-11-22
Add a custom captcha to the contact form

jorg-vr:
This pull request adds a custom captcha. The goal is to receive less spam. The idea being that even though the captcha is simple, no one will actually create a custom captcha solver for Dodona.

image


jorg-vr:
Closed in favor of replacing our captcha by https://www.cloudflare.com/en-gb/products/turnstile/ in the future

Custom captchas tend to confuse end users and there is no guarantee that bots will fail

PR #4188, opened 2022-11-23
Seasonal series completion icon

tibvdm:
# Temporal overlay for Christmas

assembled_plain

Ho Ho Ho, Santa has found his way to Dodona. He will be sitting on your status icons in order to watch your progress. Santa left his hat on every single status icon. When you hover the icon, he will come forward and take a peek.

Screenshot 2022-11-23 at 11 56 34

Screenshot 2022-11-23 at 11 56 55

This PR adds:
- A new `SeasonalHelper` that selects some css-styling based on the current time. At this moment I only do a check for Christmas (Dec 7 until Dec 31)
- Christmas stylesheet in orde to place a santa hat on top of "green checkmarks". On hover this will become Santa. Santa is also present on "rainbow checkmarks" without the hover functionality.

Screenshot 2022-11-25 at 09 44 19


app/helpers/temporal_helper.rb
```diff
@@ -0,0 +1,14 @@
+module TemporalHelper
+  def christmas(current_time)
+    return current_time.month == 12 && current_time.day.between?(25, 31)
```
jorg-vr, line 3: I would include the 24th because Christmas eve also deserves a smiling Santa

app/views/series/_series_status.html.erb
```diff
@@ -1,15 +1,27 @@
...
 <% if loaded && user.present? %>
-  <% deadline_class, deadline_icon = series_status_deadline(series, user) %>
-  <% progress_class, progress_icon = series_status_progress(series, user) %>
-  <div class="card-title-icon <%= progress_class %> <%= deadline_class %>" title="<%= series_status(series, user) %>" data-bs-toggle="tooltip" data-bs-placement="top">
-    <i class="mdi <%= progress_icon %>"></i>
-    <% if deadline_icon %>
-      <i class="mdi mdi-18 <%= deadline_icon %>"></i>
-    <% end %>
-  </div>
+    <% deadline_class, deadline_icon = series_status_deadline(series, user) %>
```
jorg-vr, line 3: There are some unneeded white space changes here, could you avoid those?

app/views/series/_series_status.html.erb
```diff
@@ -1,15 +1,27 @@
...
+    <% overlay_class = series_status_overlay() %>
+
+    <div 
+        class="card-title-icon <%= overlay_class %> <%= progress_class %> <%= deadline_class %>" 
+        title="<%= series_status(series, user) %>" 
+        data-bs-toggle="tooltip" 
+        data-bs-placement="top"
+    >
+        <% if overlay_class %>
+            <div class="overlay"></div>
```
jorg-vr, line 14: I think you can avoid this extra div using an ::after element in css

app/assets/stylesheets/temporal/christmas.css.scss
```diff
@@ -0,0 +1,71 @@
...
+            background-image: url('temporal/series_status_overlays/christmas-hover.svg') !important;
+          }
+      
+          i {
+            color: transparent;
+            background: transparent;
+            border: transparent;
+          }
+      
+          &.deadline-missed .mdi-alarm-off {
```
jorg-vr, line 27: I do not like the fact that you have to manually set every possible combination with other icons to transparent.
This will result in errors if new classes are ever added.

Either display the overlay on top (By specifying a higher z index)
Or try to find a more general css query that matches all desired classes. (You can use `!important` to make sure your css has priority over more specific classes)

jorg-vr (changes requested):
I really like the santa figure, thanks for the contribution!

BTW: I'll give you access to create branches on dodona, so you don't need to work from your own fork if you want to contribute more in the future.

app/helpers/temporal_helper.rb
bmesuere: I would even start this as soon as december first

pdawyndt: My kids would strongly object doing this earlier than December 7th: no competition allowed!

Gemfile.lock
```diff
@@ -1,579 +0,0 @@
-GEM
```
jorg-vr, line 1: You chose the wrong method to exclude it...
Please do not remove it from the repo, only exclude changes made by you

jorg-vr (changes requested):
Except for the lockfile, everything looks good.

Side note: Once you have applied changes you can click on the circle arrows (Top right) to re request a review, which will notify me

bmesuere (changes requested):
I deployed this locally and it looks really nice! 

I do agree with @chvp that this might be a bit much if we apply it to all series icons for the entire month and add other seasonal decorations in the future.

We could reduce the number of appearances by only applying the hat to the "green checkmark" icon (showing santa on hover) and show santa (without hover) instead of the "rainbow checkmark". This would make santa some kind of achievement. What do the others think?

I also checked if we could replace the icons that are shown on correct submission by a set of christmas-themed icons. Unfortunately, mdi doesn't include a lot of christmas icons.

Regarding the code: there are a lot of `!important`'s in the css which is something we try to avoid. Are these strictly necessary?

app/views/series/_series_status.html.erb
```diff
@@ -2,8 +2,16 @@
 <% if loaded && user.present? %>
   <% deadline_class, deadline_icon = series_status_deadline(series, user) %>
   <% progress_class, progress_icon = series_status_progress(series, user) %>
-  <div class="card-title-icon <%= progress_class %> <%= deadline_class %>" title="<%= series_status(series, user) %>" data-bs-toggle="tooltip" data-bs-placement="top">
+  <% overlay_class = series_status_overlay() %>
```
chvp, line 5:
```suggestion
  <% overlay_class = series_status_overlay %>
```

chvp (approved):

niknetniko (approved):

tibvdm:
I updated the description and added an extra image to show the different overlay options.

PR #4203, opened 2022-11-29
Add scoping tabs to activity list pages to improve usability

### LINKED ISSUE #4175: Improve scoping of repositories
Problem:
- With the increasing amount of advanced users, there are an increasing amount of repositories and exercises
- This even causes a lot of duplicates, with directories being copied as a first test and then never edited
- this makes it hard to find own exercises

Solution:
- I would suggest scope filters similar to the course list: Featured, Mine, My institution
- Make it harder to make your repository public. Default repo's should be private, setting it public should require a confirm

jorg-vr:
This pull request adds tabs when selecting activities. This should make it easier to find relevant exercises.
image

They all have a tooltip to explain their content
image

I created a generic filter-tabs component which also replaces the existing course tabs. I will also replace the course members tabs with this component in a future pr.

I also added a field `featured` to repositories, making it possible for us to put certain sets of decent activities in the spotlight. This field can only be edited by zeus.
image

Sorting of activities is also improved, with activities without a name in your language now being sorted last instead of first. This could be further improved in future work by allowing user to sort by every column (Eg to sort by how often an activity is used to find the most popular, see #4208 )

- [x] Tests were added

Closes #4175  .


bmesuere (changes requested):
Feedback from a quick functional test on naos:

My first reaction was a bit of confusion. Part of this might be because exercises/activities is not repeated in the tab name for all tabs (as we do on the courses page), part of this is because I wasn't expecting a tab with "Universiteit Gent" or "Scriptingtalen" as name.

Some suggestions:
-[x] Include "activities" in the name for all tabs
-[x] For the institution, use the short name (as with the courses page)
-[x] There should be a little help text explaining what is shown: or as a (bootstrap) tooltip on the tab name or as a small help text above the table
-[ ] Even with a help text, the course specific exercises are hard to explain. Wouldn't it make more sense (and easier to explain) to add this category as a search filter private/public instead?
-[ ] Suggestions for other names for "featured" are welcome. It's ok, but to me featured indicated a limited number but this section will contains hundreds of exercises. "curated" might be an alternative, but I'm also not sure about that. If we don't find a better name, featured is ok.
-[ ] The active tab isn't stored in local storage (as is done with the filters) which causes reloading (with filters selected) to fail
-[ ] Popularity of an exercise is definitely a strong indicator of interestingness of the exercise so if possible we should surface that information in the table. I don't think we want to list the exact number of courses where an exercises is used, but maybe we can use an icon with different states: low, medium, high, featured. We will then have to pick sensible threshold values. I don't know what the best strategy is for fetching this information: calculate it on demand, store it in the DB or store it in cache. In addition, we could add a filter for it.

jorg-vr:
> Even with a help text, the course specific exercises are hard to explain. Wouldn't it make more sense (and easier to explain) to add this category as a search filter private/public instead?

I agree that this is hard to explain, but I do not want to replace it with a public private filter. 
If a course has been granted access to a specific private repository, it is very likely that this will be the main source for exercises. This is why I made it the first suggested tab when in the context of a course.
If we make it a public/private filter, these activities will only be available in the "All activities" tab or the "My institution" tab. In both cases they will be mixed with results form your own private repositories if you filter on "private"

jorg-vr:
Course activities and my activities have been merged

bmesuere (changes requested):
👍 
- [x] Did you try a bootstrap popover instead of the title attribute for the info text? It's a bit hidden right now.
- [x] What's the workflow to "feature" a repository? I couldn't see a checkbox on the repo edit page on naos.
- [ ] I'm not a fan of "recommended exercises", this gives the impression that the list is tailored towards this course/user/series. For the lack of a better name, I guess we can use featured?
- [ ] I noticed that the list of labels always ends with a `,` (not strictly related to this PR)

app/assets/javascripts/components/filter_tabs.ts
```diff
@@ -0,0 +1,55 @@
...
+                this.select(this.labels[0]);
+            }
+        }
+    };
+
+    render(): TemplateResult {
+        return html`
+            <div class="card-tab">
+                <ul class="nav nav-tabs" role="tablist">
+                    ${ this.labels.map( label => html`
```
chvp, line 44:
```suggestion
                    ${this.labels.map(label => html`
```

app/models/activity.rb
```diff
@@ -86,6 +86,21 @@ class Activity < ApplicationRecord
     by_language
   }
 
+  enum repository_scope: { mine: 1, my_institution: 2, featured: 3, all: 0 }, _prefix: true
```
chvp, line 89: I'm a bit confused by this enum. Where is it used? AFAIK, this should correspond to a matching `repository_scope` field in the database, which is not there (and wouldn't make sense in this case (?)).

chvp (changes requested):

app/models/activity.rb
jorg-vr: These are all the options for the scope defined below. I used to use this to generate all tabs, but as they become more complex it was no longer that useful. I'll remove it

app/assets/javascripts/components/filter_tabs.ts
```diff
@@ -0,0 +1,55 @@
...
+ */
+@customElement("d-filter-tabs")
+export class FilterTabs extends watchMixin(FilterCollectionElement) {
+    @property()
+    multi = false;
+    @property()
+    param = "tab";
+    @property()
+    paramVal = (label: Label): string => label.id.toString();
+    @property( { type: Array })
```
niknetniko, line 22:
```suggestion
    @property({type: Array})
```

niknetniko (commented):
Besides one minor formatting nitpick, I noticed some weird behaviour when using the back button:

1. Go to the activities (or course) page
2. Press the browser back button

All params are now removed from the URL, which results in a state where no tab is selected, although they are unchanged:
image
(this is also what happens after you have done a few searches)

bmesuere (approved):
This is a good first step in solving our exercise discovery problem 👍

niknetniko (approved):

chvp (approved):

PR #4127, opened 2022-11-03
submission.js: remove jquery and convert to typescript

freyavs:
This pull request converts `submission.js` to `submission.ts`, removing jquery and converting everything to typescript.

It is progress on #3590

PR #4209, opened 2022-12-02
drag_and_drop.js: remove jquery and convert to typescript

freyavs:
This pull request converts drag_and_drop.js to drag_and_drop.ts, removing jquery and converting everything to typescript.

It also adds a new util function `findParent` to replace the jquery `.parents(...)`  method.

This is progress on #3590


app/assets/javascripts/util.js
```diff
@@ -191,6 +191,23 @@ function htmlEncode(str) {
     });
 }
 
+/**
```
jorg-vr, line 194: Could you also use this function in the other places where it is relevant. Such as https://github.com/dodona-edu/dodona/pull/4127/files#diff-c5bc65df73e1c820b39e10ea1e08c46f7ee47491e7b38b6134c380414573a9e7R14

jorg-vr (approved):
Code looks good :)

app/assets/javascripts/drag_and_drop.ts
```diff
@@ -0,0 +1,64 @@
...
+type DragAndDropArguments = {
+    table_selector: string,
+    item_selector: string,
+    item_data_selector: string,
+    order_selector: string,
+    order_data_selector: string,
+    url_from_id: (courseId: string) => string;
+}
+
+function copyWidth(clone: HTMLCanvasElement, original: HTMLCanvasElement, tag=undefined): void {
```
niknetniko, line 22: Why are these `HTMLCanvasElement`? The call on line 48 to this function seems to pass `td` elements.

app/assets/javascripts/drag_and_drop.ts
```diff
@@ -0,0 +1,64 @@
...
+    const tableBody = document.querySelectorAll(args.table_selector)[0];
+
+    dragula([tableBody], {
+        moves: function (el, source, handle, sibling) {
+            return handle.classList.contains("drag-handle") || findParent(handle, "drag-handle");
+        },
+        mirrorContainer: tableBody,
+    })
+        .on("cloned", function (clone, original, type) {
+            copyWidth(clone, original, "td");
```
niknetniko, line 48: We should add `@types/dragula` to the dev dependencies, as this will help with stuff like the `copyWidth` function expecting `HTMLCanvasElements`, while these don't seem to be canvas elements.

app/assets/javascripts/drag_and_drop.ts
```diff
@@ -0,0 +1,64 @@
...
+    for (let i = 0; i < cloneChildren.length; i++) { // make all children equally big
+        cloneChildren[i].width = originalChildren[i].clientWidth;
+    }
+}
+
+function initDragAndDrop(args: DragAndDropArguments): void {
+    const tableBody = document.querySelectorAll(args.table_selector)[0];
+
+    dragula([tableBody], {
+        moves: function (el, source, handle, sibling) {
```
niknetniko, line 42: Can these functions become arrow functions?

niknetniko (changes requested):

app/assets/javascripts/drag_and_drop.ts
freyavs: Right, they are `HTMLTableRowElement`. I was getting type errors that `.width` doesn't exist on this type, so I was testing if it worked with other types but hadn't changed it back, apparently. Actually, `.width` is deprecated anyway, so probably shouldn't be used here. I'll fix this.

app/assets/javascripts/drag_and_drop.ts
```diff
@@ -0,0 +1,63 @@
...
+type DragAndDropArguments = {
+    table_selector: string,
+    item_selector: string,
+    item_data_selector: string,
+    order_selector: string,
+    order_data_selector: string,
+    url_from_id: (courseId: string) => string;
+}
+
+function copyWidth(clone: Element, original: Element, tag: string = undefined): void {
```
niknetniko, line 22: Taking another look at this, this function is only ever used with "td" as the third param, so the case where the tag is undefined can be removed (if we were to need this, I think we propably should use `.children` instead of `childNodes`, as the latter includes nodes for text, which are not HTML elements, so they e.g. don't have a `clientWidth`)

niknetniko (changes requested):

app/assets/javascripts/drag_and_drop.ts
freyavs: I removed it, also had to cast the cloned children to `HTMLElement` as typescript was complaining that `Element` doesn't have the property `style`

app/assets/javascripts/util.js
```diff
@@ -191,6 +191,23 @@ function htmlEncode(str) {
...
+ * Searches the parents of an element until it finds the parent containing certain classes
```
bmesuere, line 195:
```suggestion
 * Returns the first parent of an element that has at least all of the given classes. Returns null if no such parent exists.
```

bmesuere (changes requested):

niknetniko (approved):

bmesuere (approved):
👍

jorg-vr (approved):

PR #4222, opened 2022-12-08
Show popularity in activities table

### LINKED ISSUE #4208: Order activities by popularity
Popularity of an exercise is definitely a strong indicator of interestingness of the exercise so if possible we should surface that information in the table. I don't think we want to list the exact number of courses where an exercises is used, but maybe we can use an icon with different states: low, medium, high, featured. We will then have to pick sensible threshold values. I don't know what the best strategy is for fetching this information: calculate it on demand, store it in the DB or store it in cache. In addition, we could add a filter for it.

_Originally posted by @bmesuere in https://github.com/dodona-edu/dodona/pull/4203#pullrequestreview-1199499438_

jorg-vr:
This pull request

- [ ] Tests were added

Closes #4208  .

### DIFF
--- a/app/controllers/activities_controller.rb
+++ b/app/controllers/activities_controller.rb
@@ -63,7 +63,7 @@ class ActivitiesController < ApplicationController
 
     unless @activities.empty?
       @activities = apply_scopes(@activities)
-      @activities = @activities.order("name_#{I18n.locale}").order(path: :asc).paginate(page: parse_pagination_param(params[:page]))
+      @activities = @activities.joins(:courses).group(:id).order(Arel.sql('COUNT(DISTINCT courses.id) DESC')).paginate(page: parse_pagination_param(params[:page]))
     end
     @labels = policy_scope(Label.all)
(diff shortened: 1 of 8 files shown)


bmesuere (changes requested):
😻 
Is it possible to also add this as a filter?

db/migrate/20221212094630_add_series_count_to_activities.rb
```diff
@@ -0,0 +1,9 @@
+class AddSeriesCountToActivities < ActiveRecord::Migration[7.0]
+  def change
+    add_column :activities, :series_count, :integer, :default => 0, :null => false
+    Activity.reset_column_information
+    Activity.all.each do |e|
```
bmesuere, line 5: I think this will instantiate an object for all of the activities at once. I think `find_each` is made for this specific case.

bmesuere (approved):

niknetniko (approved):

PR #4245, opened 2022-12-14
Only show course from latest year on home page

jorg-vr:
This pull request removes old course cards from the home page, only showing the latest year. Using a dropdown it allows to cycle through previous years.

Favorited courses are always shown on top.
image

image

No dropdown is shown when there are less then two different years
image

Part of #2882

### DIFF
--- a/app/assets/javascripts/components/standalone-dropdown-filter.ts
+++ b/app/assets/javascripts/components/standalone-dropdown-filter.ts
@@ -31,7 +31,7 @@ export class StandaloneDropdownFilter extends watchMixin(FilterCollectionElement
     };
 
     render(): TemplateResult {
-        if (this.labels.length === 0) {
+        if (this.labels.length <= 1) {
             return html``;
         }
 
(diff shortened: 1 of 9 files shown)


bmesuere (changes requested):
👍 Looks good, but I'm not sure about the current favorite behaviour. Shouldn't we add a separate "year" for favorites?
Two smaller things:
- Can you make the dropdown right-aligned (do the dropdown aligns with the right of the button?) Should be possible with `<div class="dropdown-menu dropdown-menu-right">`
- Can you style the years with an n-dash (?) as is done on the course cards?

bmesuere:
> 👍 Looks good, but I'm not sure about the current favorite behaviour. Shouldn't we add a separate "year" for favorites?

As an alternative, keep the current behaviour, but place the favorite courses in a separate "favorite courses" section above "courses".

bmesuere (approved):

chvp (approved):

PR #4261, opened 2022-12-20
Homepage course cards redesign

jorg-vr:
This pull request contains a redesign of the course cards on the home page.

Main design goals:
- Provide more useful information (Progress bars instead of total submissions number)
- Provide links directly to the latest/next exercise (Limit number of clicks before the most used page is reached)
- Improve consistency and style (All cards same height)

Just like the top cards (see #4185) I've tried to show to always show 2 deadlines (1 recently passed and/or up to 2 upcoming) and 1 activity to continue or start working on. Only visible series and activities within visible series are shown. More activities are shown when less then two deadlines are available.
If not enough activities can be found, this means the student has almost finished and a motivational message is shown.

Indexes were added to activity_statuses speed up calculating the progress bar and finding the latest worked on activity.

The resulting cards:
image
image

Christmas view for the fans:
image)

- [x] Tests were added

Progress on #2882


jorg-vr:
Mockups:
V1:
image
image
V2:
image
image
image
image
V3:
image

jorg-vr:
Always same height
image

jorg-vr:
Empty state mockup
image

jorg-vr:
WIP
image
image

pdawyndt:
Small remark that pops up when seeing these new cards (which I like a lot): by checking if a student completed a course or how far a student is from completing a course, we make the assumption that no extra series/exercises will be added to the course. Might look strange if the latter does happen. Say, for example, that a teacher opens a course for registration, but has no series/exercises in the course yet (maybe closed, but not visible to students). Then students will already start with a completed course.

pdawyndt:
When I tested on naos, loading took a long time after switching to another academic year.

pdawyndt:
The congratulations icon might do with some additional padding on top (maybe make image a tiny bit smaller to preserve the same height for all cards).

pdawyndt:
When combination of teacher name(s) and institute name is too long, the academic year might overflow to the next line, which breaks the fact that cards should have the same height.

image

pdawyndt:
Idea for "jump back in" when you don't have registered for any course yet: put option in "jump back in" to start selecting courses.

bmesuere:
- I liked the striped icon better
- the progress bar in the header needs a bit more padding on top, it is too close to the teacher name compared to the bottom padding
- "deze cursus afgewerkt" -> "alle oefeningen opgelost". I think we only include exercises in public series?
- the version with only 2 items looks a bit odd, there's too much spacing. Maybe make the "goed bezig" bigger like the one below?
- maybe put the newline after "gefeliciteerd" in the 2-line version? (but maybe the new text won't fit on a single line)

bmesuere:
Some ideas for teachers (could be for a next PR).
- number of open questions (or at least an indicator)
- class progress of deadline series
- mini graph showing activity / number of submissions and active users in last hour
- number of user/submissions

config/locales/views/pages/en.yml
```diff
@@ -29,9 +29,9 @@ en:
       course-closed: Closed course
       favorite: "Favorite course"
       few-activities-markdown:
-        one: "**Doing great!** Only one activity left to complete this course."
-        other: "**Doing great!** Only two activities left to complete this course."
-      no-activities-markdown: "**Congratulations!** You have completed this course!"
+        one: "**Doing great!** Only one activity left and you have solved all exercises."
```
pdawyndt, line 32: No consistent use of activities vs exercises. Is check for all activities or all exercises?

pdawyndt (changes requested):

jorg-vr:
kleur van cursus kijkt te veel op knoppen (Misschien terug ligte variant)

Oefening => naam van de reeks kan nuttige context zijn

Splits de beschrijving voor leesactiviteit en oefening

Goed bezig tekstje => oefening ipv activiteit

Bug tooltip blijft staan bij jaar wijziging

Altijd huidige reeks tonen bij oefening als er geen deadline is

Mogelijke bug => bart cursus programeren 2019-2020

Indicatie van fout/ al bezig

jorg-vr:
Series icons as svg's of different sizes, with the old version at the bottom
image

image

jorg-vr:
image
image
image
image

jorg-vr:
Draft of exercises status overlayed
image

bmesuere (approved):
I really like these changes, nice work!

I prefer the lighter header color (in light mode) that is now used. The darker one was to similar to the top actions and too "boring".

I also like the use of the exercise status icon when there is a previous submission. The overlayed variant is probably a bit too small.

A small suggestion to add a little bit of context to the exercises shown: in the title attribute of the exercise name, you could add "continue with xxxxx" or "start with xxxxx".

app/views/application/_progress_chart.html.erb
```diff
@@ -21,11 +26,12 @@
       .filter { |v| binding.local_variable_defined? "info_#{v}" }
       .map { |v| t(binding.local_variable_get("info_#{v}"), count: binding.local_variable_get(v)) }
       .join('<br />')
+    muted ||= false
 %>
 
 <%# You might think that this span here is useless, but you think wrong: https://github.com/twbs/bootstrap/issues/20381 %>
 <span data-bs-html="true" data-bs-toggle='tooltip' title="<%= tooltip %>">
-  <svg viewBox="0 0 1 1" preserveAspectRatio="none" height="8" width="<%= width %>" class='progress-chart'>
+  <svg viewBox="0 0 1 1" preserveAspectRatio="none" height="8" width="<%= width %>" class='progress-chart <%= muted ? 'colored-muted' : ''%>'>
```
chvp, line 34:
```suggestion
  <svg viewBox="0 0 1 1" preserveAspectRatio="none" height="8" width="<%= width %>" class='progress-chart <%= muted ? 'colored-muted' : '' %>'>
```

app/models/course.rb
```diff
@@ -195,6 +195,49 @@ def homepage_series(passed_series = 1)
     passed_deadlines.to_a + future_deadlines.to_a
   end
 
+  def homepage_activities(user, limit = 3)
```
chvp, line 198: I think this can be a private method?

app/models/course.rb
```diff
@@ -351,6 +394,18 @@ def color
...
+  end
+
+  def completed_activity_count(user)
+    ActivityStatus.where(accepted: true, user: user, series: series.visible).count
+  end
+
+  def started_activity_count(user)
+    ActivityStatus.where(started: true, user: user, series: series.visible).count
+  end
+
```
chvp, line 408: Same here?

chvp (approved):

app/models/course.rb
jorg-vr: these methods are used in the _course_card view. I am pretty sure that would break if they were private?

chvp: Ah, never mind then.

niknetniko (approved):
Looks very good!

PR #4266, opened 2022-12-25
series.js: remove jquery and convert to typescript

freyavs:
This pull request removes all jquery from series.js and converts it to typescript. It also fixes some small bugs.

**Fix 1:** a small bug for the css on activities: before, when adding `pending`  and `new` to an activity tr-element, this didn't have any effect on the css. Now when adding `pending` or `new`, the opacity is correctly changed (in the short time frames it takes to add an activity to a series):

Before (when `pending`):

After (when `pending`):

**Fix 2:** I also made sure that a user can't add/remove the same activity multiple times. This was possible when it was taking a longer time for adding/removing an activity (e.g. waiting on a response), so these changes are only visible for this edge case. For adding, the add button was not hidden when a series is being added (in the `new` and `pending` stages). The add button only became hidden when the series was successfully added.  This made it possible to add multiple instances of the same activity. The new flow is 
* the add button gets hidden directly, and becomes visible again on `addingActivityFailed`
* For removing, the remove button gets hidden (see image below), and becomes visible on `removingActivityFailed`.

Before (when `pending`):

After (when `pending`):

This is progress on #3590


app/javascript/packs/evaluation.js
```diff
@@ -1,4 +1,4 @@
-import { initDeadlinePicker } from "series.js";
+import { initDeadlinePicker } from "series.ts";
```
jorg-vr, line 1: It is probably cleaner to directly import `initDatePicker` from `util.js` as series.ts doesn't add anything

app/javascript/packs/series.js
```diff
@@ -1,4 +1,4 @@
-import { initDeadlinePicker, initSeriesEdit, initSeriesShow } from "series.js";
+import { initDeadlinePicker, initSeriesEdit, initSeriesShow } from "series.ts";
```
jorg-vr, line 1: Same comment as above

jorg-vr (approved):
Code looks good :) Thanks for the bug fixes!

Just found one small extra improvement, which we might as well include while we're busy refactoring this code

bmesuere:
(I would wait to merge this one till after the exams)

niknetniko (approved):

chvp (approved):

PR #4267, opened 2022-12-25
export.js: remove jquery and convert to typescript

freyavs:
This pull request removes jquery from export.js, and converts it to typesciprt.

It is progress on #3590


niknetniko (approved):

jorg-vr (approved):

bmesuere (approved):
Can we ignore the CodeQL warning?

jorg-vr:
Yes we can ignore it. We are using the `value` of an `input` field as a url parameter. Theoretically this could be an unsanitized uncontrolled parameter.

In our case the input field is a checkbox, and the value is always a numeric id set by our back-end. So I see no reason to sanitize this value

PR #4283, opened 2023-01-02
Don't show activities from hidden series as jump-back in option

jorg-vr:
This pull request stops activities from hidden series being shown as jump back in options on the home page.

This used to happen when a users most recent activity had been in that hidden series.

- [x] test where added


app/models/user.rb
```diff
@@ -470,6 +470,8 @@ def jump_back_in
       course: latest_submission.course
     }
 
+    return result if latest_submission.course.present? && (latest_submission.series.nil? || !latest_submission.series.open?)
```
bmesuere, line 473: Can you add a little comment here?

bmesuere (approved):

chvp (approved):

niknetniko (approved):

PR #4298, opened 2023-01-04
repository.js: remove jquery and convert to typescript

freyavs:
This pull request removes jquery from repository.js and converts it to typescript. There's also been a small change in the way buttons are shown (they get hidden in stead of removed).

It also adds a small style fix:  When the add button was removed, the row changed in height. Now heights stay the same

Before:

After:


app/controllers/repositories_controller.rb
```diff
@@ -196,12 +196,12 @@ def model_update_response(success, model, return_path)
       if success
         toast = t('controllers.updated', model: model.model_name.human)
         format.json { head :no_content }
-        format.js { render locals: { toast: toast } }
+        format.js
```
jorg-vr, line 199: You can remove this line as we no longer support js as format

app/assets/javascripts/repository.ts
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    function onAddClick(e: Event): void {
+        const button = e.currentTarget as HTMLButtonElement;
+        const cell = button.closest("td.repository-admin-button-cell") as HTMLTableCellElement;
+        const userId = cell.dataset.user_id;
+        const repositoryId = cell.dataset.repository_id;
+        const oldRow = cell.parentElement.closest("tr");
+
+        fetch(`/repositories/${repositoryId}/add_admin.js`, {
```
jorg-vr, line 30: You can change these paths to json (Or just omit the file type as it does not matter which response format you get as you're no longer using the result)

jorg-vr (changes requested):
Looks good, left some minor comments.

Thanks for fixing the small bugs as well!

jorg-vr (approved):

niknetniko (approved):

PR #4299, opened 2023-01-06
Show more relevant info for teachers on course cards

jorg-vr:
This pull request updates the course cards from #4261 , so that they display more relevant information for teachers.

The course progress is replaced by some generals statistics:
- the number of users
- the number of series
- the number of exercises
- the number of submissions

The content of the card displays the following info (In order of priority, with max three being shown):
- the number of unanswered questions, with a link to the questions page
- the number of pending users, with a link to review them
- the number of submissions that still need to be reviewed, with a link to the first of those submissions
- Up to 1 series with a recently passed deadline (within last week)
- Up to 2 series with an upcoming deadline 
- Up to 3 series which have recently been worked on by a the most users

The series progress bars are replaced by class progress. This visualizes the amount of users that have completed a certain amount of exercises.
This class progress is now also used for series displayed in the top cards on the homepage.

The icons for `open_for_institution` are also removed as this is the default and does not require extra attention of the course_admin.

I have also added a progress bar on the homepage when switching between academic year. (And slightly improved it's styling because the blue on pink clashed)

Visually this looks like:

Screen Shot 2023-01-20 at 16 13 46
Screen Shot 2023-01-20 at 16 13 28

- [x] Tests were added

Part of #2882


bmesuere:
I would leave out the number of questions from the header (but definitely include it in the card itself if there would be any). The other numbers  could be abbreviated to for example 15.5K instead of 15484.

The priority of things to show are:
- open questions
- pending registrations
- series with upcoming deadline
- series with passed deadline

For now, I would fill the rest of the slots with exercises (which is not really useful). In a later phase/PR, we could add some graphs instead of exercises.

For courses using series without deadlines, could we show something instead of the graphs? Maybe "series being worked on now"?

jorg-vr:
image
image
image

app/models/course.rb
```diff
@@ -188,13 +191,75 @@ class Course < ApplicationRecord
...
+                       .group(:series_id)
+                       .reorder(Arel.sql('COUNT(*) DESC'))
+                       .first
+    series = candidates.first if series.nil? && candidates.any?
+    result = series.nil? ? [] : [series]
+    result += series_being_worked_on(limit - 1, exclude + result)
+    result
+  end
+
+  def homepage_admin_notifications
```
chvp, line 228: This function is doing a lot of UI work for being in a model. Maybe it can be moved to a helper?

app/views/pages/_course_card.html.erb
```diff
@@ -1,20 +1,20 @@
 <div class="card home-summary-card course">
   <%= link_to course_path(course), class: 'card-title-link' do %>
-    <div class="card-title card-title-colored-container">
+    <div class="card-title card-title-colored-container" style="height: 111px;">
```
chvp, line 3: Is there a reason not to do this in CSS?

app/views/pages/_course_card.html.erb
```diff
@@ -66,60 +101,104 @@
...
+  <% if current_user.admin_of?(course) && course_series.length < 3 %>
+    <div class="card-supporting-text card-border <%= course_series.length == 0 ? 'course-activities-full-height' : '' %>">
+      <% admin_notifications.each do |an| %>
+        <div class="course-item">
+          <div class="course-item-icon" >
+            <i class="mdi mdi-24 <%= an[:icon] %>"></i>
           </div>
-          <div class="small text-muted ellipsis-overflow" title="<%= ca[:series].name %>">
-            <%= ca[:series].name %>
+          <div style="width: 100%">
```
chvp, line 143: Again, why not in CSS?

chvp (changes requested):

config/locales/views/pages/nl.yml
```diff
@@ -34,17 +34,52 @@ nl:
...
+        one: "Klik hier om deze vraag te bekijken"
+        other: "Klik hier om deze vragen te bekijken"
+      pending-members:
+        one: "Één gebruiker wacht op goedkeuring"
+        other: "%{count} gebruikers wachten op goedkeuring"
+      pending-members-subtitle:
+        one: "Klik hier om deze gebruiker te bekijken"
+        other: "Bekijk alle gebruikers die toegang to de cursus vragen"
+      incomplete-feedbacks:
+        one: "Één onbeoordeelde oplossing"
```
niknetniko, line 50: I think we use the term "afgewerkte oplossing", so maybe something like this?

```suggestion
        one: "Één onafgewerkte oplossing"
```

The docs also talk about "onbekeken oplossingen", so perhaps that's an alternative.

config/locales/views/pages/nl.yml
```diff
@@ -34,17 +34,52 @@ nl:
...
+        other: "%{count} onbeoordeelde oplossingen"
```
niknetniko, line 51:
```suggestion
        other: "%{count} onafgewerkte oplossingen"
```

config/locales/views/pages/nl.yml
```diff
@@ -34,17 +34,52 @@ nl:
...
+      incomplete-feedbacks-subtitle:
+        one: "Klik hier om deze oplossing te beoordelen"
+        other: "Ga verder met het beoordelen van deze oplossingen"
+      add-series: "Voeg een reeks toe"
+      add-series-subtitle:
+        zero: "Begin met het vormgeven van jouw cursus"
+        one: "Voeg meer inhoud toe aan jouw cursus"
+        other: "Voeg meer inhoud toe aan je cursus"
```
niknetniko, line 59:
```suggestion
        other: "Voeg meer inhoud toe aan jouw cursus"
```

For consistency

config/locales/views/pages/en.yml
```diff
@@ -34,17 +34,52 @@ en:
...
+        one: "Click here to view this question"
+        other: "Click here to view these questions"
+      pending-members:
+        one: "One pending member"
+        other: "%{count} pending members"
+      pending-members-subtitle:
+        one: "Click here to view this member"
+        other: "View all users that want to join this course"
+      incomplete-feedbacks:
+        one: "One unreviewed submission"
```
niknetniko, line 50: In the evaluations UI, we use the term "complete", so perhaps something like this:

```suggestion
        one: "One incomplete submission"
```

niknetniko (commented):
This looks very good! (I have added some small nitpicks here and there; an automated ERB formatter would be very nice)

One small side-effect I noticed though is that the home page now takes a bit longer to render. This is normal, but when using the academic year selector, there is no indication that something is actually happening. So for example, if I select another year, it feels like nothing is happening, which made me think something went wrong (although it was just loading). I think we should either add a loader or just go for a full page refresh (in which case the browser shows something is happening).

config/locales/views/pages/en.yml
jorg-vr: this would also require a change to the word submission. (As we are not completing the submission, but completing the process of reviewing the submission)

config/locales/views/pages/nl.yml
jorg-vr: For the same reason as in English, I think 'afgewerkt' does not apply.

I had `onbekeken` before, but it looked rather ugly

bmesuere: We hebben dit vanmorgen bespreken. Er stond eerst "onbekeken", maar dat voelde raar aan. Daarom is het aangepast naar "onbeoordeeld" (al kan " niet beoordeeld" voor mij ook). Ik was vergeten dat we vroeger "onafgewerkt" gebruikt hebben, maar ik denk dat dat in de context was dat we eerst enkel evaluaties zonder punten hadden. Voor mij werkt "beoordelen" zowel met als zonder punten geven, dus misschien moeten we die vertaling/docs dan overal aanpassen.

config/locales/views/pages/en.yml
bmesuere: I don't really agree here. In the evaluations UI, you already have the evaluation context which is why "complete" is good enough there. On the home page, you don't have that context and "unreviewed" provides that.

niknetniko (approved):

chvp (approved):

bmesuere (approved):

PR #4247, opened 2022-12-15
Show class progress on series homepage cards for teachers

jorg-vr:
This pull request replaces the users progress by the class progress on the homepage series cards, when the user is an admin of the course.

There is an icon in the top right that should make clear it is a different kind of card for teachers. I also removed the red/green color coding for teachers, as it makes less sense for this usecase.

image
image
image

- [x] Tests were added

Part of #2882

PR #4300, opened 2023-01-06
Make sure activity statuses always exist

jorg-vr:
This pull request makes sure Activity statuses are always present. This is necessary for fast computation of #4247 and #4261.

I found 2 case where Activity statuses didn't exist: After an activity was added to a series or after a series deadline was changed.
Both now start a delayed job to update the activity statuses.

This code was previously in #4247, but I split it of, as this has to be rolled out first.

After the roll out we will also have to run the following code to make sure an activity status exists for all current submissions:
```ruby
# We currently only need activity statusses within a course, making this enough
# Also activity statusses outside a course are never removed, so should exist since the introduction of the table
SeriesMembership.reorder(id: :asc).find_each(&:add_activity_statuses)

# safer but slower method to cover all cases
Submission.reorder(id: :asc).find_each { |s| s.exercise.activity_statuses_for(s.user, s.course) }
ActivityReadState.reorder(id: :asc).find_each { |s| s.activity.activity_statuses_for(s.user, s.course) }
```

- [x] Tests were added

### DIFF
--- a/app/models/series.rb
+++ b/app/models/series.rb
@@ -192,6 +192,7 @@ class Series < ApplicationRecord
 
   def invalidate_activity_statuses
     ActivityStatus.delete_by(series: self)
+    series_memberships.each(&:add_activity_statuses_delayed)
   end
 
   def invalidate_caches(user)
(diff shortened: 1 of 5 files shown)


chvp (approved):

niknetniko (approved):

jorg-vr:
Tested ```SeriesMembership.reorder(id: :asc).find_each(&:add_activity_statuses)``` on naos.

It took about 10 hours to complete.

bmesuere (approved):

PR #4342, opened 2023-01-23
Fix error messages when uploading incorrect user labels

### LINKED ISSUE #4129: Error message during CSV labels upload is not shown
The error is sent back using JS, but something seems to go wrong during its execution (or maybe it's not executed at all?).

jorg-vr:
While working on the code, I also removed jquery. (Progress on #3590 )

Closes #4129.

### DIFF
--- a/app/assets/javascripts/course.ts
+++ b/app/assets/javascripts/course.ts
@@ -2,8 +2,10 @@ import { initDragAndDrop } from "./drag_and_drop";
 import { fetch, getURLParameter } from "./util.js";
 import { ScrollSpy } from "./scrollspy";
 import { searchQuery } from "./search";
+import { html, render } from "lit";
+import { Modal } from "bootstrap";
 
-function loadUsers(_baseUrl: string, _status): void {
+function loadUsers(_status = undefined): void {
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/course.ts
```diff
@@ -78,7 +80,20 @@ function initCourseMembers(): void {
             fetch(`/courses/${modal.dataset.course_id}/members/upload_labels_csv`, {
                 method: "POST",
                 body: formData,
-            }).then(loadUsers);
+            }).then(async response => {
+                if (!response.ok) {
```
niknetniko, line 84: I would only show this specific error if we know there was no file. For example, if I try uploading something else, like an image, we get a 500 server error (maybe something for another PR), but the message is still "please upload a file".

niknetniko (changes requested):

app/assets/javascripts/course.ts
jorg-vr: Old alerts are now removed. And I added a specific check for malformed csv which should cover a lot of cases

niknetniko (approved):

bmesuere (approved):

PR #4353, opened 2023-01-25
Hide output if it is much longer than expected to avoid browser crashes

### LINKED ISSUE #4265: Too much feedback is not always properly stopped from showing up
In most cases, Dodona automatically stops printing the output of a specific submission if this becomes too much. However, this seems to be judge-specific behaviour and is not handled properly in all cases causing some browsers to crash.

See this submission link for an example of the issue: https://dodona.ugent.be/nl/submissions/12839569/

jorg-vr:
This pull request hides very long submission output to avoid browser crashes. This solves #4265 

To much output is defined as 5000  characters more than the expected output. (But this is open for discussion)

image

Closes #4265.

### DIFF
--- a/app/helpers/renderers/text_differ.rb
+++ b/app/helpers/renderers/text_differ.rb
@@ -21,11 +21,11 @@ class TextDiffer
                 end
               end
             end
-    if @generated_linecount > @expected_linecount + 100
-      @to_many_lines = @generated_linecount
-      @generated_linecount = @expected_linecount + 100
-      @generated = @generated.split("\n", -1).first(@generated_linecount).to_a.join("\n")
-    end
+    return unless @generated_linecount > @expected_linecount + 100
+
+    @to_many_lines = @generated_linecount
+    @generated_linecount = @expected_linecount + 100
+    @generated = @generated.split("\n", -1).first(@generated_linecount).to_a.join("\n")
   end
 
   def unified
--- a/app/helpers/renderers/text_differ.rb
+++ b/app/helpers/renderers/text_differ.rb
@@ -8,6 +8,12 @@ class TextDiffer
   def initialize(generated, expected)
     @generated = generated || ''
     @expected = expected || ''
+
+    if @generated.length > @expected.length + 5000
+      @too_many_characters = @generated.length
+      @generated = "#{@generated.first(@expected.length + 5000)}..."
+    end
+
     @generated_linecount = generated&.lines&.count || 0
     @expected_linecount = expected&.lines&.count || 0
     @simplified_table = @generated_linecount > 100 || @expected_linecount > 100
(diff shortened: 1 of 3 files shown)


app/helpers/renderers/text_differ.rb
```diff
@@ -21,6 +21,11 @@ def initialize(generated, expected)
                 end
               end
             end
+    return unless @generated_linecount > @expected_linecount + 100
+
+    @to_many_lines = @generated_linecount
```
chvp, line 26:
```suggestion
    @too_many_lines = @generated_linecount
```

app/helpers/renderers/text_differ.rb
```diff
@@ -47,6 +52,8 @@ def unified
         nil
       end
     end.html_safe
+    cut_off_message(builder) if @to_many_lines
```
chvp, line 55:
```suggestion
    cut_off_message(builder) if @too_many_lines
```

app/helpers/renderers/text_differ.rb
```diff
@@ -237,4 +244,10 @@ def diff_strings(generated, expected)
     exp_result += '</strong>' if in_exp_strong
     [gen_result, exp_result]
   end
+
+  def cut_off_message(builder)
+    builder.div(class: 'callout callout-warning') do
+      builder << I18n.t('submissions.show.cut_off_message', generated: @to_many_lines, count: @expected_linecount)
```
chvp, line 250:
```suggestion
      builder << I18n.t('submissions.show.cut_off_message', generated: @too_many_lines, count: @expected_linecount)
```

app/helpers/renderers/text_differ.rb
```diff
@@ -21,6 +21,11 @@ def initialize(generated, expected)
                 end
               end
             end
+    return unless @generated_linecount > @expected_linecount + 100
```
chvp, line 24: A return in the constructor feels weird to me.

chvp (changes requested):
I think we should tackle lines that are far too long as well. In the case of #4265, we could imagine a student not adding a newline character. This would then result in one huge line, and the same problem would occur. Cutting off the output if there are 2000 (or something like that) characters more than expected would solve both cases I think.

The interaction with the simplified_table functionality also feels weird to me somehow. In the case where there is such a difference between generated and expected, I would also expect to also use the simplified table, but that's not very clear from the code.

app/helpers/renderers/text_differ.rb
jorg-vr: The constructor is also just a function...

This was also forced by our linter (Which prefers guard clauses over a regular if)

chvp (approved):

bmesuere (approved):

niknetniko (approved):

PR #4373, opened 2023-01-30
Limit the number of notifications requests

### LINKED ISSUE #3576: Tweak notification fetch frequency
Right now, we check for new notifications every 60 seconds even if the tab is in the background. This can be improved by only checking if the tab is visible and maybe adding some kind of backoff. The good new is that we have already implemented such logic in the `auto_reload.ts` file.

The reason for this is a substantial server load by the notification controller.
Last week, the notifications accounted for a third of the cpu time:
image

And 2/3 of the number of hits:
image

jorg-vr:
This pull request limits the numbers of notifications requests in two ways;
1. Notifications are only updated automatically if a user has open questions
2. Notifications get updated less often if a tab is inactive

A new gem [counter_culture](https://github.com/magnusvk/counter_culture) was added that introduces conditional counter cache. This is used to cache the number of open questions for a user.

This should be run manually in the console, to initialize counter cache:
```
User.find_each do |u|
   u.update(open_questions_count: u.questions.where.not(question_state: :answered).count)
 end
 ```

- [x] Tests were added

Closes #3576

### CODE REVIEW

File: app/assets/javascripts/code_listing/question_annotation.ts [hunk 1]
```diff
@@ -154,6 +155,9 @@ export async function createUserAnnotation(formData: UserAnnotationFormData,
     const data = await response.json();
 
     if (response.ok) {
+        if (mode == "question") {
```
bmesuere, line 158:
```suggestion
        if (mode === "question") {
```

app/assets/javascripts/notification.ts
```diff
@@ -83,8 +84,17 @@ export class Notification {
...
+        60 * 1000,
+        async () => {
+            const response = await fetch("/notifications.js?per_page=5");
+            eval(await response.text());
+        },
+        60 * 1000
+    ));
+
+    static checkNotifications(): void {
+        Notification.checkNotificationsTimeout.then( t => t.start());
```
bmesuere, line 98: This is hard to read and understand

bmesuere (changes requested):

bmesuere (approved):
This is a lot more readable! 👍

niknetniko (approved):

PR #4374, opened 2023-01-31
Rename application_old and only include where relevant

### LINKED ISSUE #4360: Rename application_old.js
application_old.js now only contains ACE. We can thus rename the file (and update the sprockets configuration) and only include it on the exercise page.

jorg-vr:
This pull request renames `application_old.js` to `ace_editor.js` as this is the only legacy code still included. 
It is also no longer loaded on every page, as the code is only relevant on the detail page of an exercise, where the code editor is actually shown.

Closes #4360.


app/views/activities/show.html.erb
```diff
@@ -2,6 +2,7 @@
   <%= javascript_include_tag 'exercise' %>
   <% if @activity.exercise? %>
     <%= javascript_include_tag 'submission' %>
+    <%= javascript_include_tag 'ace_editor' %>
```
bmesuere, line 5: I'm not sure and it's hard to test, but exercise.ts uses the `ace` global which I think is defined in the ace editor file. Because the order of inclusion changed (app_old was included as the first js file) this might break in some cases. To be sure, I think it's better to add this include first here.

bmesuere (changes requested):

bmesuere (approved):

chvp (approved):

PR #4378, opened 2023-02-02
Cache submission annotation count to avoid many slow queries

jorg-vr:
This pull request caches the released_annotation_count for every submission in the database. The goal is to avoid a lot of queries that check if a submission is annotated.

This query was third in our server time usage:
image
The cause was the check if a submission has annotations on every page were submissions are listed. (And for example on the activity page, this is loaded every time a user submits)

I also updated the `released` scope to also include `left_joins(:evaluations)` as this was a mandatory addition to get this scope working and thus always included when this scope was used.

I also had to trigger an update of counter culture manually from evaluations, as this use case was not covered.

To avoid slow migrations, the following should be run on production afterwards:
```ruby
Annotation.counter_culture_fix_counts
```

Until this is run, no `has_annotations` announcements will be show, but this seems too minor an issue to require separate rollout

- [x] Tests were added

### DIFF
--- a/app/models/annotation.rb
+++ b/app/models/annotation.rb
@@ -56,7 +56,6 @@ class Annotation < ApplicationRecord
                   column_name: proc { |annotation| annotation.released? ? 'released_annotation_count' : nil },
                   column_names: { Annotation.released => 'released_annotation_count' }
 
-
   def released?
     evaluation.blank? || evaluation.released
   end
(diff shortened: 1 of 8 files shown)


bmesuere (commented):
Do we need the actual counts? If not, a boolean would do. Since this is by far our biggest table, this can make a big difference.

jorg-vr:
I'll make a new pr with a boolean

jorg-vr:
replaced by #4379

PR #4379, opened 2023-02-02
Cache submission `annotated?` as boolean

jorg-vr:
This pull request caches `annotated?` for every submission in the database. The goal is to avoid a lot of queries that check if a submission is annotated.

This query was third in our server time usage:
image
The cause was the check if a submission has annotations on every page were submissions are listed. (And for example on the activity page, this is loaded every time a user submits)

I also updated the `released` scope to also include `left_joins(:evaluations)` as this was a mandatory addition to get this scope working and thus always included when this scope was used.

I also used the new boolean to check whether annotations should be fetched on submission show.

I left the migration query in this time as it took only 52002.3ms on Naos and the impact is larger then in #4378 because we also use it to determine if annotations should be fetched.

- [x] Tests were added

replaces #4378


chvp (approved):

bmesuere (approved):
Looks good! I'm not sure about the migration query because a huge where clause with id's will be generated. I would iterate in batches over the annotations and update per batch. But if it works on naos, it will probably be fine.

jorg-vr:
Tested on naos. It does not generate a where clause with the ids. Rails is smart enough to generate a subquery, so  everything happens in sql.

Issue #4362, opened 2023-01-27
Tracking: Repository fixes


bmesuere:
```[tasklist]
- [ ] #4380 
- [ ] #4381 
- [ ] #4382
- [ ] #3430 
- [x] The placeholder for the clone url on the repositories/new page mentions github.ugent.be. While correct, github.com would be more logical. From #4240
- [x] When adding a repo, add a better error message when permissions are missing. From #4240
- [x] The learning activities table (at least the one on the repo page) has no empty state. From #4240
- [x] When adding a repository, the existing exercises were not added automatically. (needs to be double checked) From #4240
- [ ] https://github.com/dodona-edu/dodona/issues/4402
- [ ] https://github.com/dodona-edu/dodona-edu.github.io/pull/231
- [x] add link of judge overview to repo page
```

bmesuere:
About the judge selection: I don't think we can really solve this right now. Instead, I suggest we add a page to the documentation website containing an overview of all judges with a 1 paragraph description (and a link to more in depth docs, example exercises,...)

We can link to that page from the form on Dodona.

chvp:
Can't we solve this by introducing an n*n relationship between programming languages and judges, with an optional comment that can be shown? (Relying on manual data entry.)

bmesuere:
Once we have tested up and running for single programming languages such as c# and Python, I would indeed add them as a separate "fake" option to the selection.

PR #4382, opened 2023-02-03
Limit the amount of users exposed by the repository admin screen

jorg-vr:
This pull request stops the repository admin sceen from leaking all users.

All users can still be found by typing a specific enough query, but the full list is no longer loaded by default. This is a compromise between usability and privacy.

Results are only returned if at least three characters are used as filter and at most 10 results are shown at any time

image
image
image
image

Part of #4362

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -17,7 +17,7 @@ class UsersController < ApplicationController
 
     @too_many_results = @users.count > 10 || params[:filter].nil? || params[:filter].length < 3
     @users = @users.none if params[:filter].nil? || params[:filter].length < 3
-    @users = @users.first(10)
+    @users = @users.order_by_name_length(:asc).first(10)
 
     respond_to do |format|
       format.html { redirect_to @repository }
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -12,11 +12,16 @@ class UsersController < ApplicationController
     @title = I18n.t('users.index.title')
   end
 
+  MAXIMUM_USERS_AFR = 10
+  MINIMUM_FILTER_LENGTH_AFR = 3
+
   def available_for_repository
     @repository = Repository.find(params[:repository_id]) if params[:repository_id]
 
-    @too_many_results = @users.count > 10 || params[:filter].nil? || params[:filter].length < 3
-    @users = @users.none if params[:filter].nil? || params[:filter].length < 3
+    @imprecise_filter = params[:filter].nil? || params[:filter].length < MINIMUM_FILTER_LENGTH_AFR
+    @too_many_results = @users.count > MAXIMUM_USERS_AFR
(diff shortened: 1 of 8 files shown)


niknetniko (approved):

app/controllers/users_controller.rb
```diff
@@ -14,6 +14,11 @@ def index
 
   def available_for_repository
     @repository = Repository.find(params[:repository_id]) if params[:repository_id]
+
+    @too_many_results = @users.count > 10 || params[:filter].nil? || params[:filter].length < 3
+    @users = @users.none if params[:filter].nil? || params[:filter].length < 3
+    @users = @users.order_by_name_length(:asc).first(10)
```
bmesuere, line 20: the `params[:filter].nil? || params[:filter].length < 3` condition is duplicated and the magic number (10) is used twice.

bmesuere (changes requested):
- I think we want a different empty state message when the page is loaded. It now says "Er werden te veel gebruikers gevonden. Gebruik de zoekbalk om de resultaten te filteren." but I didn't do anything yet. Something like "Enter at least x characters to search for users by first or last name" seems appropriate.
- If the list is truncated, "Er werden te veel gebruikers gevonden" seems strange. I would say explicitly that only the first few results are shown. Maybe also something like "Can't find who you're looking for? Try searching on their full name or filter by institution"
- The sort by length seems a bit strange/random on naos. I think we can just sort alphabetically now that we have an institution filter?

config/locales/views/users/en.yml
```diff
@@ -48,3 +48,6 @@ en:
       no_institution: This user has no institution
     repository_users_table:
       no_institution: "No institution"
+      too_many_results: "Only the first 10 results are shown. Can't find who you're looking for? Try searching on their full name or filter by institution."
```
bmesuere, line 51:
```suggestion
      too_many_results: "Only the first 10 results are shown. Can't find who you're looking for? Try searching for their full name or filter by institution."
```

db/schema.rb
```diff
@@ -104,8 +104,8 @@
     t.integer "series_id_non_nil", null: false
     t.index ["accepted", "user_id", "series_id"], name: "index_activity_statuses_on_accepted_and_user_id_and_series_id"
     t.index ["activity_id"], name: "index_activity_statuses_on_activity_id"
+    t.index ["series_id", "started", "user_id", "last_submission_id"], name: "index_as_on_series_and_started_and_user_and_last_submission"
     t.index ["series_id"], name: "fk_rails_1bc42c2178"
-    t.index ["started", "user_id", "last_submission_id"], name: "index_as_on_started_and_user_and_last_submission"
     t.index ["started", "user_id", "series_id"], name: "index_activity_statuses_on_started_and_user_id_and_series_id"
```
bmesuere, line 109: This was probably included by accident?

bmesuere (approved):

niknetniko (approved):

PR #4383, opened 2023-02-03
Fix multiple repository related paper cuts

jorg-vr:
This pull request fixes multiple paper cuts related to repositories.

- [x] The placeholder for the clone url on the repositories/new page mentions github.ugent.be. While correct, github.com would be more logical. From #4240
- [x] When adding a repo, add a better error message when permissions are missing. From #4240
New message:
image

- [x] The learning activities table (at least the one on the repo page) has no empty state. From #4240
New empty state:
image

- [x] When adding a repository, the existing exercises were not added automatically. From #4240
This issue was only present in staging and production, where the git worker queue allowed the tasks to be done simultaneously/out of order. This should now be forced to consecutive execution.

Part of #4362 .

### DIFF
--- a/config/locales/models/en.yml
+++ b/config/locales/models/en.yml
@@ -10,6 +10,9 @@ en:
           attributes:
             activity:
               taken: You have already read this activity
+        gitable:
+          repository:
+            not_accessible: The repository is not accessible for our git user. Please double check if you have given the correct permissions. Take notice that it might take a while before a github/gitlab invitation is accepted by our team.
     models:
       course: Course
       activity: Activity
--- a/config/locales/models/en.yml
+++ b/config/locales/models/en.yml
@@ -12,7 +12,7 @@ en:
               taken: You have already read this activity
         gitable:
           repository:
-            not_accessible: The repository is not accessible for our git user. Please double check if you have given the correct permissions. Take notice that it might take a while before a github/gitlab invitation is accepted by our team.
+            not_accessible: The repository is not accessible for our git user. Please double check if you have given the correct permissions. Note that it might take a while before a GitHub/GitLab invitation is accepted by our team.
     models:
       course: Course
       activity: Activity
--- a/config/locales/models/en.yml
+++ b/config/locales/models/en.yml
@@ -12,7 +12,7 @@ en:
               taken: You have already read this activity
         gitable:
           repository:
-            not_accessible: The repository is not accessible for our git user. Please double check if you have given the correct permissions. Note that it might take a while before a GitHub/GitLab invitation is accepted by our team.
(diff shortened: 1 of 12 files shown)


config/locales/models/en.yml
```diff
@@ -10,6 +10,9 @@ en:
           attributes:
             activity:
               taken: You have already read this activity
+        gitable:
+          repository:
+            not_accessible: The repository is not accessible for our git user. Please double check if you have given the correct permissions. Take notice that it might take a while before a github/gitlab invitation is accepted by our team.
```
niknetniko, line 15:
```suggestion
            not_accessible: The repository is not accessible for our git user. Please double check if you have given the correct permissions. Note that it might take a while before a GitHub/GitLab invitation is accepted by our team.
```

config/locales/models/nl.yml
```diff
@@ -10,6 +10,9 @@ nl:
           attributes:
             activity:
               taken: Je hebt deze activiteit al gelezen
+        gitable:
+          repository:
+            not_accessible: De repository is niet toegankelijk voor onze git-gebruiker. Controleer of je de juiste permissies hebt ingesteld. Hou er rekening mee dat het even kan duren voor een github/gitlab uitnodiging wordt geaccepteerd door ons team.
```
niknetniko, line 15:
```suggestion
            not_accessible: De repository is niet toegankelijk voor onze git-gebruiker. Controleer of je de juiste permissies hebt ingesteld. Hou er rekening mee dat het even kan duren eer een uitnodiging voor GitHub/GitLab wordt geaccepteerd door ons team.
```

niknetniko (approved):

config/locales/models/en.yml
```diff
@@ -10,6 +10,9 @@ en:
           attributes:
             activity:
               taken: You have already read this activity
+        gitable:
+          repository:
+            not_accessible: The repository is not accessible for our git user. Please double check if you have given the correct permissions. Note that it might take a while before a GitHub/GitLab invitation is accepted by our team.
```
bmesuere, line 15:
```suggestion
            not_accessible: The repository is not accessible to our git user. Please double check that you have given the correct permissions. Note that it may take a while before a GitHub/GitLab invitation is accepted by our team.
```

Can we include a link to the docs here that explains the github/gitlab user thing?

bmesuere (changes requested):
I don't like the email solution either. Can't we just send an email to the address associated with the repo admin in this case? Since the repo has just been added by that user, using that address seems like a good fallback.

jorg-vr:
the repo admin is only added after the repo is saved. 
Because cloning the repo is a delayed task, and it also takes some time to perform, this admin will almost always be added before the start of `process activities`.
However, this order of execution is not at all guaranteed and could result into bugs in some cases. (Most notably it causes this bug in our test environment where  deleyad_jobs are executed in order)

This was my first solution which I reverted because of a failing test. Do you want to continue with this approach? (In which case I would just remove the test because we know it will fail)

bmesuere:
In that case, maybe add dodona@ugent.be as ultimate fallback?

bmesuere (approved):

PR #4404, opened 2023-02-10
Allow responding to annotations in a thread

### LINKED ISSUE #2827: Thread questions and annotations
From https://github.com/dodona-edu/dodona/issues/1852#issue-599457917

jorg-vr:
This pull request adds threading to annotations. This allows users to respond to a specific annotation.
In order to add this functionality in the frontend, a major refactor of the frontend implementation of annotation is also contained in this pr.

I will first go over the functional changes from a user perspective. I will list the technical changes later.

## Functional changes

First of all, the visuals of a user annotation have changed a bit:
old:
image
new:
image

The options to edit, delete or save an annotation for reuse are al moved to less intrusive dropdown button:
image

Clicking on the new answer input field opens up the existing larger form, that allows you to create a response on this annotation:
image

An open question inside a thread is clearly marked as such using an icon:
image

When you click the answer form, the question will be marked as in progress for one hour:
image

Adding an answer to a question will automatically resolve it:
image

An open question can also be resolved by clicking 'mark as answered', which will appear on each thread with an open question:
image

It is no longer possible to change the question status back to 'unanswered' using this interface. If desired I could add this option underneath dropdown.

### Backend
The backend changes are rather minimal. An annotation now keeps track of it's `thread_root_id`.
If no thread root is defined an annotation is assumed to be the thread root.
Annotations with a thread root id are considered responses to that root annotation.

Only one level of threading is allowed. Thus a response to an annotation cannot be the thread root of another annotation. Validations and tests for this case are included.

### Frontend
All annotation related code has been refactored to webcomponents and the state event system.

The webcomponents contain all code related to visualizing a component.

The state contains all data displayed on the page. The state is responsible for making requests to the backend, and notifying webcomponents if the data they are using has changed.

Splitting up these responsibilities should lead to more readable and more maintainable code.

As the amount of webcomponents increased a lot, I also decided to move some of them into more dedicated folders.

- [x] Tests were added
- [ ] Documentation update can be found at dodona-edu/dodona-edu.github.io# (will do this after approval)

Closes #2827.


tsconfig.json
```diff
@@ -4,7 +4,7 @@
     "emitDecoratorMetadata": true,
     "esModuleInterop": true,
     "experimentalDecorators": true,
-    "module": "es6",
+    "module": "es2020",
```
niknetniko, line 7: Is this on purpose (just checking to make sure)?

niknetniko (commented):
Looks good and works fine for me.

A few small comments:
- The two icons don't seem aligned to me:
image
- These don't seem to have the same size for me:
image
- The "hover" bubbles on the toggles seem to have both the bootstrap hover in English, and also the built-in one in Dutch:
image

I am not 100% convinced hiding the options to edit, delete or save an annotation, as this reduces the number of buttons from 2 to 1 in most cases. Would it still be too cluttered if those buttons were also muted?

tsconfig.json
jorg-vr: this was intentional, because I used non es6 features to fix tests
But in the end I solved it in another way, so it's no longer required

jorg-vr:
> * The two icons don't seem aligned to me:
>       image

I agree that it does not look ideal, but this is due to different design choices of the mdi svg's: The circle and the rectangle are both center aligned. The bottom part of the comment icon makes it look misaligned. As these are rendered inline this is not easily fixed by css (And fixing it would require some very specific css margins). Unless others also find this important, I would leave it be

>  I am not 100% convinced hiding the options to edit, delete or save an annotation, as this reduces the number of buttons from 2 to 1 in most cases. Would it still be too cluttered if those buttons were also muted?

Three icons will be default for staff once the save button is rolled out. And I prefer this placement for the delete button (compared to inside the edit form or directly on the annotation)

niknetniko:
> As these are rendered inline this is not easily fixed by css (And fixing it would require some very specific css margins). Unless others also find this important, I would leave it be

Agree.

> Three icons will be default for staff once the save button is rolled out. And I prefer this placement for the delete button (compared to inside the edit form or directly on the annotation)

:+1:

app/assets/javascripts/components/annotations/annotations_toggles.ts
```diff
@@ -0,0 +1,62 @@
...
+
+    get annotationVisibility(): AnnotationVisibilityOptions {
+        return getAnnotationVisibility();
+    }
+
+    protected render(): TemplateResult {
+        return html`
+            <span class="diff-switch-buttons switch-buttons">
+                <span>${I18n.t("js.annotations.toggles.title")}</span>
+                <div class="btn-group btn-toggle" role="group" aria-label="Annotaties" data-bs-toggle="buttons">
```
chvp, line 32: Untranslated string.

app/assets/javascripts/components/annotations/user_annotation.ts
```diff
@@ -0,0 +1,168 @@
...
+                                    </a>
+                                </li>
+                                ${ this.data.permission.destroy ? html`
+                                    <li>
+                                        <a class="dropdown-item" @click="${() => this.deleteAnnotation()}">
+                                            <i class="mdi mdi-delete mdi-18"></i> ${I18n.t(`js.user_annotation.delete`)}
+                                        </a>
+                                    </li>
+                                ` : ""}
+                                <d-new-saved-annotation
```
chvp, line 142: I would move this to the top of the dropdown.

app/models/annotation.rb
```diff
@@ -82,4 +91,14 @@ def annotate_submission
   def reset_submission_annotated
     submission.update(annotated: false) unless submission.annotations.released.any?
   end
+
+  def previous_annotation
+    return nil if thread_root.nil?
+
+    [thread_root, thread_root&.responses&.where('created_at < ?', created_at)].flatten.last
```
chvp, line 98:
```suggestion
    [thread_root, thread_root&.responses&.where(created_at: ..created_at)].flatten.last
```

app/models/question.rb
```diff
@@ -52,4 +54,16 @@ def clear_transition
     @transition_to = nil
     @transition_from = nil
   end
+
+  def reset_in_progress
+    return unless question_state == 'in_progress'
```
chvp, line 59: Is this tested? You have to be careful whether the deserialized object from the database is updated before this is executed (to be clear: I'm not sure whether this is the case).

app/models/annotation.rb
jorg-vr: Thanks for teaching me this notation :)

I looked it up, to better understand it: https://stackoverflow.com/a/23936233/7769243

It actually needs three dots (`.where(created_at: ...created_at)`) to correctly represent a `<`, your answer would result in a `<=` (Which would result in returning itself instead of the previous annotation)

app/models/question.rb
jorg-vr: I have added extra tests for this (it works)

niknetniko (approved):

chvp (approved):

bmesuere (changes requested):
(I know not all remarks are related to changes made in this PR)
---
image

- [x] The text below is a bit strange. `[Onze handleiding]() bevat meer informatie over het stellen van vragen`. The manual should also be updated to include threading.

---

image

- [ ] After asking a question, this is the result. My first reaction was that "Beantwoorden..." was a bit strange here because I just asked a question. Maybe it should say "Zelf beantwoorden..." in this case.

---
image

- [x] If I clicked "als beantwoord markeren", this was the result which is not what I expected.
I would expect some kind of message below the question like "This question was marked as answered by xxx on xxx" and an explicit option to reopen de question.

If a question is answered by a teacher, the same UI is present for the student. I understand that from an internal perspective the question is "closed", but for a student the UI also feels a bit strange here. A teacher just answered your question and your option is "beantwoorden"

---

Question overview: the icons are a bit cryptic, so maybe use a bootstrap tooltip instead of the title tag? But maybe we can do better. The primary action by a large margin should now be to answer the question (and thus open the submission), so maybe a "split" text button where the primary action is answer and the other options are in the dropdown, or the combination of a text button and a "triple dot" overflow menu...

app/assets/javascripts/code_listing.ts
```diff
@@ -0,0 +1,75 @@
+import { CodeListingRow } from "components/annotations/code_listing_row";
```
jorg-vr, line 1: test

jorg-vr:
@bmesuere 

> After asking a question, this is the result. My first reaction was that "Beantwoorden..." was a bit strange here because I just asked a question. Maybe it should say "Zelf beantwoorden..." in this case.

I am not a fan of `zelf beantwoorden`, it gives no extra clarity compared to `beantwoorden`, where the actor is clearly implied. I would be willing to change it to the imparitive `beatwoord... of markeer als beantwoord`

I thought about using another word, but reply really is the default actions you take in a communication thread, event if the thread start is your own:
image

app/assets/javascripts/state/UserAnnotations.ts
```diff
@@ -284,3 +286,11 @@ export async function transition(annotation: UserAnnotationData, newState: Quest
         invalidateUserAnnotation(annotation.id);
     }
 }
+
+export async function transitionAll(annotations: UserAnnotationData[], newState: QuestionState): Promise<void> {
+    for (const annotation of annotations) {
+        // we wait for each transition to finish before starting the next one
+        // this prevents inconsistencies questionstates being shown
```
chvp, line 293:
```suggestion
        // this prevents inconsistent questionstates being shown
```

chvp (approved):

PR #4405, opened 2023-02-12
Remove Indianio support

### LINKED ISSUE #2919: Remove Indianio support
There are currently a few open issues involving streamlining the format of the several exports we offer. Before tackling those, it would make sense to investigate if Indianio support can be removed. This would simplify the code and thus the changes needed for a more uniform export.

Indianio support is disabled by default and can only be enabled for courses associated with Ghent University. This means we can easily track the use of this feature. In 2020-2021, only a single course enabled Indianio support. We will discuss the need for this feature with the teacher of this course.

bmesuere:
This pull request removes Indianio support since the service was taken down.

Closes #2919


bmesuere:
@jorg-vr can you add a migration and remove the token from the database?

I tried to generate a migration itself, but puma crashed due to a ssl3 lib error. I then tried to reinstall Ruby, but that failed and now I have no working ruby version on my system :(

jorg-vr (approved):

niknetniko (approved):

PR #4415, opened 2023-02-15
Fix overlap in question table

### LINKED ISSUE #4190: Question table is smushed on mobile

freyavs:
This pull request fixes the overlapping columns in the question table and the table now becomes scrollable when the screen shrinks.

Closes #4190

### DIFF
--- a/app/assets/stylesheets/models/questions.css.scss
+++ b/app/assets/stylesheets/models/questions.css.scss
@@ -4,7 +4,7 @@
   }
 
   col.question {
-    width: 150px;
+    min-width: 150px;
   }
 
   col.age {
--- a/app/assets/stylesheets/models/questions.css.scss
+++ b/app/assets/stylesheets/models/questions.css.scss
@@ -3,6 +3,10 @@
     width: 100px;
   }
 
+  col.question {
+    width: 150px;
+  }
+
   col.age {
     width: 240px;
   }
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/models/questions.css.scss
```diff
@@ -3,6 +3,10 @@
     width: 100px;
   }
 
+  col.question {
+    width: 150px;
```
jorg-vr, line 7: Does this column still grow larger on wide screens? Seeing a preview of the question is quite useful if the screen width allows it

freyavs: It does, though it wouldn't hurt to make that column a little bit wider maybe.

jorg-vr: image
In current implementation it takes all available width if possible.

I would prefer if we could keep something like this, instead of a fixed number of pixels.

We often use 33vw as width to allow columns to become rather large, but still be reasonable on phones

jorg-vr: But the question in the test data are very short, maybe it is already fixed with longer questions?

jorg-vr: min-width sounds like a good solution

jorg-vr (approved):

bmesuere (approved):

PR #4418, opened 2023-02-16
Fix comma's at the end of activity labels 

### LINKED ISSUE #4220: Activity labels always end on a `,`
image

_Originally posted by @bmesuere in https://github.com/dodona-edu/dodona/pull/4203#pullrequestreview-1204696037_

freyavs:
This pull request fixes the dynamic comma's in the activity labels list (e.g. in the exercices table).

Closes #4220

### DIFF
--- a/app/views/activities/_labels.html.erb
+++ b/app/views/activities/_labels.html.erb
@@ -16,9 +16,11 @@
         <% end %>
         ...</span>
     <% elsif labels.any? %>
-      <%labels.each_with_index do |label, index| %><%#
-        %><d-filter-button param="labels" value="<%= label.name %>" multi><%= index == 0 ? "" : ", " %><%=label.name%></d-filter-button><%#
-      %><% end %>
+      <% labels.each_with_index do |label, index| %>
+        <d-filter-button param="labels" value="<%= label.name %>" multi>
+          <%= label.name %><%= "," unless index == labels.count - 1 %>
+        </d-filter-button>
+      <% end %>
     <% end %>
   </small>
 </div>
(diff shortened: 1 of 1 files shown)


bmesuere (changes requested):
We try to avoid `raw`, especially with input coming from the user because of injection attacks.

In this case, I would use the original code, but replace the `each` with a `each_with_index `. You can then put a comma in front of the label name, except when the index is zero.

jorg-vr (approved):

app/views/activities/_labels.html.erb
```diff
@@ -16,11 +16,9 @@
...
         ...</span>
     <% elsif labels.any? %>
-      <% labels.each do |label| %>
-        <d-filter-button param="labels" value="<%= label.name %>" multi>
-          <%= label.name %>,
-        </d-filter-button>
-      <% end %>
+      <%labels.each_with_index do |label, index| %><%#
+        %><d-filter-button param="labels" value="<%= label.name %>" multi><%= index == 0 ? "" : ", " %><%=label.name%></d-filter-button><%#
+      %><% end %>
```
niknetniko, line 21: Couldn't this be solved using something like this? To prevent the ugly comments hack

```suggestion
      <% labels.each_with_index do |label, index| %>
        <d-filter-button param="labels" value="<%= label.name %>" multi>
          <%= label.name %><%= index == labels.count - 1 ? "" : "," %>
        </d-filter-button>
      <% end %>
```

niknetniko (changes requested):

app/views/activities/_labels.html.erb
freyavs: Yes actually that was the other way to avoid the problem, didn't know which one was cleaner. I'll apply this

bmesuere (approved):

niknetniko (approved):

PR #4419, opened 2023-02-17
Only show icons if course is closed or hidden on home and course page

### LINKED ISSUE #4399: Update registration state icon on course page
When the course cards were redesigned, we removed the icon when registrations were only open for a single institution because that is the default setting.

On the course page, however, we still show that icon:
image

jorg-vr:
This pull request moves the visibility icon logic to a separate file, to make sure it is the same everywhere.

It is also changed functionally to only show icons when a course is closed or hidden.

I am unsure if this was actually the goal of  #4399 as this issue mentions not showing an icon for "open for institution" and also defends this by saying this is the default. But the current default is actually "open_for_institutional users".

This confusion was probably caused by me confusing those two states in the homepage redesign.

I also removed an unused view, which I encountered because it triggered extra icons on the course cards.

image
image

Closes #4399 .


app/views/courses/show.html.erb
```diff
@@ -12,21 +12,11 @@
...
-        <% end %>
-        <% unless @course.visible_for_all? %>
-          <h2 class='card-title-text float-end' title='<%= t ".visibility-#{@course.visibility}-info", institution: @course.institution&.name %>'>
-            <i class="mdi mdi-eye-off-outline"></i>
-          </h2>
-        <% end %>
-        <h2 class="card-title-text">
+        <h2 class="card-title-text justify-content-between">
           <%= @course.name %> (<%= @course.formatted_year %>)
+            <span class="float-end text-nowrap">
```
chvp, line 17:
```suggestion
          <span class="float-end text-nowrap">
```

app/views/courses/_grouped_course_cards.html.erb
```diff
@@ -1,16 +0,0 @@
-<% grouped_courses.each do |year, courses| %>
```
chvp, line 1: This file was unused before? It doesn't seem to be referenced in the rest of the PR.

chvp (approved):

app/views/courses/_grouped_course_cards.html.erb
jorg-vr: see
> I also removed an unused view, which I encountered because it triggered extra icons on the course cards.

But yes, I couldn't find any mentions to it, doing a text search in the project.
If you know of a page other then the home page that uses course cards that I should check, let me know.

bmesuere (commented):
Since almost all code in the new partial is wrapped in `<% %>`, isn't this an indication that converting this to a helper instead of partial is a better option?

I'm also a bit confused by the first commit in this PR. It mentions a revert of a previous commit, but I can't find that.

jorg-vr:
I'll replace it by a helper.

I accidentally committed and pushed to develop first, by mistake. I immediately reverted this commit on develop,  created a new branch and then reverted the revert commit, to reintroduce the changes on this new branch

app/helpers/courses_helper.rb
```diff
@@ -33,4 +33,17 @@ def registration_action_for(**args)
...
+
+  def visibility_icons_for(course)
+    icons = []
+    if course.closed? || course.moderated
+      title = []
+      title.push t("courses.show.registration-#{course.registration}-info", institution: course.institution&.name) if course.hidden?
+      title.push t('courses.show.moderated-info') if course.moderated
+      icons.push tag.i class: 'mdi mdi-account-remove-outline', title: title.join("\n")
+    end
+    icons.push ' ' if course.hidden? && (course.closed? || course.moderated)
```
bmesuere, line 45: why is an empty string pushed here?

jorg-vr: styling was broken otherwise. In the template there was some white space between the icons because they were on a different line.

Could also be fixed by css if you prefer

chvp: Can't this be fixed by using `.join` instead of `.reduce` on the last line?

chvp (approved):

bmesuere (approved):

app/views/courses/_grouped_course_cards.html.erb
jorg-vr: test

jorg-vr: test

PR #4431, opened 2023-02-26
Ask for confirmation when making series visible

### LINKED ISSUE #4243: Ask confirmation when making series visible from main course page
Making a series visible using the link on the main course is only one click away from inadvertently publishing a series that should be kept hidden/closed. Asking confirmation before performing the action might prevent unwanted actions (but no guarantee).

image

freyavs:
This pull request makes sure the user/admin is asked for confirmation when using the link on the main course for changing the visibility of the series.

There is a slight change, before:

Now the link is completely bold:

Closes #4243

### DIFF
--- a/app/views/series/_series.html.erb
+++ b/app/views/series/_series.html.erb
@@ -139,19 +139,19 @@
   <% if current_user&.course_admin?(series.course) && series.hidden? %>
     <div class="alert alert-info hidden-print">
       <%= t "series.show.series_not_visible" %>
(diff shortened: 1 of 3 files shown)


freyavs:
Another option, to keep the original look, would be to do the following, but seems less clean (unless there's another way in ruby):

bmesuere:
I'm ok with using the entire sentence as link

jorg-vr (approved):

bmesuere (approved):

PR #4462, opened 2023-03-09
Add tests for search components

jorg-vr:
This pull request adds tests for multiple search components.

All these components now also take a `searchQuery` object as property. This makes it easier to test without a global searchquery state involved. It improves the reusability of these components, as this allows multiple searchcomponents belonging to  a different query on the same page.


niknetniko (approved):

bmesuere (approved):
👍

PR #4464, opened 2023-03-11
Make submission page wider (on smaller screens)

### LINKED ISSUE #4191: Submission page margin is too wide on smaller screens
When my screen becomes less wide than 1400px, Dodona shrinks the page width. However, this results in relatively wasteful margins; the content becomes smaller than needed.

For comparison, the course page with the same browser width is attached as the second screenshot.

image

image

freyavs:
This pull request makes the submission page a bit wider. More specifically, when the screen becomes smaller than 1400px, the margins shouldn't be too big. The width is now more in line with e.g. the course page (also for screen sizes bigger than 1400px, see below).

For screen size <1400px  (comparison with the course page):

For screens bigger than 1400px the submission page is also slightly wider now (let me know if you'd like to keep it as it was before):

**Before:**

 **After**:

Closes #4191

### DIFF
--- a/app/views/submissions/show.html.erb
+++ b/app/views/submissions/show.html.erb
@@ -1,6 +1,6 @@
 <%= render 'navbar_links' %>
 <div class="row justify-content-center">
-  <div class="col-lg-8 col-xxl-9 col-12">
+  <div class="col-md-10 col-12">
     <div class="card">
       <div class="card-title card-title-colored">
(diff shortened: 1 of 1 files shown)


jorg-vr (changes requested):

jorg-vr:
I would put this on hold, as I will remove that sidecard soon, which will solve this issue

PR #4485, opened 2023-03-16
Preload the latest submission in the editor on the activity show page

### LINKED ISSUE #2886: Easy fetch last submission in hand in area

jorg-vr:
This pull request preloads the latest submission in the code editor. When this happens, an allert is also edit that explains what happened and allows the user to restore the boilerplate.

image
image

- [x] tests are added

Closes #2886


bmesuere (changes requested):
Restoring now happens with a reload and a query string. It would be nicer for the user to do this with JS without reload.

bmesuere (changes requested):
If you impersonate me on naos, You'll see an error on https://naos.dodona.be/nl/courses/2026/activities/1944243494/

The console logs
> caught SyntaxError: missing ) after argument list

bmesuere (approved):

chvp (approved):

PR #4487, opened 2023-03-16
Replace most_recent_correct by most_resent submissions filter

### LINKED ISSUE #1441: Add option to rejudge latest submissions
Right now there's a filter option to rejudge only the latest _correct_ submission for students, but I think it would be very useful if it's possible to rejudge the latest submission (regardless of it being correct or not).

jorg-vr:
This pull request removes `most_recent_correct_per_user` scope from submissions and replaces it by `most_recent_per_user`.

Due to the fixed order of filter execution, the old functionality can still be achieved by filtering both on `most_recent_per_user` and the status `correct`.

image

Order of filter execution could be hard to interpret as an end user. If anyone has any suggestions how to explain this to the user in the UI, please give them.

As it is a rather hidden/advanced option anyway, I wouldn't mind shipping it as is without further clarification.

Closes #1441 .

### DIFF
--- a/app/views/submissions/index.html.erb
+++ b/app/views/submissions/index.html.erb
@@ -38,7 +38,7 @@
         <% if current_user&.course_admin?(@course) %>
           <%
              actions << {icon: 'replay', text: t(".reevaluate_submissions"), confirm: t(".confirm_reevaluate_submissions"), action: mass_rejudge_submissions_path(user_id: @user&.id, activity_id: @activity&.id, course_id: @course&.id, series_id: @series&.id, judge_id: @judge&.id)} if policy(Submission).mass_rejudge?
-             actions << {icon: 'done', text: t('.most_recent_correct'), search: {most_recent_correct_per_user: true}} if @activity
+             actions << {icon: 'done', text: t('.most_recent'), search: {most_recent_per_user: true}} if @activity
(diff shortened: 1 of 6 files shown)


app/views/submissions/index.html.erb
```diff
@@ -38,7 +38,7 @@
         <% if current_user&.course_admin?(@course) %>
           <%
              actions << {icon: 'replay', text: t(".reevaluate_submissions"), confirm: t(".confirm_reevaluate_submissions"), action: mass_rejudge_submissions_path(user_id: @user&.id, activity_id: @activity&.id, course_id: @course&.id, series_id: @series&.id, judge_id: @judge&.id)} if policy(Submission).mass_rejudge?
```
chvp, line 40: Not strictly required, but as far as I know this is to allow Steven to rejudge only the latest submissions, right? The `most_recent_correct` parameter isn't replicated here, so that still wouldn't work. Not necessarily something to fix in this PR, but maybe an issue should be created?

chvp (approved):

bmesuere (approved):
I agree with the changes in this PR.

To me, it is strange to put the filter in the dropdown. Especially because is it under the "options" header and filtering is different from the auto refresh option. Conceptually, it matches with the category filter dropdowns we already have.

In addition, the auto reload should conceptually be a toggle button instead of checkbox ("an action that takes place immediately")

PR #4488, opened 2023-03-16
Add copy code button to code fragments in descriptions

### LINKED ISSUE #3532: Copy code fragment to clipboard
For a lot of exercises it is useful to copy code-fragments from the assignment to use them for testing.
This is especially handy in python where the given examples can easily be used in doctests.

However, some of these examples are very long on 1 line (see [this](https://dodona.ugent.be/nl/courses/1173/series/13213/activities/835239763/) example).

In this case it could be very handy to have a button in the top right of this code-block to just copy the whole codeblock to the clipboard.

This would be very similar to the implementation of code fragments in the documentation of some frameworks.
Small example from the [prisma](https://www.prisma.io/docs/reference/api-reference/prisma-client-reference#create)-documentation:
image

jorg-vr:
This pull request adds a copy code button to code fragments in descriptions.

I matched on the `<pre><code></code></pre>` html tags in descriptions to match all logical cases. This tag is also correctly used by markdown.

image

Should the copying fail, it selects the text and suggests to use Ctrl + c

Closes #3532

### DIFF
--- a/app/assets/javascripts/components/copy_button.ts
+++ b/app/assets/javascripts/components/copy_button.ts
@@ -1,0 +1,56 @@
+import { ShadowlessLitElement } from "components/shadowless_lit_element";
+import { html, PropertyValues, TemplateResult } from "lit";
+import { customElement, property } from "lit/decorators.js";
+import { initTooltips, ready } from "util.js";
+
+@customElement("d-copy-button")
+export class CopyButton extends ShadowlessLitElement {
+    @property({ type: String })
+    code: string;
+
+    @property({ state: true })
+    status: "idle" | "success" | "error" = "idle";
+
+    async copyCode(): Promise<void> {
+        try {
+            await navigator.clipboard.writeText(this.code);
+            this.status = "success";
+        } catch (err) {
+            this.status = "error";
+        }
+    }
+
+    get tooltip(): string {
+        switch (this.status) {
+        case "success":
+            return I18n.t("js.copy-success");
+        case "error":
+            return I18n.t("js.copy-fail");
+        default:
+            return I18n.t("js.code.copy-to-clipboard");
+        }
+    }
+
+    protected updated(_changedProperties: PropertyValues): void {
+        super.updated(_changedProperties);
+        initTooltips(this);
+    }
+
+    constructor() {
+        super();
+
+        // Reload when I18n is loaded
+        ready.then(() => this.requestUpdate());
+    }
+
+    protected render(): TemplateResult {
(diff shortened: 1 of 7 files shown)


niknetniko (changes requested):
- For some reason it doesn't seem to work for me in Chrome (using version 111).
- The error message that appears is also not that useful I think ("Druk Ctrl-C om te kopiëren")
- Some (or most?) of the HTML problem statements don't use a `<code>` tag, only a `<pre>`, so they don't get a button.

jorg-vr:
> * For some reason it doesn't seem to work for me in Chrome (using version 111).

Fixed, it needed extra permissions because it is in an Iframe

> * The error message that appears is also not that useful I think ("Druk Ctrl-C om te kopiëren")

I now actually select the text if copying failed.

> * Some (or most?) of the HTML problem statements don't use a `<code>` tag, only a `<pre>`, so they don't get a button.

These problem statements do not follow the specification: https://docs.dodona.be/nl/references/exercise-description/#blokken-code Thus we cannot guarantee the layout.

I did update the matching form `code` to `pre code`, to avoid matching inline code

niknetniko (changes requested):
Would it be an idea to trim the copied code? For example, when copying this code:
image
It will result in this being copied:
image

Looks good otherwise.

bmesuere (changes requested):
Something goes wrong with inline code. They are now rendered as a block:
Before:
image

After:
image

The background of the "button" is transparent which causes un unwanted effect:
image

jorg-vr:
> Something goes wrong with inline code. They are now rendered as a block:

Forgot to push the fix for this

> Would it be an idea to trim the copied code?

That was an artifact of adding the copy button inside the codeblock. The copy button is now no longer included in the copied text. I prefer this solution to a trim, because leading and ending whitespace might be relevant in some code examples.

> The background of the "button" is transparent which causes un unwanted effect:

Fixed

app/assets/javascripts/components/copy_button.ts
```diff
@@ -0,0 +1,61 @@
+import { ShadowlessLitElement } from "components/shadowless_lit_element";
+import { html, PropertyValues, TemplateResult } from "lit";
+import { customElement, property } from "lit/decorators.js";
+import { initTooltips, ready } from "util.js";
+
+@customElement("d-copy-button")
```
bmesuere, line 6: The "default" web component comments are missing here.

bmesuere (approved):
Can we also replace the existing copy buttons by the new web component?

PR #4498, opened 2023-03-21
Fix relative time indication for submission history

### LINKED ISSUE #4490: Confusing time indication on submission page
image
Above screenshot was taken on March 20 at 11h30.

The time indication in the top right next to the submission number, says **today**, however the submission was submitted yesterday at 17h. While it is less than 24 hours ago, I would still mark these submissions as being submitted "yesterday".

I would only expect submissions submitted between midnight and now to be labeled as today.

jorg-vr:
This pull request fixes the relative time indication shown on the submission detail page.

The goal of this more custom version of time_ago_in_words was to avoid to many different groupings for recent submissions.
But I grouped to many times under today. Introducing yesterday fixes this.

image

- [x] Tests were added

Closes #4490

### DIFF
--- a/app/helpers/time_helper.rb
+++ b/app/helpers/time_helper.rb
@@ -1,0 +1,13 @@
+module TimeHelper
+  include ActionView::Helpers::DateHelper
+  def days_ago_in_words(time)
+    if time.today?
+      t 'time.today'
+    elsif time.yesterday?
+      t 'time.yesterday'
+    else
+      # need to use I18n.t here because the minitest helper dos not support kwargs yet for t
+      I18n.t 'time.ago', time: time_ago_in_words(time)
+    end
+  end
+end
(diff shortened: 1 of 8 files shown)


app/helpers/time_helper.rb
```diff
@@ -0,0 +1,13 @@
+module TimeHelper
+  include ActionView::Helpers::DateHelper
+  def days_ago_in_words(time)
+    if time.today?
+      t 'time.today'
+    elsif time.yesterday?
+      t 'time.yesterday'
+    else
+      # need to use I18n.t here because the minitest helper dos not support kwargs yet for t
+      I18n.t 'time.ago', time: time_ago_in_words(time)
```
jorg-vr, line 10: Made a pr on minitest-utils to fix the need for this: https://github.com/fnando/minitest-utils/pull/15

PR #4500, opened 2023-03-21
Check lower limit for memory_limit setting

chvp:
Docker does not allow a limit below 6MB, which results in internal errors if people try to use one that low anyway.

- [x] Tests were added

### DIFF
--- a/app/models/exercise.rb
+++ b/app/models/exercise.rb
@@ -163,7 +163,7 @@ class Exercise < Activity
       c = config
       c['evaluation'] ||= {}
       c['evaluation']['memory_limit'] = 10_000_000 # 10 MB
-      store_config(c, "raised memory limit for #{name}\n\nThe underlying system used to evaluate submissions does not allow a memory limity lower than 6MB.")
+      store_config(c, "raised memory limit for #{name}\n\nThe underlying system used to evaluate submissions does not allow a memory limit lower than 6MB.")
     end
   end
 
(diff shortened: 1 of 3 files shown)


app/models/exercise.rb
```diff
@@ -148,14 +148,23 @@ def last_submission!(user, deadline = nil, course = nil)
...
+      c['evaluation'] ||= {}
+      c['evaluation']['memory_limit'] = 500_000_000
+      store_config(c, "lowered memory limit for #{name}\n\nThe workers running the student's code only have 4 GB of memory " \
+                      "and can run 6 students' code at the same time. The maximum memory limit is 500 MB so that if 6 students submit " \
+                      'bad code at the same time, there is still 1 GB of memory left for Dodona itself and the operating system.')
+    elsif limit < 6_000_000 # 6 MB
+      c = config
+      c['evaluation'] ||= {}
+      c['evaluation']['memory_limit'] = 10_000_000 # 10 MB
+      store_config(c, "raised memory limit for #{name}\n\nThe underlying system used to evaluate submissions does not allow a memory limity lower than 6MB.")
```
niknetniko, line 166:
```suggestion
      store_config(c, "raised memory limit for #{name}\n\nThe underlying system used to evaluate submissions does not allow a memory limit lower than 6MB.")
```

niknetniko (approved):

jorg-vr (approved):

PR #4510, opened 2023-03-27
Add extra motivational messages generated by chatGPT

jorg-vr:
This pull request adds extra motivational messages that were generated by chatGPT. (There are now six times as many, for a total of 36)

I already removed and adjusted some, and generated some new dutch text using copilot. But the quality of the English texts is still a lot higher compared to dutch.

Feel free to remove/adjust text that you don't like.

I also changed the alert to a callout to better render all the emoji.
Old:
image
New:
image


bmesuere (changes requested):
In English, I don't like: 10, 15, 17, 20, 31, 32, 34, 7
In Dutch: 11, 15, 16, 17, 18, 20, 35, 36, 6, 7, 9 (actually most of the long ones, maybe we should keep those in Dutch short?)
25 -> Klus geklaard!

jorg-vr:
I removed those mentioned and generated some new ones through chatgpt.
I first asked for some with a 'ludieke kwinkslag', but chatgpt isn't great at that in dutch (I only kept two out of twenty texts). So then I just asked motivational messages, without the 'kwinkslag'.

I prefer replacing to removing messages, because the high variety of messages is one of the main benefits of this pr.

niknetniko (approved):

bmesuere (approved):

PR #4526, opened 2023-03-28
Take all javascript into account for coverage calculation

jorg-vr:
This pull request adjust the code coverage calculations to get a better view of our code coverage.

It fixes the coverage of javascript tests by including all javascript files. (Not just those touched by at least one test)
It fixes the javascript coverage of the system tests (which was not calculated before).

I have also updated the codecov setting to lower the default allowed deviation for the project to 0.25 (Used to be 1).


jorg-vr:
Currently:
- system coverage goes down => expected, more files considered
- project coverage goes down => unexpected, all javascript files should have already been considered by javascript tests
- project coverage is accepted with large decrease => I only allowed 0.25% decrease in #4524 
image

niknetniko (approved):

bmesuere (approved):

PR #4550, opened 2023-04-06
Fix broken links from feedback tab to code-tab

### LINKED ISSUE #4532: Broken links from stack trace to line in code
The links from a stack trace in the feedback report to a line in the Code tab seem to be broken.

Example: https://dodona.ugent.be/nl/feedbacks/133341763/

jorg-vr:
This pull request fixes the links to the code-tab from feedback tabs.

The bug in #4532 was caused because the tab names in the exercise start with the text _'code'_, eg:  https://dodona.ugent.be/nl/feedbacks/133341763/

The bug in https://github.com/dodona-edu/universal-judge/issues/285 was caused by the normal rendere using a different id for the code-tab than the pythia renderer.

Fixed both by giving the code tab a unique id.

Closes #4532 and https://github.com/dodona-edu/universal-judge/issues/285

### DIFF
--- a/app/assets/javascripts/submission.ts
+++ b/app/assets/javascripts/submission.ts
@@ -57,7 +57,11 @@ function initSubmissionShow(parentClass: string, mediaPath: string, token: strin
                 // prevent automatic scrolling to top of the page when clicking a link
                 e.preventDefault();
 
-                const tab = document.querySelector(`.feedback-table .nav-tabs > li a[href*='#tab-${tabName}']`);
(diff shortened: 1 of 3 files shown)


pdawyndt:
@jorg-vr Can you also check if [this issue](https://github.com/dodona-edu/universal-judge/issues/285) is resolved as well, or is this another issue that is only related to the TESTed judge?

jorg-vr:
@pdawyndt thanks for the reminder to also double check that issue. The cause was different and I have now fixed both.

niknetniko (approved):

bmesuere (approved):

PR #4566, opened 2023-04-12
Rework the question overview page

jorg-vr:
This pull request

- [ ] Tests were added

Closes # .

### DIFF
--- a/app/assets/javascripts/question_table.ts
+++ b/app/assets/javascripts/question_table.ts
@@ -47,6 +47,12 @@ export class QuestionTable {
                 return;
             }
 
+            const dropdownToggle = target.closest<HTMLDivElement>(".dropdown-toggle");
+            if (dropdownToggle) {
+                e.preventDefault();
+                return;
+            }
+
             const tr = target.closest<HTMLElement>("tr.selection-row");
             if (enableClick && tr) {
                 e.preventDefault();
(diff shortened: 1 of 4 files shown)


app/views/annotations/_questions_table.html.erb
```diff
@@ -1,81 +1,74 @@
...
+        <% end %>
+        <div class="ellipsis-overflow" title="<%= question.submission.exercise.name %>">
+          <%= question.submission.exercise.name %>
+        </div>
+        <div class="ellipsis-overflow text-muted <%= "icon-indent" unless @course.present? %>">
+          <%= question.question_text %>
+        </div>
+      </td>
+      <td class="d-none d-md-table-cell">
           <span title="<%= l question.created_at, format: :question %>">
```
niknetniko, line 36: This seems like it is indented too much now.

app/views/annotations/_questions_table_with_pagination.html.erb
```diff
@@ -0,0 +1,4 @@
+<%= render partial: 'questions_table', locals: { questions: questions} %>
```
niknetniko, line 1:
```suggestion
<%= render partial: 'questions_table', locals: { questions: questions } %>
```

niknetniko (changes requested):
The icon's positions look weird on the global question page for me (in Chrome):

image

niknetniko (approved):

bmesuere (changes requested):
The questions table has an auto refresh. To prevent unintended clicks due to layoutshifts, the table is never updated on mouse over (I think). This behaviour should be expanded to also not update when a dropdown menu is shown because such updates will cause the dropdown to collapse.

bmesuere (approved):

Issue #2826, opened 2021-06-05
Allow multi-line annotations


chvp:
Allow to add comments/questions on a continuous selection of multiple lines.

From https://github.com/dodona-edu/dodona/issues/1852#issue-599457917

bmesuere:
Github behaviour:
- if you hover over a line, a plus button appears in the left margin
- if you hover over the button, the button grows in size
- if you click the button, the comment form appears
- if you drag the button, the button stays in place but a thick line in the color of the button appears in the margin
  - you can only drag down, not up
- if you release the drag, the comment form appears. The button and line disappear, but the selected lines are highlighted in yellow. Above the text area, a small help text "Commenting on lines xx to xx" is shown. For single line comments, the highlight and info text is not present.
- when rendering a multi-line comment in the code view, GitHub does **not** show which lines the comment belongs to. It is rendered under the last line of the selection as a normal comment. In the PR-comment view where just the comments are rendered the additional lines are included in the small code snippet that shows context.

pdawyndt:
Maybe related:

- comments could be moved to another location by updating the numbers in a help text like "Commenting on lines xx to xx" (both for single line and multiline)
- I often repeat the same comment within the same submission: copy followed by move might speed up this process

PR #4568, opened 2023-04-14
Render machine annotations for multiple lines or columns

jorg-vr:
This pull request renders machine annotations more precisely if the judge provided the optional parameters. Machine annotations are now marked in the code. The annotation can be found as an interactive tooltip on the marked code. If multiple annotations cover the same code, the marking will indicate the most important annotation, while the tooltip will show all annotations.

I updated the default display for inline annotations to 'important' when any annotations exist on page load, or when in an evaluation, as annotations are easily accessible by tooltip.

This is a first step towards more precise human annotations.
I started with machine annotations as there is less code involved and it can help to test designs already. My idea to mark human annotation ranges in a similar way with a background color.

This pr introduces [Tippy.js](https://atomiks.github.io/tippyjs/v6/all-props/#placement): This helps with the interactive tooltips. It is written on top of popper.js which we already use through bootstrap. I added is because it greatly improved the UX experience in some edge case compared to using only popper.js.

Visuals of some manually generated annotations: (They are not realistic as I tried to cover all edge cases)
image
image

Currently for most judges it will look like this, with only full line markings.
image
image
Some more naos examples with comments:
- https://naos.ugent.be/nl/submissions/13036445/
- https://naos.ugent.be/nl/submissions/13036538/
- https://naos.ugent.be/nl/submissions/13036580/
- https://naos.ugent.be/nl/submissions/13036618/
- https://naos.ugent.be/nl/submissions/13036388/
- https://naos.ugent.be/nl/submissions/13036326/

But tested has already been updated in https://github.com/dodona-edu/universal-judge/pull/341

- [x] Tests were added
- [x] Documentation update can be found at https://github.com/dodona-edu/dodona-edu.github.io/pull/240

Progress on #2826


bmesuere (changes requested):
This looks very promising! Some quick (and incomplete) feedback:

- [x] I would still show all linting comments by default for the "student view" and only "hide" them if there is at least 1 manual comment/question. Maybe hiding them during grading might also be the right approach, but @chvp is better positioned to answer that question.
- [x] When showing all annotations, the tooltips are still enabled which should not be the case I think?
- [x] The order in the tooltips is different then when shown normally e.g.: line 21 on https://naos.ugent.be/nl/submissions/13036445/ In the tooltip, the info remark is shown in the top, but in the inline view it is shown at the bottom.

niknetniko (approved):

bmesuere (approved):

pdawyndt:
This is how the wave underline looks for me (Chrome, Windows 10):

image

PR #4582, opened 2023-04-19
Speed up submission result construction

### LINKED ISSUE #4579: Handling of partial output format is slow
From executing https://dodona.ugent.be/nl/courses/2263/activities/649388223/ on my local machine:

image

Looking at the code, some obvious things seem the validation of each JSON object individually. The same exercise without validating each command gives:

image

niknetniko:
This pull request modifies how we construct results from the judge output for a submission.

First, it replaces the JSON Schema validator library with another one, which already speeds things up a lot (locally, it went from 5.5s to 0.5s).

Secondly, it modifies the logic to (in production) only validate once at the end, instead of for each command.

I am not 100% convinced if the second part is necessary, since the speedup is fairly small, but the code is uglier in my opinion (and there is then a difference between development/production in an important part of the code).

- [x] Tests were added

Closes #4579 .

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -37,7 +37,7 @@ gem 'image_processing', '~> 1.12.2'
 gem 'bootsnap', '~> 1.16.0', require: false
 
 # used to validate container responses
-gem 'json-schema', '~> 3.0.0'
+gem 'json_schemer', '~> 0.2.24'
 
 # delayed jobs
 gem 'delayed_job_active_record', '~> 4.1.7'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -160,6 +160,8 @@ GEM
     docker-api (2.2.0)
       excon (>= 0.47.0)
       multi_json
+    ecma-re-validator (0.4.0)
+      regexp_parser (~> 2.2)
     ed25519 (1.3.0)
     erubi (1.12.0)
     exception_notification (4.5.0)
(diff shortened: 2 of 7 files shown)


jorg-vr (commented):
I am also not a fan of different behavior between development and production here.

As I understood this output is provided for judge developers. I am not sure if running dodona in development is a requirement to develop a judge (might as well test it on naos for example). And it might become rather unintuitive for someone developing a judge that their error messages in production are different from those in their development environment 

My proposed solution would be:
- only do final full validation (As production mode is written right now)
- if errors are caught in this process, run the partial validations in an effort to provide a more useful error message

jorg-vr:
> 2\. If the invalid json is only caught at the end, there no way to know which of the updates from the partial format caused the issue without saving all partial updates and the re-checking all of them, which seems like a lot of work to me.

The work is storing a reference to `judge_output`. As all updates happen with cloned json, this will not be changed.

Then you need to run something like this in case of an error:
```ruby
     split_jsons(judge_output).each do |json|
       unless PART_SCHEMER.valid?(json.deep_stringify_keys) || FULL_SCHEMER.valid?(json.deep_stringify_keys)
          raise ResultConstructorError.new(
            'Judge output is not a valid json',
            json.to_s
          )
        end
      end
```

niknetniko:
For now, we'll begin with using the gem (#4583) and then think about not validating as much later (since there various factors to consider there).

PR #4596, opened 2023-04-25
Allow creation of user annotations based on a selection

### LINKED ISSUE #3976: Selecting code also includes the annotations
It used to be possible to copy code without including the annotations. This is no longer possible:

image

Fixing this probably means that you can no longer copy annotations (which was also the case previously). 

However, especially during grading, I would argue not being able to copy a piece of code is more annoying than not being able to copy an annotation, as you can just click edit on the annotation and copy it that way. Copying code without the annotation means you need to scroll back to the top, hide the annotations, scroll back down, copy the code, scroll back up, re-enable annotations and scroll back down again.

jorg-vr:
This pull request marks user annotations inside the code. It uses a background instead of a text decoration. This way one machine annotation and one user annotation can be visualized simultaneously. The most important type of each will get priority.

It is possible to mark part of a line, or multiple whole lines. If a line is only partially selected in a multi line selection, the selection will be adjusted to select the whole line.

Examples:
image

While selecting code, selecting other text or annotations is disabled. This makes it easier to select code across annotations. This also fixes #3976 . It is still possible to select text within an annotation if your selection starts there. Code selections and other selections are differentiated by the selection color and the side button becoming larger.

Peek 2023-05-15 14-51

Using the old create annotation button marks the whole line:
Peek 2023-05-15 14-48

When all annotations are hidden, the marked annotations have tooltips:
Peek 2023-05-15 14-49

- [x] Tests were added
- [ ] Documentation update can be found at dodona-edu/dodona-edu.github.io# (Will be updated after approval)

Note: Testing was hard as firefox implements selection differently than chrome. The testing framework works similar to chrome, so the firefox code could not be tested.

Closes #3976 and #2826


app/assets/javascripts/state/UserAnnotations.ts
```diff
@@ -33,57 +32,77 @@ export interface UserAnnotationPermissionData {
...
+        this.selectedRange = null;
+        this.showForm = false;
+    }
+
     // public for testing purposes
     public async addToMap(annotation: UserAnnotationData): Promise<void> {
         this.byId.set(annotation.id, annotation);
         if (!annotation.thread_root_id) {
-            const line = annotation.line_nr ?? 0;
+            const line = annotation.line_nr + annotation.rows - 1 ?? 0;
```
chvp, line 91: Does this still work with the nullish coalescing?

chvp (changes requested):
- [x] Not sure if new, but the badge for annotation count is always shown, even when it's 0: screenshot_2023-05-04-171835

- [x] I would add some sort of background to the add button, the text overlapping the border looks weird to me.
screenshot_2023-05-04-172047

pdawyndt:
What is the reasoning to snap selections that run across multiple lines to whole lines?

jorg-vr:
@pdawyndt this was requested by @bmesuere 

The reasoning was that this would make it a lot faster to select whole lines, which in Bart's opinion would be the most common use case.
We also could not think of any use case for a range spanning multiple lines with an exact end and starting point.

If people with more experience grading have a different opinion on this, I am willing to change this.

jorg-vr:
I used the elevated button style, is this better @chvp?
image
image
image
image

chvp (approved):

bmesuere (changes requested):
- [x] Regarding the button: why not make this a proper [FAB](https://m3.material.io/components/extended-fab/overview)? The collapsed version could then expand on mouseover instead of the tooltip now.

- [ ] Would it be possible to also drag the button to select lines? Drag release would then automatically trigger the button click which would be the fasted way to add a multi line comment: big click target to start drag, drag a few lines, release and directly start typing.

- [ ] The background color while selecting is ok, but seems a bit strange after adding the comment. For comments of a substring, I would expect it to mimic the curly underline of the linting comments. For single or multi-line comments it seems very prominent. Alternatives: show something in the margin or show no formatting in the code but add a "comment on line x" "comment on line x to y" to the top of the comment (which is what gitlab does). Mouseover on the line(s) or comment could then show the range using a  bg-color.

bmesuere (changes requested):
looks good, some nitpicks:
- [ ] I think it will look better if you keep the icon left and put the text on the left. Maybe a little css transition to expand the button will also help with the smoothness
- [ ] on mouse over, the button in slightly darker (in light mode), however, the little arrow stays in the base color
- [ ] dragging the FAB doesn't work for me (chrome). I can drag the button (both in x- and y-direction, where I would expect the x-position to be locked), but when I release it, the button is gone and no form appears. In addition, the button also doesn't show up when I hover over the code. There is no error in the console.
- the default line highlights look more subtle now, well done! 
- the darker  highlight on mouse over is also very nice!

bmesuere:
Maybe we can move the button drag behaviour to a later PR to be able to make a release tomorrow.

jorg-vr:
@bmesuere I did not add the transition because I use the display properties right now got the button change, which don't work with css transitions.
For transitions I would need to use something like visibility, but this would also require a rework of the html structuring and some nit picky positioning. 

I will work on this after the release if desired.

niknetniko (approved):

PR #4616, opened 2023-05-05
Strike through code

jorg-vr:
This pull request

- [ ] Tests were added

Closes # .


pdawyndt:
Alas, but my 50ct is that this will be a feature that we'll use a lot!

jorg-vr:
For now this change is not desired.

It would promote more lazy reviewing, which would result in a worse quality of code reviews.

PR #4617, opened 2023-05-05
Check if course id is provided for series create

### LINKED ISSUE #4147: Series create crashes if no course id is provided

freyavs:
This pull requests prevents series create from crashing when no course id is provided.
Closes #4147


jorg-vr (approved):

app/controllers/series_controller.rb
```diff
@@ -108,9 +108,13 @@ def edit
...
-    authorize @series.course, :add_series?
+    if @series.course.present?
+      authorize @series.course, :add_series?
+    else
+      @series.errors.add(:course_id, I18n.t('errors.messages.blank'))
+    end
     respond_to do |format|
-      if @series.save
+      if @series.course.present? && @series.save
         format.html { redirect_to edit_series_path(@series), notice: I18n.t('controllers.created', model: Series.model_name.human) }
```
niknetniko, line 118: I am not entirely convinced this was the best solution, primarily because:

- We now do validation directly in the controller: the model itself should validate the presence of the course, not the controller.
- The check is done twice in the controller: once on line 111 and again on 117. I'm also not sure why 117 is necessary: if there are validation errors (as there are here, since we manually add them on line 114), the save method should not succeed.

Besides those, some more thoughts:

- This isn't something that this PR did, but why do we authorise here on the `Course` policy, instead of the `create?` method of the `Series` policy (they seem to do the same thing)? Or, the other way around then, why is there a `create?` method on the `Series` policy if it isn't used here?
- If we were to authorise on the `Series` policy, the crash would have also gone away, but the response would be 403 instead of 422. While maybe not the best response, it is more consistent with how we do it in other controllers (e.g. the `evaluation` controller) I think.

(The current code does work however, so this isn't very urgent)

bmesuere: You're right, authorizing on series would make more sense

PR #4618, opened 2023-05-05
modal.js: remove jquery and convert to typescript

freyavs:
This pull request removes all jquery from modal.js and converts it to typescript.

Progress on #3590


jorg-vr (approved):

bmesuere (commented):
The CodeQL warnings seem to be valid concerns. @jorg-vr do you know if we need to be able to inject html into these?

jorg-vr:
Looking at the three usecases it should support html. (all usecases are for pythia submissions)
But one case has user input in the template string (a filename, chosen by a teacher, is used in the title), thus we best improve the safety of the method

Might be best to use the `render` function from `lit`

Eg `render(title, document.querySelector("#info-modal .modal-title"))`
The title and content arguments will have to be of type `TemplateResult | string`
And usages of the function could use `html\`\`` to generate template strings, which will make sure everything is safe

app/assets/javascripts/modal.ts
```diff
@@ -0,0 +1,19 @@
+import { Modal } from "bootstrap";
+
+function showInfoModal(title: string, content: string, options?: {allowFullscreen: boolean}): void {
+    const button = document.querySelector("#info-modal .modal-header #fullscreen-button") as HTMLElement;
+
+    if (options && options.allowFullscreen) {
```
niknetniko, line 6:
```suggestion
    if (options?.allowFullscreen) {
```

app/assets/javascripts/modal.ts
```diff
@@ -0,0 +1,19 @@
...
+        button.style.display = "inline";
+    } else {
+        button.style.display = "none";
+    }
+
+    document.querySelector("#info-modal .modal-title").innerHTML = title;
+    document.querySelector("#info-modal .modal-body").innerHTML = `<p>${content}</p>`;
```
niknetniko, line 13: Looking at the usage of this function, we should probably remove the `<p>` tags. Most content we pass is wrapped in a `<div>`, which is not allowed inside a `<p>` tag.

niknetniko (changes requested):

jorg-vr (approved):

bmesuere (approved):

niknetniko (approved):

jorg-vr (approved):

Issue #4564, opened 2023-04-11
Tracking: saved annotation feedback


jorg-vr:
### Feature requests
- [ ]  Broader annotation scope
  Some comments can be relevent for every exercise in a course
  Or for the same exercise in different courses
  Or for multiple admins of the same course
  - https://github.com/orgs/dodona-edu/discussions/4064#discussion-4446267
  - https://github.com/orgs/dodona-edu/discussions/4064#discussioncomment-4123290
  - https://github.com/orgs/dodona-edu/discussions/4064#discussioncomment-4128124
  
conclusions after discussion:
- dont share annotations between users
- determine scope when saving
- possible scopes: exercise, course, global
Timing:  version 2

- [x] **Don't require anticipations of reuse**
  Make it posible to search for annotations that have not been 'marked' for reuse
  - https://github.com/orgs/dodona-edu/discussions/4064#discussioncomment-3801922
  - https://github.com/orgs/dodona-edu/discussions/4064#discussioncomment-4134181

Conclusions after discussion:
- create annotations index page with search
- can expand saved annotations index page
Timing: version 1/2?
  
  
### Enhancements
- [x] **Use arrow keys to select saved annotaion**
  - https://github.com/orgs/dodona-edu/discussions/4064#discussioncomment-4119363
 Timing version 1
  
- [ ] **order saved annotation by most recently used/added**
  - https://github.com/orgs/dodona-edu/discussions/4064#discussioncomment-4117857
Timing: depends on research CHVP
  
- [x] **allow editing saved annotation by clicking 'linked' icon**
  - https://github.com/orgs/dodona-edu/discussions/4064#discussioncomment-4118022 But I would not display the title inline
COD:
- Rethink link icon (Could create wrong expectations) => comment bank
- Open separate page for saved_annotation
> from https://github.com/dodona-edu/dodona/pull/4701#pullrequestreview-1562299868: when picking a suggestion, the "save annotation" checkbox is disabled, but it is also shown as checked which might be confusion. Maybe replace it with a link to the page where you can edit that annotation?

Timing: version 1
  
<del>

- [ ] **allow updating all usages of a saved annotation**
  - https://github.com/orgs/dodona-edu/discussions/4064#discussioncomment-5228124
  - https://github.com/orgs/dodona-edu/discussions/4064#discussioncomment-4118151
 Timing: never
  
- [ ] **Allow multiple language versions of saved annotations**
[...]

PR #4701, opened 2023-06-01
Update annotation form

jorg-vr:
This pull request restructures the annotation form in such a way that the saved annotations options don't take up to much extra space.

New:
image
Old:
image

New (save annotation):
image
Old (save annotation):
image

New (mobile):
image
Old (mobile):
image

Part of #4564


bmesuere (changes requested):
Feedback from running this on metra:
- [x] the hover button text shows ask question instead of add comment (the global button does show add comment)
- [x] if there are no saved annotations, the area to the right of the textarea is left blank. I would add an empty state there or show the textarea full width.
- [x] if you check "save annotation", the title form is shown to the right of the text area. This somewhat breaks the flow. Maybe placing it inline after the checkbox + save annotation text would make more sense?
- [ ] ~~when save annotation is checked, it would make sense to change the text on the submit button to also say something that hints that the annotation will be saved.~~
- [x] the help text under the "search for annotations" combobox is too long (at least in Dutch)
- [x] in the search annotations input, the list of suggestions is not the same width as the text input
- [ ] in the search annotations input, the list of suggestions "flickers" when searching
- [x] when picking a suggestion, the "save annotation" checkbox is disabled, but it is also shown as checked which might be confusion. Maybe replace it with a link to the page where you can edit that annotation?
- [x] when loading a saved annotation and then changing the text, a "not equals" sign is shown in the search annotation input. The icon is not aligned well (shown somewhere in the middle of the input field)

jorg-vr:
> when picking a suggestion, the "save annotation" checkbox is disabled, but it is also shown as checked which might be confusion. Maybe replace it with a link to the page where you can edit that annotation?

This page does not exist yet (It is a modal currently)
I will just hide the checkbox for now and reintroduce that link when that page is created.
Added this note in https://github.com/dodona-edu/dodona/issues/4564

jorg-vr:
> when save annotation is checked, it would make sense to change the text on the submit button to also say something that hints that the annotation will be saved.

I am not the biggest fan of this. I find it hard to come up with better wording for the button. A longer text is also makes it harder to properly layout the buttons (as they are on the same line as the checkbox currently)

The checkbox is also right next to the button, so the information is already provided on almost the same place

jorg-vr:
>  the help text under the "search for annotations" combobox is too long (at least in Dutch)

I removed `Saved annotations are automatically filtered by course and exercise. `
This will change in version 2 anyway

niknetniko (approved):

bmesuere (changes requested):
- [x] The wording of "sla opmerking op om te hergebruiken" is quite verbose. Maybe make it "Opmerking opslaan" (maybe with an info icon with tooltip?). This would also help with my previous comment that it wasn't really clear that the comment would also be save when clicking the submit button
-  [ ] The "Titel:" label is a bit strange. Maybe [floating labels](https://getbootstrap.com/docs/5.3/forms/floating-labels/) are an option here?
- [x] Upon clicking "sla opmerking op", the titel input should get the focus
- [ ] On smaller screens, the form layout isn't optimal:
  image

jorg-vr:
> The "Titel:" label is a bit strange. Maybe [floating labels](https://getbootstrap.com/docs/5.3/forms/floating-labels/) are an option here?

The floating labels don't work for our inputs, as we have a lot less vertical padding compared to the bootstrap default. Thus their is no vertical space for a label inside the input, without making the input look comically large compared to others.

I did make some custom styling based on the bootstrap css to put the label inside the input before the typing area 
image

app/assets/stylesheets/components/code_listing.css.scss
```diff
@@ -478,35 +478,44 @@
...
-    }
 
-    .annotation-submission-input {
-      border-radius: 0;
-      resize: vertical;
+      label {
+        color: rgba(var(--d-on-surface-rgb), 0.75);
+        transform: scale(0.85) translateY(-0.4rem);
+        position: absolute;
+        top: 0;
```
bmesuere, line 503:
```suggestion
        top: 1px;
```

In my Chrome, the label was not aligned with the textbox text.

app/assets/javascripts/components/annotations/annotation_form.ts
```diff
@@ -176,95 +188,120 @@ export class AnnotationForm extends watchMixin(ShadowlessLitElement) {
         this.inputRef.value.focus();
     }
 
+    updated(changedProperties: Map<string, any>): void {
+        // Focus the newly shown title input if the user wants to save the annotation.
+        if (changedProperties.has("saveAnnotation") && this.saveAnnotation) {
+            this.titleRef.value.focus();
```
bmesuere, line 194: I would also call `select()` here: https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/select

bmesuere (approved):
There's a JS error on mestra when signed in as staff. Otherwise, looks good 👍

Issue #4673, opened 2023-05-24
Change deploy process


bmesuere:
The goal of this plan is to limit the impact of changes by deploying as often as possible, preferably after each pull request merge. This would also allow further automation by using GitHub Actions. However, there are some drawbacks and unknowns that need to be addressed.

## Deployment process
Ideally, all deployments would be done through GitHub Actions. It should be possible to start a deployment manually and automatically when merging a PR. Ideally, a deployment would go through the following steps:
- Run all tests
- Test the deployment on naos (can we deploy the current production version first?)
- Deploy on dodona

## Steps to implement this

```[tasklist]
### Steps
- [ ]  add a new main branch
- [ ] replace all mentions of master by main
- [ ] set the "main branch" of this repo to main instead of develop
- [ ] change the branch protection rules on main/remove them on develop
- [ ] add a GitHub Action to deploy to production
- [ ] change the capistrano config: allow to deploy commit hashes to production, remove mentions of develop
- [ ] remove the hotfix creation script
```

```[tasklist]
### To discuss
- [ ] how will we implement rollback? Does capistrano offer something?
- [ ] can we automatically add some kind of version number to the footer to indicate the deployed version?
```

### What about dependabot PRs?
In this case, we probably only want to do a single deploy instead of one every merge. I see two possible strategies:
Only start a deployment if the PR has the bug, feature or enhancement label (and thus not on dependencies and chores).
Only deploy PRs that have a deploy label (adding the label could be automated)

### What about grouped PRs?
This is somewhat similar to the question above, but in this case I would use a different strategy and only deploy if the PR target is the main branch. Grouped PRs could then merge into an intermediate branch first.

### What about the exam period?
We tend not to do large releases during the exam period. Releasing more often would reduce the risk of a release, so not releasing during these periods would be less of an issue with this new strategy.

If we think that a PR might cause problems, we can still postpone it until the evening/weekend or after the exams. Note that the exam period becomes a very large part of the term if you combine secondary and higher education.

### Don’t we need version numbers for marketing?
We still want some sort of changelog. Creating a new GitHub release for each deployment will quickly become a mess. One solution might be to decouple GitHub releases from deployments. We could create a new release every month that contains all the changes from the previous month. We can then still document our changes with a topic in GitHub Discussions.

[...]

PR #4705, opened 2023-06-02
Set version as timestamp during capistrano deploy

jorg-vr:
This pull request adds a new Capistrano task to save a formatted timestamp of deployment in a config  file.

This is used as the version number to be displayed in the app.

New versioning syntax used `"%y.%m.%d%H%M"`. This would match with monthly releases of the form `%y.%m`. Eg. `23.07`
part of #4673


bmesuere:
maybe add a dash between day and hour?

bmesuere (approved):

PR #4744, opened 2023-06-07
Set version number during capistrano deploy

jorg-vr:
This pull request overwrites the version config file on Capistrano deploy to make sure it uses a fixed timestamp of deploy time.

part of #4673

### DIFF
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -118,11 +118,12 @@ namespace :deploy do
   end
 end
 
+version = Time.now.strftime("%y.%m.%d-%H:%M")
+
 namespace :deploy do
-  before :publishing, :asset_stuff do
-    on roles :web do
+  before :publishing, :set_version do
+    on roles :app do
       within release_path do
-        version = Time.now.strftime("%y.%m.%d-%H:%M")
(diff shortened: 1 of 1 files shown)


config/deploy.rb
```diff
@@ -117,3 +117,14 @@
     end
   end
 end
+
+namespace :deploy do
+  before :publishing, :asset_stuff do
```
chvp, line 122:
```suggestion
  before :publishing, :set_version do
```

config/deploy.rb
```diff
@@ -117,3 +117,14 @@
...
+    on roles :web do
+      within release_path do
+        version = Time.now.strftime("%y.%m.%d-%H:%M")
```
chvp, line 125: If possible, this should probably be set outside the `namespace` statement to make absolutely sure we use the same version on all hosts.

config/deploy.rb
```diff
@@ -117,3 +117,14 @@
     end
   end
 end
+
+namespace :deploy do
+  before :publishing, :asset_stuff do
+    on roles :web do
```
chvp, line 123: We want to do this on the workers as well, right?

chvp (changes requested):

chvp (approved):

bmesuere (approved):

PR #4745, opened 2023-06-08
Unify css with bootstrap themes

jorg-vr:
This pull request removes the separate dark mode css file. It is replaced by the new bootstrap 5.3 css theming.

A new js theme picker is also added, supporting, darkmode, lightmode and system mode. This more general approach will allow more themes in the future.

There were a lot of css variable to overwrite, more so because every component has some own css variable.
I have thus not overwritten every possible variable, but started with the obvious ones and then did some visuals checks while walking through the application. I then updated those colors which stood out as odd.
This means I might have missed some variables that should be overwritten because the relevant bootstrap class is only used on one specific page. Please let me know if you find such an occurrence.

I was also disappointed that I had to write my own `btn-danger` and `btn-success` variant classes as the bootstrap instances did not support theming well.

Changes:
- [x] Rewrite js theme selection logic
- [x] Rename toggle_dark_mode to set_theme
- [x] Remove application_dark.scss
- [x] Use css variables instead of sass variables everywhere
- [x] Set correct variables for newly introduced bootstrap scss variables
- [x] d3 graphs now use theme colors instead of redefining them

New theme selector:
image
image

The preferred theme is now stored for the user

Other pages should be unchanged.

Closes #4711 as all changes are contained here


app/views/announcements/edit.html.erb
```diff
@@ -11,7 +11,7 @@
       </div>
       <div class="card-actions card-border">
         <% if policy(@announcement).destroy? %>
-          <%= link_to @announcement, method: :delete, data: { confirm: t("general.are_you_sure") }, class: "btn btn-filled with-icon  bg-danger" do %>
+          <%= link_to @announcement, method: :delete, data: { confirm: t("general.are_you_sure") }, class: "btn btn-filled with-icon  d-btn-danger" do %>
```
chvp, line 14:
```suggestion
          <%= link_to @announcement, method: :delete, data: { confirm: t("general.are_you_sure") }, class: "btn btn-filled with-icon d-btn-danger" do %>
```

app/controllers/pages_controller.rb
```diff
@@ -50,9 +50,9 @@ def toggle_demo_mode
     session[:demo] = !Current.demo_mode
   end
 
-  def toggle_dark_mode
+  def set_theme
     authorize :pages
-    session[:dark] = params[:dark].nil? ? !session[:dark] : ActiveModel::Type::Boolean.new.cast(params[:dark])
+    session[:theme] = params[:theme].nil? ? 'auto' : ActiveModel::Type::String.new.cast(params[:theme])
```
chvp, line 55: Do we want to do validation of the options here? The string cast isn't necessary, params are strings by default.

chvp (changes requested):
Some differences I noticed. Dodona on the left, naos on the right.

screenshot_2023-07-06-100519
Color of "Nog niet opgelost".

screenshot_2023-07-06-100944
Same color issue.
screenshot_2023-07-06-101908
There is more padding in the scrollspy on the right of a course.

These same issues exist in dark mode.

jorg-vr:
`.text-muted` (eg niet opgelost)  still used bootstrap default color `#6c757d` which was not part of our color theme
 I replaced it with $gray-50 in both light and dark mode (which looks similar)
 I could also use $on-surface-muted which follows the m3 specs, but is a lot less muted.
 
$gray-50
image

$on-surface-muted
image

Old: `#6c757d`
image

bmesuere (changes requested):
👍 nice work. This removes a few scss hacks and unifies our variables to css.

Some remarks:
- [x] I would not add a theme menu in the main navigation bar and instead make it a sub menu in the current user menu.
- [ ] The contrast of the selected option isn't great. The text should be white here (as is the case with the year selector on the signed in home page)
  image
- [ ] I found it odd that the icon for auto displayed the current state. Maybe use https://pictogrammers.com/library/mdi/icon/theme-light-dark/ ?
- [x] Is there a reason why the icons are inlined?
- [ ] The sign in page looks bad in dark mode due to the icons. Should we disable dark mode here?
- [ ] Why do we store theme in the session? (I know you didn't change this, just asking the question) I would make it entirely client side (and thus use local storage) or store it in the database as a user setting. If we store it in the database, we could also add the option to the profile page.
- [ ] The theme controller call should be excluded from being "stored". If you go to the sign in page, then switch themes and then sign in, you are redirected to /theme which doesn't exist.

app/assets/javascripts/heatmap.ts
```diff
@@ -70,7 +71,7 @@ function initHeatmap(url: string, oldestFirst: boolean, year: string | undefined
 }
 
 function drawHeatmap(data: [dayjs.Dayjs, number][], oldestFirst: boolean, year: string | undefined): void {
-    const darkMode = window.dodona.darkMode;
+    const darkMode = themeState.theme === "dark";
```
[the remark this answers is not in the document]
jorg-vr: I have instead removed the hardcoded colors by using the css variables in javascript

bmesuere (approved):
- label for dark mode in menu
- always light mode when not signed in

chvp (approved):

PR #4808, opened 2023-07-05
Create monthly release workflow

### LINKED ISSUE #4673: Change deploy process

jorg-vr:
This pull request adds a new workflow that will automatically create a release with a discussion every month on the first day of the month.

Closes #4673

### DIFF
--- a/.github/workflows/monthly-release.yml
+++ b/.github/workflows/monthly-release.yml
@@ -6,7 +6,6 @@ on:
 jobs:
   create-release:
     runs-on: ubuntu-latest
-    needs: create-tag
     steps:
       - name: Get tag name
         id: tag_name
--- a/.github/workflows/monthly-release.yml
+++ b/.github/workflows/monthly-release.yml
@@ -10,7 +10,7 @@ jobs:
     steps:
       - name: Get tag name
         id: tag_name
-        run: echo "{tag_name}={$(date +'%Y.%m')}" >> $GITHUB_STATE
+        run: echo "tag_name=$(date +'%Y.%m')" >> "$GITHUB_ENV"
       - name: log tag name
         run: echo ${{ env.tag_name }}
       - uses: octokit/request-action@v2.x
--- a/.github/workflows/monthly-release.yml
+++ b/.github/workflows/monthly-release.yml
@@ -19,10 +19,8 @@ jobs:
           route: POST /repos/{owner}/{repo}/releases
           owner: dodona-edu
           repo: dodona
-          tag_name: |
-            ${{ env.tag_name }}
-          name: |
-            ${{ env.tag_name }}
+          tag_name: "'${{ env.tag_name }}'"
+          name: "'${{ env.tag_name }}'"
           discussion_category_name: "Release notes"
           generate_release_notes: true
         env:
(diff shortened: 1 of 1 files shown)


bmesuere (commented):
Don't you also have to create the tag?

jorg-vr:
There is no need to create the tag. The tag will be created by the api call
https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#create-a-release

Eg this tag was created by me testing this script: https://github.com/dodona-edu/dodona/releases/tag/2023.07-test

bmesuere (approved):

PR #4821, opened 2023-07-10
Simplify full line annotation markers

jorg-vr:
This pull request creates only a single marker for annotations that span a full line.

In a complex example that reduced the number of rendered marker webcomponents from 948 to 433

- [ ] Tests were added
- [ ] We learned with dolos that using ... isn't a good idea with variable length arrays. This had unexpected performance implications.
- [ ] Don recreate tooltips every time

Closes # .


jorg-vr:
There is already a more relevant speedup included in #4643 
When reviewing #4643 we also decided that we will do the markings on hidden text, making these changes irrelevant

PR #4837, opened 2023-07-21
Changed from 'paste' to 'copy' icon.

dvanderfaeillie:
Dodona uses the Material Design Icon `mdi-clipboard-outline` icon to symbolize "copy to clipboard". But that icon is widely recognized as a "paste" icon (The Material Design Icon `mdi-content-paste` is only a tiny bit longer).

Doesn't it make more sense to use `mdi-content-copy`? Proposing the following tiny changes.

### DIFF
--- a/app/assets/javascripts/components/copy_button.ts
+++ b/app/assets/javascripts/components/copy_button.ts
@@ -64,7 +64,7 @@ export class CopyButton extends ShadowlessLitElement {
                             data-bs-placement="top"
                             data-bs-toggle="tooltip"
                             title="${this.tooltip}">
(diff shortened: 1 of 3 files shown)


bmesuere:
Thank you for your suggestion!

We've talked about this internally before. There are indeed "default" cut, copy and paste icons, but those are often used together which helps with adding context. In Dodona, the copy icon is used on a button without any further context. In addition, the material design copy icon is very abstract: just 2 squares.

With that in mind, we decided that the clipboard icon does a better job at conveying the "copy to clipboard" meaning.

We are not alone in doing this, by the way, both vitepress and the MDN docs also use a clipboard icon:
image
image

PR #4846, opened 2023-07-25
Split code rendering into different functional layers

### LINKED ISSUE #4654: Wave pattern not always easily visible
This [submission](https://dodona.ugent.be/nl/courses/2263/series/24188/activities/30756865/) has a wave pattern below both the `i` and the  `j` characters on line 98. Because the character `j` hangs out of the line, the wave pattern is interrupted there, but as a result not a lot of it is visible anymore.

image

Browser: Chrome (version 113.0.5672.93, 64-bits) on Windows 10

jorg-vr:
This pull request splits up code rendering into four layers:
- the background layer renders the markup for annotations on the code (background color and underline)
- the selection layer renders the background color for the selection (split off because keeping selection logic separate from existing annotations makes it much faster)
- the tooltip layer handles all pointer events (text selection and tooltips)
- the text layer renders the code with syntax highlighting

The more complex wrapping methods are now removed from the codebase, because we can now wrap non formatted text in the relevant layers.

The old implementation had a couple of issues:
The wrapping was done on the already syntax highlighted code. This involved rendering the textual html as html objects, then manipulating those objects, then reading the created html objects as plaintext only to render that textual html later. This process was rather inefficient.
Every re render of a line a code triggered a redo of this process. These re renders happened quite often when the users was making selections, due to the reactive state system. Because of the complex flow described above, Lit could not do any smart optimizations when re rendering. This slowed down the page and forced me to use rather complex caching functions.
Another issue which was quite annoying is that the `wavy underline` styling in chrome was split in different parts for each separate span in the marked text. Because the syntax highlighting always used different spans, the only solution was to apply the way underline on hidden text without the syntax highlighting.

All the above issues should now be resolved.

Closes #4654


app/assets/javascripts/components/annotations/code_listing_row.ts
```diff
@@ -164,13 +77,7 @@ export class CodeListingRow extends i18nMixin(ShadowlessLitElement) {
...
                 </td>
                 <td class="rouge-code">
-                    ${this.fullLineAnnotations.length > 0 ? html`
-                        <d-annotation-marker style="width: 100%; display: block" .annotations=${this.fullLineAnnotations}>
-                            <pre class="code-line ${this.codeLineClass}">${unsafeHTML(this.wrappedCode)}</pre>
-                        </d-annotation-marker>
-                    ` : html`
-                        <pre class="code-line ${this.codeLineClass}">${unsafeHTML(this.wrappedCode)}</pre>
-                    `}
+                    <d-code-layers .row=${this.row} .renderedCode=${this.renderedCode}></d-code-layers>
```
bmesuere, line 80: Is "code-layers" the best name for this component? The layer part is an implementation detail and shouldn't be relevant here. Maybe code-listing is better?

app/assets/javascripts/components/annotations/selection_layer.ts
```diff
@@ -0,0 +1,53 @@
...
+        }
+
+        const start = userAnnotationState.selectedRange.column ?? 0;
+        const end = userAnnotationState.selectedRange.columns ? start + userAnnotationState.selectedRange.columns : this.code.length;
+
+        return html`<pre class="code-line selection-layer"><span></span>${
+            this.code.substring(0, start)}${
+            html`<span class="${this.markingClass}">${this.code.substring(start, end)}</span>`}${
+            this.code.substring(end)
+        }</pre>`;
```
bmesuere, line 51: Is there a reason for the empty span and the strange indentation? The opening `${` tags are on the end of the previous line instead of the beginning.

bmesuere (changes requested):
👍 looks good. I propose we go over this together once you're back in Ghent.

app/assets/javascripts/components/annotations/selection_layer.ts
jorg-vr: empty span is a hack. If the `<pre>` is empty it's height slightly deviates from a `pre` element with content. I found that adding an empty span resolves this problem. I have not yet found another better solution. (can't set a fixed height because because it could span multiple lines. And line height is already set, but ignored when empty)
I should probably add that as a comment

The `${` start on different lines to be able to split this code in multiple lines. Within the `<pre>` tag every space in the html is rendered (as expected for code rendering). Spaces and line breaks within `${` are ignored as that part is parsed as javascript

app/assets/stylesheets/components/code_listing.css.scss
```diff
@@ -288,12 +251,62 @@ $annotation-user-background-intense: color.mix($annotation-user, $background, 20
...
+          user-select: none;
+        }
+
+        .tooltip-layer {
+          position: absolute;
+          width: 100%;
+          height: 100%;
+          top: 0;
+          color: transparent;
+        }
```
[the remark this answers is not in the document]
jorg-vr: I have grouped the css. But I am not really a fan of the result as it has become less readable and only three lines shorter (Most things are only common for three out of four classes, with always different sets of three)

chvp (changes requested):
There's something wrong with the rendering in my Firefox: screenshot_2023-07-31-103231

chvp (changes requested):
The styling of the code is not monospace.

jorg-vr:
That was a bug from #4745, fixed now

chvp (approved):

bmesuere (approved):

PR #4847, opened 2023-07-27
Avoid crash when calling a delayed job on a deleted question

### LINKED ISSUE #4696: Delayed job to reset question state fails
If the question gets removed before the delayed job is executed, the delayed job crashes. Should be fixed by a simple if test.

jorg-vr:
Closes #4696


bmesuere (approved):
Can we test this easily?

niknetniko (approved):

PR #4848, opened 2023-07-28
Prevent crash on series creation without course

bmesuere:
Reverts dodona-edu/dodona#4617 because of the comment in https://github.com/dodona-edu/dodona/pull/4617#discussion_r1277311420

### DIFF
--- a/app/controllers/series_controller.rb
+++ b/app/controllers/series_controller.rb
@@ -108,7 +108,7 @@ class SeriesController < ApplicationController
   # POST /series.json
   def create
     @series = Series.new(permitted_attributes(Series))
-    authorize @series, :create?
+    authorize @series
     respond_to do |format|
       if @series.save
         format.html { redirect_to edit_series_path(@series), notice: I18n.t('controllers.created', model: Series.model_name.human) }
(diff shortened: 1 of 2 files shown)


freyavs:
just for info and checking if this is okay: by authorizing on series the error is now 422 `"course": ["must exist"]`

jorg-vr (approved):

jorg-vr:
You could maybe leave the test (Without the text check)

app/controllers/series_controller.rb
```diff
@@ -108,13 +108,9 @@ def edit
   # POST /series.json
   def create
     @series = Series.new(permitted_attributes(Series))
-    if @series.course.present?
-      authorize @series.course, :add_series?
-    else
-      @series.errors.add(:course_id, I18n.t('errors.messages.blank'))
-    end
+    authorize @series, :create?
```
niknetniko, line 111:
```suggestion
    authorize @series
```

If you don't pass the second parameter, it is inferred from the action's name (in this case we are in the `create` action of the controller, so it should infer `create?`.

niknetniko (changes requested):

PR #4850, opened 2023-07-30
question_table.ts: remove jquery

freyavs:
This pull request removes jQuery from question_table.ts

This is progress on #3590

### DIFF
--- a/app/assets/javascripts/question_table.ts
+++ b/app/assets/javascripts/question_table.ts
@@ -78,7 +78,7 @@ export class QuestionTable {
         fetch(this.getRefreshUrl(), {
             headers: {
                 "accept": "text/javascript",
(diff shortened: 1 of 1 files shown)


bmesuere (approved):

niknetniko (approved):

app/assets/javascripts/question_table.ts
```diff
@@ -78,7 +78,7 @@ export class QuestionTable {
         fetch(this.getRefreshUrl(), {
```
jorg-vr, line 78: It would be better to use the fetch function from util here. This function already sets some of the headers that we need for every request, which avoids duplication

jorg-vr (changes requested):

jorg-vr (approved):

PR #4851, opened 2023-07-31
rails views: start removing jQuery

freyavs:
This pull request removes jQuery in the rails views where html is being set for activities, courses, users, ... (there's still some views left with .html being used)

This is progress on #3590


app/assets/javascripts/util.js
```diff
@@ -208,6 +208,26 @@ function getParentByClassName(element, classNames) {
     return null;
 }
 
+/**
+ * When setting innerHTML, code inside the <script> tag doesn't get executed.
+ * This function solves that problem.
+ */
+function setInnerHTML(el, html) {
```
bmesuere, line 215: - This will need to be moved to utilities once #4849 is merged.
- Since the main thing that is different from simply using `innerHTML` is the execution of scripts, I would like to have this reflected in the name (e.g. `setHTMLExecuteScripts` or something). Alternatively, the initial innerHTML call is removed here and added where the function is called, followed by explicitly calling this function (which could then be called `executeScripts` or something like that.

@chvp @jorg-vr What is your opinion?

jorg-vr: I agree with the name change to make script execution explicit
Although I prefer the shorter name of your second suggestion, I would go for the first option you suggested to avoid  code duplication in all js rails views

app/assets/javascripts/util.js
```diff
@@ -208,6 +208,26 @@ function getParentByClassName(element, classNames) {
...
+            newScriptEl.setAttribute(attr.name, attr.value);
+        });
+        const scriptText = document.createTextNode(oldScriptEl.innerHTML);
+        newScriptEl.appendChild(scriptText);
+        oldScriptEl.parentNode.replaceChild(newScriptEl, oldScriptEl);
+    });
+}
+
+// insert `cached` function here after move to typescript
+// the function is currently in `app/assets/javascripts/mark.ts`
```
jorg-vr, line 229: As mark.ts was removed, this comment should not be reintroduced here (Probably overlooked in a merge with main

app/assets/javascripts/util.js
chvp: Yeah, I also agree (and would also keep the `innerHTML` here).

jorg-vr: yes

And fix the merge conflicts by moving this to utilities.ts

freyavs:
Added more to this PR because I was finishing my last hours and needed the function added here

jorg-vr (approved):

chvp (approved):

bmesuere (approved):

PR #4862, opened 2023-08-02
Reintroduce highlighting annotations on hover

jorg-vr:
This pull request reintroduces highlighting of hovered annotations which was removed in #4771

I have taken a slightly different approach in the implementation, making it a bit more object orientated.
IsHovered is now a property of an annotation, instead of a global map.

To make this property reactive, using the existing state mechanism, I had to make annotations proper instances of a UserAnnotation or MachineAnnotation class. (Instead of just parsed json objects)

### DIFF
--- a/app/assets/javascripts/components/annotations/annotation_marker.ts
+++ b/app/assets/javascripts/components/annotations/annotation_marker.ts
@@ -6,6 +6,7 @@ import {
     isUserAnnotation
 } from "state/Annotations";
 import { MachineAnnotation } from "state/MachineAnnotations";
+import { StateController } from "state/state_system/StateController";
 /**
  * A marker that styles the slotted content based on the relevant annotations.
  * It applies a background color to user annotations and a wavy underline to machine annotations.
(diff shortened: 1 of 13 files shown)


app/assets/javascripts/components/annotations/annotation_marker.ts
```diff
@@ -35,24 +38,35 @@ export class AnnotationMarker extends LitElement {
...
+        return this.annotations.sort( (a, b) => {
+            if (a.isHovered && !b.isHovered) {
+                return -1;
+            }
+
+            if (b.isHovered && !a.isHovered) {
+                return 1;
+            }
+
+            return compareAnnotationOrders(a, b);
```
bmesuere, line 63: can you add a little comment here?

app/assets/javascripts/components/annotations/user_annotation.ts
```diff
@@ -175,7 +175,9 @@ export class UserAnnotation extends i18nMixin(ShadowlessLitElement) {
 
     render(): TemplateResult {
         return html`
-            <div class="annotation ${this.data.type == "annotation" ? "user" : "question"}">
+            <div class="annotation ${this.data.type == "annotation" ? "user" : "question"}"
```
bmesuere, line 178:
```suggestion
            <div class="annotation ${this.data.type === "annotation" ? "user" : "question"}"
```

bmesuere (approved):

niknetniko (approved):

PR #4868, opened 2023-08-07
Allow HOGent to sign in without SAML

jorg-vr:
This pull request removes the account verification step for linking HOGent accounts. 
We do this because the SAML sign in method is currently broken.

The main sign in button now redirects to office365 as well.

I would prefer not to this. so we'll wait a bit to see if the HOGent can fix the SAML integration fast.


chvp (approved):
Code LGTM, but I agree with waiting a bit to see if HOGent can quickly fix it on their side.

niknetniko (approved):

jorg-vr:
The SAML connection has been fixed thus this pr is no longer required.

PR #4887, opened 2023-08-11
Make saved annotations filterable and sortable

jorg-vr:
This pull request makes saved_annotations sortable and filterable on the list page.

image

It also contains a new rails concern `sortable` that avoids duplicating  complex code fragment for the 4th time.

Because of #4886 I could also simplify the `annotations_list` component, as the small layout with prefiltering is no longer used.

I also updated the table content:
- exercise and course no longer link to the item but act as easy filter buttons.
- number of usages is a separate column to make it more easily sortable.

- [x] fix flickering similar to #4701 

Part of #4564


app/controllers/activities_controller.rb
```diff
@@ -30,14 +30,8 @@ class ActivitiesController < ApplicationController
...
 
-  has_scope :order_by, using: %i[column direction], type: :hash do |_controller, scope, value|
-    column, direction = value
-    if %w[ASC DESC].include?(direction) && %w[name popularity].include?(column)
-      scope.send "order_by_#{column}", direction
-    else
-      scope
-    end
-  end
+  include Sortable
```
niknetniko, line 33: Is there a particular reason to include the concern here instead of at the top of the controller with the others?

app/controllers/submissions_controller.rb
```diff
@@ -22,14 +22,8 @@ class SubmissionsController < ApplicationController
...
 
-  has_scope :order_by, using: %i[column direction], type: :hash do |_controller, scope, value|
-    column, direction = value
-    if %w[ASC DESC].include?(direction) && %w[user exercise created_at status].include?(column)
-      scope.send "order_by_#{column}", direction
-    else
-      scope
-    end
-  end
+  include Sortable
```
niknetniko, line 25: Same here

niknetniko (changes requested):

bmesuere (approved):

app/controllers/activities_controller.rb
jorg-vr: Not any more with final implementation, moved them now

niknetniko (approved):

PR #4902, opened 2023-08-16
Avoid 403 errors when a user cannot access linked saved annotations

jorg-vr:
This pull request fixes console error logs when visiting a page with saved annotations you cannot see.

### DIFF
--- a/app/assets/javascripts/state/SavedAnnotations.ts
+++ b/app/assets/javascripts/state/SavedAnnotations.ts
@@ -43,6 +43,11 @@ class SavedAnnotationState extends State {
     private async fetch(id: number): Promise<SavedAnnotation> {
         const url = `${URL}/${id}.json`;
         const response = await fetch(url);
+
+        if (response.status === 403) {
+            return undefined;
+        }
+
         this.byId.set(id, await response.json());
         return this.byId.get(id);
     }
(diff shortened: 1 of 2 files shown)


app/views/annotations/_annotation.json.jbuilder
```diff
@@ -41,3 +41,5 @@ json.responses annotation.responses do |response|
   json.partial! response, as: :annotation
 end
 json.thread_root_id annotation.thread_root_id
+
+json.saved_annotation_id annotation.saved_annotation_id if annotation.saved_annotation.present? && SavedAnnotationPolicy.new(current_user, annotation.saved_annotation).show?
```
bmesuere, line 45: can you add a comment here (as with the other checks in this view),

app/assets/javascripts/state/SavedAnnotations.ts
```diff
@@ -43,6 +43,11 @@ class SavedAnnotationState extends State {
     private async fetch(id: number): Promise<SavedAnnotation> {
         const url = `${URL}/${id}.json`;
         const response = await fetch(url);
+
+        if (response.status === 403) {
+            return undefined;
+        }
```
bmesuere, line 49: This is a bit odd as it would now fail silently? We should at least log something to the console as this shouldn't happen.

bmesuere (changes requested):

app/assets/javascripts/state/SavedAnnotations.ts
jorg-vr: The 403 error is still shown in the console, but the error parsing `response.json()` is avoided.

But can leave this out as the other changes fixes the whole issue

bmesuere (approved):

niknetniko (approved):

PR #4917, opened 2023-08-22
Rotate office365 OAuth client secret for 2023-2024

rien:
This pull request adds a new client secret for Office365 in both the staging and production credentials. It will expire in 1 year and 1 month (2024-09-21) and we will get notified by email a month in advance.


rien:
Confirmed that this works on Naos

niknetniko (approved):

PR #4921, opened 2023-08-28
Constrain annotations to start within the length of the line

### LINKED ISSUE #4920: Client-side processing of annotations sometimes eats code
See e.g. https://dodona.be/nl/submissions/14682928/. On line 19 there is an `else`. This can also be seen in the raw HTML returned from the server. This `else` is however not visible in the code view and also can't be found with "Inspect element".

chvp:
I opted to just highlight the whole line if we encounter one of these annotations. It could of course also be an option to try to highlight past the last character of a line.

Closes #4920.


niknetniko (approved):

bmesuere (approved):
This is used twice, so maybe extract it to a separate hasValidColumn function to make the code more readable?

PR #4933, opened 2023-09-04
Replace Ace Editor with CodeMirror

### LINKED ISSUE #4585: ACE Editor does not support latest JavaScript features
For example, from https://dodona.ugent.be/nl/submissions/13861144/:

image

They seem to use JSHint for this, and there should be an option to use a higher version (https://github.com/ajaxorg/ace/issues/3160). Alternatively, we could disable the editor linting completely (https://stackoverflow.com/questions/12886857/how-can-i-disable-the-syntax-checker-in-ace-editor), and leave it for the judges.

niknetniko:
This pull request replaces the Ace Editor with [CodeMirror](https://codemirror.net/) (which is a commonly-used text editing component for the web).

While functionally equivalent for end-users, it does have some benefits:

- No longer uses Sprockets
- Ready for use within WebComponents

On the other hand, it does need a bit more code to:

- Set up the language support (since the set of first-party language modules is smaller)
- Set up the theme. It applies the Rouge CSS classes to the code, so it uses the same theme. However, sometimes CodeMirror and Rouge don't agree on what a particular piece of code is, so I foresee that we might need to tweak the theme a bit sometimes.

Closes #4585 as a side effect.


chvp (approved):

app/assets/javascripts/editor.ts
```diff
@@ -0,0 +1,221 @@
...
+
+// A custom theme for CodeMirror that applies the same CSS as Rouge does,
+// meaning we can use our existing themes.
+const rougeStyle = HighlightStyle.define([
+    { tag: tags.comment, class: "c" },
+    { tag: tags.lineComment, class: "c" },
+    { tag: tags.blockComment, class: "cm" },
+    { tag: tags.docComment, class: "cs" },
+    { tag: tags.name, class: "n" },
+    { tag: tags.variableName, class: "nv" },
```
jorg-vr, line 36:
```suggestion
    { tag: tags.variableName, class: "n" },
```

It bothered me that a lot of code was italic, which was not the case in the rouge code rendering. I found that rouge uses `n` or `nf` instead of `nv` as class for the same variables.

Another solution for this might be to remove the italic styling from `nv`

eg:
image
image
vs:
image
image

app/assets/javascripts/editor.ts
```diff
@@ -0,0 +1,221 @@
...
+} from "@codemirror/view";
+import { tags } from "@lezer/highlight";
+import { Extension } from "@codemirror/state";
+
+declare type EditorEventHandler = (event: FocusEvent, view: EditorView) => boolean | void;
+
+
+// A custom theme for CodeMirror that applies the same CSS as Rouge does,
+// meaning we can use our existing themes.
+const rougeStyle = HighlightStyle.define([
```
jorg-vr, line 30: Could this also be used for the papyros code editor? (as it also uses code mirror)
This would greatly improve styling

jorg-vr (changes requested):
Thank you for replacing the legacy editor!

app/assets/javascripts/editor.ts
niknetniko: Since Papyros sets up its own editor, this is no small task: we would need to either hardcode a "dodona" mode (which isn't nice) or change the API to support configuring the CodeMirror instance used by Papyros (which is a lot of work).

jorg-vr: Can't we make the 'dodona mode' the default for papyros? ANd just make their default styling consistent.
I'll have a look at this at some later point. No need to fix that for this pr

app/assets/javascripts/editor.ts
niknetniko: I removed the italic from `nv`, as it felt like a better solution than to make the css less specific.

jorg-vr (approved):

PR #4936, opened 2023-09-04
Add a few RuboCop extensions

niknetniko:
This pull request adds three more RuboCop extensions, as suggested by RuboCop and people.


chvp (changes requested):
`rubocop-minitest` exists as well, maybe we want to enable that too?

bmesuere (approved):

bmesuere (approved):
I don't know how often they update, but it might be useful to group the dependabot PRs for these.

chvp (approved):

PR #4965, opened 2023-09-13
Allow creating saved annotations by searching annotations

jorg-vr:
This pull request tries to remove the need for anticipation of reuse when saving annotations. A saved annotation can easily be created by searching the list of existing annotations.

Clicking the link under the saved annotations searchfield:
image
Brings the user the the prefiltered saved annotations list page
image
This page now has a `+` button to create new saved annotations. This + button links to the prefiltered list of annotations:
image
These can of course be filtered or sorted as desired.

Here a user can easily save the annotation (I reused the existing modal for this)
image

The saved annotation detail page now also uses this list view of annotations instead of a list of submission:
image

This pr also contains:
- bug fix: saved annotation titles sometimes incorrectly detected a duplicate
- chore: the tracking of userId in the frontend to filter saved annotations is removed. The policy already always filters by current user.

Part of #4564


bmesuere (approved):
I prefer the outlined button in the table instead of a primary

app/assets/javascripts/components/saved_annotations/new_saved_annotation.ts
```diff
@@ -4,10 +4,10 @@
 import { SavedAnnotation, savedAnnotationState } from "state/SavedAnnotations";
 import "./saved_annotation_form";
 import { modalMixin } from "components/modal_mixin";
-import { isBetaCourse } from "saved_annotation_beta";
 import { exerciseState } from "state/Exercises";
 import { courseState } from "state/Courses";
-import { userState } from "state/Users";
+import { searchQueryState } from "state/SearchQuery";
+import { updateURLParameter } from "utilities";
```
niknetniko, line 10: These two imports don't seem to be used

niknetniko (changes requested):
As mentioned IRL, the button to save a comment doesn't seem to be translated sometimes.

(and we'll ignore the 403 as the beta will go away soon)

niknetniko (approved):

PR #4970, opened 2023-09-15
Replace Ace Editor with CodeMirror 6

### LINKED ISSUE #4585: ACE Editor does not support latest JavaScript features

niknetniko:
This reverts commit 98c93fec44dca91bbb548b83528726fb0a089ec8, and thus re-introduces #4933.

---

This pull request replaces the Ace Editor with [CodeMirror](https://codemirror.net/) (which is a commonly-used text editing component for the web).

Feature-wise, codemirror is mostly equivalent. There are some changes:
- autocomplete is now shown by default instead of only on ctrl + space
- ace had linting for Javascript, but this caused confusion (#4585). It is now disabled (also for other languages)
- the syntax highlighting is slightly different, but now equivalent to rouge which is used on all other places on Dodona
- the font size is slightly larger

There are a number of other benefits:

- Uses javascript packages instead of ruby gems which rely on sprockets magic
- Ready for use within WebComponents

On the other hand, it does need a bit more code to:

- Set up the language support (since the set of first-party language modules is smaller)
- Set up the theme. It applies the Rouge CSS classes to the code, so it uses the same theme. However, sometimes CodeMirror and Rouge don't agree on what a particular piece of code is, so I foresee that we might need to tweak the theme a bit sometimes.

Closes #4585 as a side effect.


jorg-vr:
The potential problem might be dynamic loading of javascript

If codemirror loads certain packages only if the language requires them, it needs to use the correct paths.
The assets paths are still managed by rails sprockets, so the issue might be there.

I remember you also saying that on the broken pages there was a javascript fetch error

jorg-vr:
confirmed bug to be related to #5082
image

So fix might have fixed this

jorg-vr (approved):

app/assets/javascripts/editor.ts
```diff
@@ -0,0 +1,221 @@
...
+        name: "R",
+        alias: ["rlang"],
+        extensions: ["r"],
+        load() {
+            return import("codemirror-lang-r").then(m => m.r());
+        }
+    }),
+    LanguageDescription.of({
+        name: "Prolog",
+        alias: ["rlang"],
```
bmesuere, line 146: This shouldn't be rlang?

bmesuere (changes requested):
- [x] Autocomplete doesn't seem to work. E.g. html autocomplete in the docs https://codemirror.net/examples/autocompletion/ vs on https://naos.dodona.be/nl/activities/1412931994/

- [ ] Font size is a little bigger, but not sure if this is a problem.

- [x] We seem to miss the highlighting of functions compared to ace.

- [x] Ace had basic linting (in javascript), this is missing now. Not sure if this is a problem (since the linting messages could be different from the ones in the judge and thus confusing), but maybe mention it in the PR.

- [x] I also think the python indentation is different (or at least not very smart). In a solution that used 4 spaces, it added just 2 when adding a new line.

- [ ] The active line isn't marked anymore.

app/assets/javascripts/exercise.ts
```diff
@@ -136,14 +138,14 @@ function initExerciseDescription(): void {
...
+async function initExerciseShow(exerciseId: number, programmingLanguage: string, loggedIn: boolean, editorShown: boolean, courseId: number, _deadline: string, baseSubmissionsUrl: string, boilerplate: string): Promise<void> {
+    let editor: EditorView;
     let lastSubmission: string;
     let lastTimeout: number;
 
-    function init(): void {
+    async function init(): Promise<void> {
         if (editorShown) {
-            initEditor();
+            await initEditor();
```
bmesuere, line 148: I don't think the await is needed on this line since the next few functions don't depend on it?
If this does stuff in a worker (or loads additional files), the await itself can probably move down a few lines to do things more async.

jorg-vr:
Changes:
- autocomplete now works and is applied by default
- font size > I prefer the larger size, but did make the line numbers a bit smaller
- The python lexer does not indicate builtins in the same way as  ace, so we canot color them specifically. For the fans of a bit more color, I did change the highlighting for functions to yellow in both the editor and rouge (to keep it consistent)
- Added a comment that linting is removed in pr
- I made the default indent 4 spaces. This is the most common for python and python is our most used language. I could not find a smart detection algorithm in the editor. We could write such an algorithm ourselves, or provide a different default for each language if desired
- I reintroduced active line marking. The color has to be a partially transparant version of the slection color to work well with selected code

bmesuere (approved):

PR #4987, opened 2023-09-19
Fix docker images not being shut down

jorg-vr:
This pull request fixes infinitely running docker images.

It reverts changes made in #4979, suggested by rubocop, which were incorrect. The `any` function does not exist on `Docker::Container`
image

The rubocop issue is already reported https://github.com/rubocop/rubocop-rails/issues/1124
With a potential fix https://github.com/rubocop/rubocop-rails/pull/1127


jorg-vr:
I have double checked all other removals of `all` in https://github.com/dodona-edu/dodona/pull/4979 and they all seem valid. Thus no other changes seem to be required

chvp (approved):

PR #4989, opened 2023-09-20
Make wavy underline markings smaller

### LINKED ISSUE #4943: Avoid using wavy text decoration for multiline annotations
Multiline annotations from the linter use the wavy text decoration, which hampers readability of the code fragment. Might be better to use the same background-color annotations as with manual multiline annotations. Using wavy text decoration for inline annotations with background highlighting for multiline annotations also seems to mix better if both are combined in the same code fragment.

image

jorg-vr:
This pull request replaces wavy underline with custom svg styling. I copied the svg from codemirror.

The background image has to be defined within the theme as we cannot use css variables inside the image. I wrote a mixin to avoid a lot of duplicate sass.

The styles are also moved to static css classes, which should render more efficiently.

### Old
image

### New
image

### Old
image

### New
image

Closes #4943

### DIFF
--- a/app/assets/javascripts/components/annotations/annotation_marker.ts
+++ b/app/assets/javascripts/components/annotations/annotation_marker.ts
@@ -1,5 +1,5 @@
 import { customElement, property } from "lit/decorators.js";
-import { html, LitElement, TemplateResult } from "lit";
+import { css, html, LitElement, TemplateResult, CSSResultGroup, unsafeCSS } from "lit";
 import {
     Annotation,
     compareAnnotationOrders,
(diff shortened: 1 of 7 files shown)


bmesuere (changes requested):
I can live with using an image mask for this, but I don't think this is implemented in the most efficient way.

Right now, the style is added inline to each instance of the custom element (it is added to the template). I would expect it to be in a general css file, or in the static styles of the element.

jorg-vr:
The styling is also broken in combination with comments
image

I knew about the inefficiency of inline style, but this problem has indeed become worse now that we use svg images

chvp (approved):

PR #4990, opened 2023-09-20
Fix annotating part of a line

jorg-vr:
This pull request fixes creating an annotation for a subselection of a line.

this was broken since #4959


bmesuere (approved):
I agree with the fix, but it still doesn't look very robust and is hard to test.

niknetniko (approved):

PR #4995, opened 2023-09-22
Fix refreshing of course members page after actions

### LINKED ISSUE #4968: Delete student from course
When deleting a student from a course as a course admin, the table is not refreshed (student is removed though).

jorg-vr:
Closes #4968

### DIFF
--- a/app/views/courses/_reload_users.js.erb
+++ b/app/views/courses/_reload_users.js.erb
@@ -3,7 +3,7 @@
   new dodona.Toast("<%= toast %>");
 <% end %>
 <% count = @course.pending_members.count %>
-if(document.getElementById("pending-count")) {
+if (document.getElementById("pending-count")) {
     document.getElementById("pending-count").innerText = "<%= count if count.nonzero? %>";
 }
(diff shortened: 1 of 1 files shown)


app/views/courses/_reload_users.js.erb
```diff
@@ -3,5 +3,7 @@
   new dodona.Toast("<%= toast %>");
 <% end %>
 <% count = @course.pending_members.count %>
-document.getElementById("pending-count").innerText = "<%= count if count.nonzero? %>";
+if(document.getElementById("pending-count")) {
```
chvp, line 6:
```suggestion
if (document.getElementById("pending-count")) {
```

chvp (approved):

bmesuere (approved):
I think the problem here was that getElementById returned null which caused innerText to fail, which in turn stopped the execution of the rest of the code? I expect we'll have more issues like this due to the jquery migration.

Too bad you can't use optional chaining in the left hand side of an assignment and that there is no way to set innerText using a function.

PR #4996, opened 2023-09-22
Fix infinite progress bar when filtering students in an evaluation

### LINKED ISSUE #4918: Evaluation search acts as if it keeps searching
When searching for a student in an evaluation, the search indicator keeps spinning even after the search has completed.

image

jorg-vr:
Closes #4918


bmesuere (commented):
This feels a bit like a hack. Why isn't the element present in the first place?

jorg-vr:
It is context depended. If there is scoring present, there will be a tab with scores of users.
If no scoring is made, this tab is not present. Which makes this feature crash.

If you prefer I can probably also fix this by a rails template check? (But that risks an edge case where the tab was not present before, but should be present now, which could create a bug....)

jorg-vr:
The issue is very similar to #4996

niknetniko (changes requested):
I we want to fix this like this (I have no opinion), could we do only one `querySelector`?

niknetniko (approved):

bmesuere (approved):

PR #4997, opened 2023-09-22
Remove browser outline styling for focus-within on nav-tabs

### LINKED ISSUE #4569: Blue outline on focussed tab
Our tabs sometimes show a blue line at the right:
image

The line disappears when you interact with the page and seems to be consistently present when we switch tabs programmatically.

I think this is caused by `:focus-visible` on the `a` inside the `li`. At least, I can reproduce it when I force that state using the dev tools. Setting the outline to none for that pseudo class will probably solve this issue.

jorg-vr:
This pull request removes the outline styling applied by some browsers on tab links
focus-within is specifically applied by browsers to improve ux (https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible). But it does not seem relevant in our case as we have custom styling to make UX clear for tabs.

Closes #4569

### DIFF
--- a/app/assets/stylesheets/base.css.scss
+++ b/app/assets/stylesheets/base.css.scss
@@ -115,7 +115,3 @@ figcaption {
   justify-content: space-between;
   flex-direction: row;
 }
-
-.nav.nav-tabs a:focus-visible {
-  outline: none;
-}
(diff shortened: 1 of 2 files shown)


chvp (approved):

bmesuere (commented):
This might be useful in other cases where we don't expect it. Can't we just disable it in our tab bar?

jorg-vr:
I'll make it more specific

bmesuere (approved):

PR #4998, opened 2023-09-22
Add draft flag to activities

### LINKED ISSUE #4845: Add draft state for activities
When people are creating new exercises, this often doesn't work from the start. This is not a problem, but we can try to improve the experience. This issue suggests adding a "draft" flag to the exercise config. Based on this flag, we can adapt our behaviour:

- If a submission results in an internal error, this triggers an email and slack mention. We could disable this for draft exercises.
- We can limit the visibility of draft exercises, i.e. not show them to everyone. We do have to think about the use case where an exercise is drafted by multiple people at the same time.
- If possible, show more information when something goes wrong.
- On the top of the exercise description, show a clear warning that the exercise is in draft mode (and can thus not be used by others) and give the option to disable draft mode from Dodona itself. Maybe also add a more prominent edit link to such exercises.
- If a draft exercise is removed, actually remove it from the Dodona database
- When creating an exercise, it is not super easy to find said exercise to try it out. I think most people add it to a (dummy) course. We could add a list of "your draft exercises" to the home page.
- Add the draft flag to all our example exercises so we don't get tens of copies of the same exercise.

jorg-vr:
This pull request introduces the concept of draft activities.

## What is a draft activity?
Every newly added activity is considered a draft activity until they are published. 

The idea is that an activity stays in draft mode until its creator greenlights it manually using the Dodona UI.

## What is the effect of a draft activity?
The concept of draft activities serves several goals.

### Prevent clutter
Draft activities are only visible for repository owners and course admins for courses with the activity. This prevents that multiple copies of our example exercises are present in the global database for everyone to see.

### Reduce false-positive error messages
The Dodona admins are notified of severe errors (`internal error`) during the execution of a submission. When a teacher is creating a new exercise and experimenting with adding tests, these errors are quite common. Because of this, the Dodona admins often ignore these messages, causing real issues to be unnoticed. When an exercise is still in draft mode, we will no longer notify the Dodona admins.

### Improve discoverability
When you add a new exercise to Dodona, it is not always easy to find that exercise and try it yourself. We will now list all draft exercises of a user on the home page, making these easier to find.

## Future additions
This is just a first set of changes. In the future, we might add additional features for draft activities:

- If a draft activity is removed from the repository, we may automatically remove it from Dodona instead of just marking it as removed.
- If a teacher marks a draft as finalized, we may explicitly ask if they want to make the exercise public or private. This can further reduce the number of (unwanted) public exercises.
- Provide extra debug information. If the judge crashes due to an internal error, some debug information is logged for the Dodona admins. We may also show this information to teachers if an exercise is in draft mode.

---

image
image
In the below layout, the activity status icon is also moved to avoid a to long list of symbols
image
image
image
image
image

- [x] Tests were added
- [x] Documentation update can be found at dodona-edu/dodona-edu.github.io#331

Closes #4845


niknetniko (changes requested):
One small edge case I noticed:

- When an existing exercise is changed to draft, the exercise still seems to appear on the home page of students who have submitted to this exercise (although if you click on it, it correctly says you don't have access). Not sure if this needs fixing.

jorg-vr:
The activity is not directly linked if it is not accessible in the recent activities:
image

On some other places (such as submission detail), the link to the activity can still be found, But gives a correct error message.

This behavior is similar to that of series without access our private exercises etc, I would not change it.

I guess it also won't happen often that an existing activity is set to draft after students used it.

pdawyndt:
Maybe it could help if students also see the draft status, which is in contrast with the private status that is completely transparent for students. At least, having the draft status visible for students would explain for them why clicking the exercise results in a "no access", and then students could remind their teachers that exercises are still in draft mode.

niknetniko (approved):
I don't immediately have better suggestions, so more as a mental note, but the "draft mode" and especially "ontwerpmodus" terminology feel a bit weird to me.

pdawyndt:
@niknetniko: would the mental model feel more natural if it were branded as published/unpublished?

config/locales/views/activities/en.yml
```diff
@@ -27,6 +27,10 @@ en:
       labels: New label
       info: Information
       info_description: Go to the information page of this learning activity
+      draft:
+        title: "Draft"
+        toggle-label: "Put activity in draft mode."
```
bmesuere, line 32:
```suggestion
        toggle-label: "Mark activity as draft."
```

config/locales/views/activities/en.yml
```diff
@@ -27,6 +27,10 @@ en:
...
+        help: "Draft mode provides better error logging for teachers and hides the activity from students. Draft mode is activated by default for new activities."
```
bmesuere, line 33:
```suggestion
        help: "Mark activities as draft while you are still working on them."
```
+ add link to documentation page about drafts

config/locales/views/activities/en.yml
```diff
@@ -67,6 +71,7 @@ en:
         mine: "My activities"
         featured: "Featured activities"
         all: "All activities"
+      draft-notice: "This activity is in draft mode. It is not visible for students."
```
bmesuere, line 74:
```suggestion
      draft-notice: "This activity is a draft and thus not visible for students."
```

+ het woord "draft" linken naar de docs (zelfde met andere notice strings in deze PR)

config/locales/views/activities/en.yml
```diff
@@ -97,6 +102,8 @@ en:
       preloaded_info: "We have preloaded your latest submission into the editor."
       preloaded_clear: Clear editor.
       preloaded_restore: Restore the initial code.
+      alert_draft: This activity is in draft mode. It is not visible for students.
```
bmesuere, line 105:
```suggestion
      alert_draft: This activity is a draft and thus not visible for students.
```

config/locales/views/activities/en.yml
```diff
@@ -97,6 +102,8 @@ en:
...
+      alert_draft_edit: Disable draft mode.
```
bmesuere, line 106:
```suggestion
      alert_draft_edit: Publish activity.
```

config/locales/views/activities/nl.yml
```diff
@@ -27,6 +27,10 @@ nl:
       labels: Nieuw label
       info: Informatie
       info_description: Ga naar de informatiepagina van deze leeractiviteit
+      draft:
+        title: "Concept"
+        toggle-label: "Markeer als concept"
+        help: "Wanneer een oefening in concept modus staat, is ze onzichtbaar voor studenten en krijgen leerkrachten betere foutmeldingen."
```
bmesuere, line 33:
```suggestion
        help: "Markeer een oefening als concept zolang je ze nog aan het opstellen bent."
```
+ link naar docs

config/locales/views/activities/nl.yml
```diff
@@ -67,6 +71,7 @@ nl:
         mine: "Mijn activiteiten"
         featured: "Uitgelichte activiteiten"
         all: "Alle activiteiten"
+      draft-notice: "Deze activiteit is een concept. Ze is niet zichtbaar voor studenten."
```
bmesuere, line 74:
```suggestion
      draft-notice: "Deze activiteit is nog een concept. Ze is niet zichtbaar voor studenten."
```

config/locales/views/activities/nl.yml
```diff
@@ -97,6 +102,8 @@ nl:
       preloaded_info: We hebben jouw laatste oplossing ingeladen in de editor.
       preloaded_clear: Maak de editor leeg.
       preloaded_restore: Zet de voorbeeldcode terug.
+      alert_draft: Deze oefening is een concept. Ze is niet zichtbaar voor studenten.
```
bmesuere, line 105:
```suggestion
      alert_draft: Deze oefening is nog een concept. Ze is niet zichtbaar voor studenten.
```

config/locales/views/activities/nl.yml
```diff
@@ -97,6 +102,8 @@ nl:
...
+      alert_draft_edit: Markeer deze oefening als af.
```
bmesuere, line 106:
```suggestion
      alert_draft_edit: Deze oefening publiceren.
```

config/locales/views/pages/nl.yml
```diff
@@ -101,6 +101,8 @@ nl:
       submissions: Ingediende oplossingen
       correct-exercises: Correcte oefeningen
       recent-exercises: Recente oefeningen
+      draft-exercises: Concept oefeningen
+      all-draft-exercises: Alle concept oefeningen
```
bmesuere, line 105: Maybe "ongepubliceerde oefeningen"?

bmesuere (changes requested):

bmesuere (commented):
I've been thinking about this some more. I there a good reason why someone might convert an existing exercise back into draft? Right now, doing so might cause problems if the exercise is used in other courses.

jorg-vr:
When updating an exercise this might be usefull. Eg adding more tests: there might be mistakes in those new tests so while working on them it can be logical to make the exercise temporarily unavailable for students

Teachers of these other course will still have access and see the draft notice

app/models/activity.rb
```diff
@@ -315,19 +319,22 @@ def accessible?(user, course)
       if user&.course_admin? course
         return false unless course.activities.pluck(:id).include? id
       elsif user&.member_of? course
-        return false unless course.accessible_activities.pluck(:id).include? id
-      else
-        return false unless course.visible_activities.pluck(:id).include? id
+        return false if course.accessible_activities.pluck(:id).exclude?(id) || draft?
+      elsif course.visible_activities.pluck(:id).exclude?(id) || draft?
+        return false
```
chvp, line 324: Wouldn't it be better to change the definition of `accessible_activities` and `visible_activities`? (Or at least accessible, visible might make less sense semantically.)

app/views/activities/_form.html.erb
```diff
@@ -1,7 +1,3 @@
-<% content_for :javascripts do %>
-  <% # TODO rename pack to activity %>
-  <% javascript_include_tag 'exercise' %>
-<% end %>
```
chvp, line 4: Why is this suddenly removed?

app/models/activity.rb
```diff
@@ -315,19 +319,22 @@ def accessible?(user, course)
       if user&.course_admin? course
         return false unless course.activities.pluck(:id).include? id
       elsif user&.member_of? course
-        return false unless course.accessible_activities.pluck(:id).include? id
-      else
-        return false unless course.visible_activities.pluck(:id).include? id
+        return false if course.accessible_activities.pluck(:id).exclude?(id) || draft?
```
chvp, line 322: Even if it wouldn't make sense to change the definitions, I would still split this up to be two return statements each with their own condition. This code is already quite complicated and that would make it easier to reason about IMO.

chvp: (And same for the `||` in the `elsif` below.)

bmesuere (commented):
As discussed during the day: if we make draft a one-way setting, we can just store it in the database instead of also in the config file. This would simplify the code a bit.

- The default value in the database can be set to true so we can remove the before_create action and the other logic to add it to the config file and set it for new exercises. We could also set it on creation in the tests and prevent a manual update. The migration would probably be something like
```ruby
add_column :activities, :draft, :boolean, default: true
Activity.reset_column_information # don't know if this is strictly needed
Activity.update_all(draft: false)
```
- The setting can be removed from the form. I would add the alert that is now shown on the show page to the edit and info page as well. `publish` could be added as a separate action to the controller (which would prevent adding a new check that only allows it to be set from true to false)

app/models/activity.rb
jorg-vr: I have removed these draft checks as they are redundant considering the return false if draft below

app/views/activities/_form.html.erb
jorg-vr: This was only imported for `initLabelsEdit` below. This function does not exist thus this was a noop.
Related to https://github.com/dodona-edu/dodona/issues/5055
It can be reintroduced when I work on that issue

jorg-vr:
Still have to add to alert to other pages, will fix.

jorg-vr:
> The migration would probably be something like

I took a slightly different approach: first create column with default false, then change the default to true
This results in the default only being true for new exercises

> `publish` could be added as a separate action to the controller

I prefer the current approach with an extra validation in activity. This keeps our api more default and more RESTfull

app/models/activity.rb
```diff
@@ -56,6 +57,7 @@ class Activity < ApplicationRecord
   has_many :labels, through: :activity_labels
 
   validates :path, uniqueness: { scope: :repository_id, case_sensitive: false }, allow_nil: true
+  validates :draft, inclusion: { in: [false] }, unless: :draft_was
```
bmesuere, line 60: what does this do?

bmesuere (approved):

niknetniko (approved):

app/models/activity.rb
jorg-vr: if not `draft_was`, validate if `draft` is in the list of possible values, with the only possible value being `false` 
`draft_was` is the previous value of `draft`

so if the previous value of `draft` was `false`, only allow `false` as new value

bmesuere: You could argument that this check should be in the controller instead of the model, but not a strong opinion so ok for me.

app/policies/activity_policy.rb
```diff
@@ -75,7 +75,7 @@ def read?
 
   def permitted_attributes
     if update?
-      %i[access name_nl name_en]
+      %i[access name_nl name_en draft]
```
chvp, line 78: You could also check here that `draft` can only be updated if it's currently true. Might be a more clear than the `validates` in the model.

chvp (approved):

PR #5024, opened 2023-10-03
Avoid git race conditions by allowing only one worker

jorg-vr:
This pull request fixes race conditions when multiple repository reprocess jobs are running simultaneously.

Closes #5023 as it replaces it

### DIFF
--- a/config/deploy/production.rb
+++ b/config/deploy/production.rb
@@ -12,7 +12,8 @@ server 'tityos.ugent.be',   user: 'dodona', port: '4840', roles: %w[app worker]
 
 set :delayed_job_pools_per_server,
     'dodona' => {
-      'default,git,statistics,exports,cleaning' => 2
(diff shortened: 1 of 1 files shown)


bmesuere (approved):

chvp (approved):
Ugly, but sure.

bmesuere:
@chvp why do you think this is ugly?
- There isn't a real issue in the first place, the second `git commit` is just a no-op which happens to return a non-zero exit code which in turn triggers the logging of exception we could just as well ignore (unless I'm missing something).
- Setting the workers to 1 guarantees single threaded execution without having to handle threading ourselves
- The git queue is very low traffic and the operations are async by default so halving the working will hardly be noticeable. The fact that we had 2 workers in the first place was arbitrary.
- If for some reason we need concurrent git operations, we can always opt for something more complicated, but to me this change is minimal, low risk and low maintenance.

PR #5050, opened 2023-10-19
Allow python tutor for every judge

jorg-vr:
This pull request allows the python tutor for every judge.

If a test case contains the following:
```json5
"data": {
  "statements": "", // A string containing the "test code"
  "stdin": "", // A string containing stdin
}
```
A link to the python tutor is added.

For pythia `"files": {} // Map of names to their contents and other metadata, ignored for now.` is also still supported, both to init file links and for the tutor.
It is not yet generalized as Tested is not yet ready to support it.
We will add this support in a future pr.

- [x] Documentation update can be found at dodona-edu/dodona-edu.github.io#343
- [x] test changes with Tested


jorg-vr:
@niknetniko you can test your Tested implementation using this pr if you also run https://github.com/dodona-edu/python-tutor-webservice locally.

if you have a better suggestion for the fields I am open to that. Then I will simply add support for both and put the new names in the docs.

niknetniko:
Summary of my changes:

- Add support for this for the partial output
- Move the description in the full format to the context level (which I think is the correct level)


See https://github.com/dodona-edu/universal-judge/pull/449 for the implementation in TESTed.

chvp (approved):

app/javascript/packs/pythia_submission.js
```diff
@@ -1,6 +1,3 @@
-import { initPythiaSubmissionShow } from "pythia_submission.ts";
+import { initFileViewers } from "file_viewer";
 
-window.dodona.initPythiaSubmissionShow = initPythiaSubmissionShow;
-
-// will automatically bind to window.iFrameResize()
-require("iframe-resizer"); // eslint-disable-line no-undef
+window.dodona.initFileViewers = initFileViewers;
```
bmesuere, line 3: Since only the file viewer is left in this pack, wouldn't it be a good time to eliminate it and just add the file viewer init to the general submission pack and the initialisation code to the general feedback table renderer?

bmesuere (approved):

PR #5052, opened 2023-10-19
Run python tutor backend in browser

jorg-vr:
This pull request removes the need for pandora by running the python tutor backend client side.
The hard work is done by https://github.com/dodona-edu/pyodide-trace-library

Except for the passing of parameters, the backend code running in pyodide is exactly the same as the code that used to run on pandora.

I have added/fixed a loading bar in the modal when pyodide is still computing.

The pr also contains changes to the webpack config. This adds a hash to js assets that are indirectly imported. Which fixes caching issues with these files.


bmesuere (changes requested):
Nice work!

Is it possible to delay the loading of pyodide until after a tutor link gets clicked for the first time? Right now, 13MB of files is loaded with the submission table.

If I submit this code to a Python echo exercise on mestra:

```
a = 5
b = 7
c = a + b
print(c)
```

And trigger the tutor, I get a tutor modal with blank values:
image

bmesuere (approved):
- Everything seems to work in Chrome, Firefox and Safari on Mac
- The heavy code only loads on (first) tutor initiation
- The code looks good to me. Maybe 2 more changes for the other repo: add a little comment on the line of the worker call with a link to the webpack docs and move typescript and the node types to a dev dependency.

Very nice work!

chvp (approved):

PR #5079, opened 2023-10-24
added extra motivational messages

tyboro2002:
some more motivational messages are added


jorg-vr:
I actually do agree with @GoedendagMC here, and would be willing to revert to the original simple messages.
@bmesuere , @chvp what do you think?

bmesuere:
@GoedendagMC While you are free to voice your opinion (which you already did in 2 places) and we are open to constructive comments, comments like the one above are not the way to do this. We want to be open and welcoming to new contributors, especially students, and such comment from an anonymous account may do more harm than good.

@tyboro2002 Thank you for your contribution. At this time, we have no need for additional messages. Dodona is used by a very wide range of people (going from kids to civil servants and researchers) and not all features/text/UI are a good fit for everyone. This is why we'll probably differentiate certain features (such as these messages) depending on whether they come from secondary education or not. We currently have no timeline for this.

PR #5082, opened 2023-10-26
Fix assets not available in production

jorg-vr:
This pull request fixes the fetching of undigested files. 

Since https://github.com/rails/sprockets-rails/issues/49#issuecomment-20535134 rails does not support this by default and requires a custom task or gem. We already had such a task, which I updated.

Step 1 was to reproduce the bug on naos.
The relevant setting was  `config.assets.compile = false`
It is not an option to set this to true in production: https://stackoverflow.com/questions/8821864/config-assets-compile-true-in-rails-production-why-not

Step 2 was to fix the precompilation step for these files.
To fix this on naos I executed the `assets:nodigest` on naos.
I updated this task, which was unused, to support glob variables and be rooted in the builds directory.

It is important to keep in mind that these files are not managed by sprockets, so we need to add the hash for cache bumping ourselves.
For this reason I updated the naming scheme in webpack and only copied the files that certainly received a hash by webpack.

- [x] tested on naos

### DIFF
--- a/config/deploy/staging.rb
+++ b/config/deploy/staging.rb
@@ -22,4 +22,13 @@ namespace :deploy do
       end
     end
   end
+  before :publishing, :asset_stuff do
+    on roles :web do
+      within release_path do
+        with rails_env: fetch(:rails_env) do
+          execute :rake, 'assets:nodigest'
+        end
+      end
+    end
+  end
 end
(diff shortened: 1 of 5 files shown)


bmesuere (approved):
Nice reuse of the old functionality 👍

chvp (approved):

PR #5095, opened 2023-10-30
Create task that destroys removed activities

### LINKED ISSUE #4969: Auto remove deleted exercises
If exercises are deleted from a repository, they remain visible in Dodona. This is because there might be submissions which must remain available.

However, if there are no submissions to such exercise, they can safely be removed to reduce clutter in the exercise list.

We used to do this manually once in a while using the rails console on the production server. It would be better to add a utility function to do this + schedule this to run every night or so.

_Originally posted by @bmesuere in https://github.com/dodona-edu/dodona/discussions/4764#discussioncomment-6189458_

jorg-vr:
This pull request adds an active job that permanently destroys removed activities that match all of the following criteria:
  - status is 'removed'
  - updated_at is more than 1 month ago
  - one of the following is true:
    - draft is true (never published)
    - series_memberships is empty and less then 25 submissions and latest submission is more than 1 month ago

This job should be started once and reschedules itself every month.

---

Some relevant stats from naos:
```ruby
> Exercise.where(status: 'removed').where('updated_at < ?', 1.month.ago).count
411
>Exercise.where(status: 'removed').where('updated_at < ?', 1.month.ago).filter { |activity| activity.draft? || ( activity.series_memberships.empty? && ( activity.submissions.empty? || (activity.submissions.count < 25 && activity.submissions.order(:created_at).last.created_at < 1.month.ago))) }.count
366
> ContentPage.where(status: 'removed').where('updated_at < ?', 1.month.ago).count
18
> ContentPage.where(status: 'removed').where('updated_at < ?', 1.month.ago).filter { |activity| activity.draft? || activity.series_memberships.empty? }.count
16
```

- [x] tests are added

Closes #4969 and #4764


niknetniko (approved):
While the code looks good, does this need a test? Or it might be too much work to properly test, I don't know.

config/deploy/production.rb
```diff
@@ -13,7 +13,7 @@
 set :delayed_job_pools_per_server,
     'dodona' => {
       'default,statistics,exports,cleaning' => 2,
-      'git' => 1,
+      'git,cron' => 1,
```
bmesuere, line 16: Is there a reason why this is added to the git worker and not to the other one?

bmesuere (approved):

config/deploy/production.rb
jorg-vr: no

app/jobs/remove_activities_job.rb
```diff
@@ -0,0 +1,42 @@
...
+  queue_as :cron
+
+  def perform
+    ContentPage.where(status: 'removed').where('updated_at < ?', 1.month.ago).find_each do |activity|
+      if activity.draft? || activity.series_memberships.empty?
+        # destroy series memberships first explicitly, as they are dependent: :restrict_with_error
+        activity.series_memberships.destroy_all
+
+        activity.destroy
+      end
```
chvp, line 21: Will this work if the content page has been marked as read? And if so, do we want it to?

app/jobs/remove_activities_job.rb
```diff
@@ -0,0 +1,42 @@
...
+    end
+
+    Exercise.where(status: 'removed').where('updated_at < ?', 1.month.ago).find_each do |activity|
+      if activity.draft? ||
+         (activity.series_memberships.empty? &&
+           (activity.submissions.empty? ||
+             (activity.submissions.count < 25 && activity.submissions.reorder(:created_at).last.created_at < 1.month.ago)))
```
chvp, line 28: This is quite a complex `if`. Can it be simplified with `next if/unless ...` syntax?

app/jobs/remove_activities_job.rb
```diff
@@ -0,0 +1,42 @@
...
+
+        # destroy series memberships first explicitly, as they are dependent: :restrict_with_error
+        activity.series_memberships.destroy_all
+
+        activity.destroy
+      end
+    end
+
+    # rerun this job in 1 month
+    RemoveActivitiesJob.set(wait: 1.month).perform_later
```
chvp, line 40: There is also [delayed_cron_job](https://github.com/codez/delayed_cron_job) that takes care of this, but we might not want to introduce a dependency to replace one line.

config/deploy/production.rb
chvp: I would give a more semantically clear name to this queue. Maybe the job can even be scheduled on the `cleaning` queue?

app/jobs/remove_activities_job.rb
jorg-vr: It works with activity_read_states (see the first test)

We could also add a limit of at most 25 activity_read_states, but as a lot less information is lost in the case of the activity_read_states compared to submissions, I did not care about it

app/jobs/remove_activities_job.rb
jorg-vr: I have no preference, do you want me to replace this functionality?

bmesuere: I would leave the code as-is for now. If we add more similar tasks later, we can always add the gem.

chvp (approved):

Issue #5101, opened 2023-11-03
Rework the feedback table


bmesuere:
The existing design of the Dodona feedback table has remained unchanged since its initial release. However, several challenges have emerged over time that the current layout does not adequately address:
- The debugger is hidden and there is no easy way to add a button in the current design.
- In cases where there are numerous tests with only a few failures, it becomes difficult to quickly identify the errors.
- There is no current system for numbering the tests, which would aid in referencing them individually. (also see #357)
- Sometimes, the indication of correctness relies solely on color, which can be problematic for users with color vision deficiencies.

The main obstacle in redesigning the feedback table is its inherent flexibility, which complicates the design process. The feedback is structured across several layers: tab, group, testcase, and test. A practical first step would be to **focus on improving the group level** and evaluating if this addresses the majority of the identified issues.

Currently, the group level, which includes the debugger, features minimal UI elements—a colored border on the left that also functions as a tutor button.

My suggestion is to **transform each group into a card** style element (no elevation, rounded corners, and a subtle hairline border). The card would have a header with a lightly colored background and a content area containing the current content (excluding the colored left border). These cards would be **collapsible**, showing only the header when collapsed.

The header would contain the number of the test group (linkable), the status (correct, wrong), a button to start debugging and a button to expand/collapse the group.

By default, I would **collapse all correct groups**, but also add a toggle to expand all (similar to the current diff option).

Atop the current tab, a **visual representation of all the groups** could be introduced—perhaps using colored circles—to improve the visibility of failed tests. Clicking a circle would bring the corresponding group into focus, potentially with an additional border color change to accentuate it further. This could be achieved using the `:target` CSS selector. To maintain visibility while scrolling, the tabs and group visualization would utilize `position: sticky`.

pdawyndt:
If valuable for the rethinking the design: Individual testcases and groups now only have a Boolean status, but for students it might be more informative to make more distinction why a testcase/group failed (runtime error, time limit exceeded, memory limit exceeded). For that purpose, we're also missing a status for testcases/groups that have not been executed due to an earlier fatal issue (e.g. time limit exceeded for an earlier group).

https://github.com/dodona-edu/universal-judge/issues/342

pdawyndt:
When drafting the TESTed 2.0 paper, we also renamed the tab-context-testcase-test grouping into unit-case-test-output, to align it better with unittest lingo.

[...]

PR #5102, opened 2023-11-03
Rework the feedback table

bmesuere:
This pull request partly reworks the feedback table according to #5101. The main goal was to make the debug button more prominent. In the old design, this was not possible since the feedback level that was associated with the debugger (`group`) did not have any UI. 

### Cards with header
Every group now has its own card with a header. The header contains an incremental number of the group/test, the status in words, a debug button if relevant, and a button to collapse the card. Correct cards are collapsed by default, except if everything is correct. The toggle button that was used to hide correct "tests" was repurposed to collapse/expand correct cards.

image
image

### Tab summary

In between the tabs and the feedback table, we showed 0, 1 or 2 toggle buttons in the old design. The empty space at the left is now used to add a summary of that tab. Every card is represented by a colored circle, indicating its status. These circles are clickable and link to the corresponding card. The card then scrolls into view, the border size is increased to 2px and expands if needed.

The height of this summary is fixed (to the height of the toggle buttons) and will show a vertical scroll bar if there would be too many tests.

image
image

### Other changes
- The tabs + summary are now "sticky", which means they'll stay visible when scrolling down
- The lower test levels were left unchanged for now. I just tweaked the spacing a bit and moved a status icon from the right to the left.

## Implementation
Most of the changes are done in the feedback table renderer. Some of the html that is generated there could in theory move to javascript to reduce the time that is needed to generate a server response. This could probably be done in a later PR as we'll probably merge the pythia renderer at a later time.

Collapsing the cards is almost entirely done in css by adding/removing the `collapsed` class.

---

If you test this, please try exercises with complex feedback tables. Since our format is very flexible, there are probably edge cases which I have missed.


pdawyndt:
You may see some place to integrate these features at the card level:

- toggles (now students have to scroll back to the top of the page to perform that action)
- copy testcase description to clipboard (= stdin, statements/expressions)

pdawyndt:
Lena and me prefer option 3. We find extra border for 2 too hard. Green/red text on blue background doesn’t blend so well.

bmesuere:
For showing the group status, I made a few alternatives:
1) Show an icon + status in bold and in the status color. Bgcolor of the card is light blue for all, target card gets dark blue border:
image

2) Same as 1, but also set the border to the status color. Target card gets a 2px border.
image

3) Change the header color to the status
image

---

I don't like 3 (in part because I never liked the light green for areas). 1 is maybe to subtle, 2 seems like a good middle ground.
On a busy page, this might become too busy. An alternative might be to only add the border for wrong groups.

I'll implement 2 for now.

bmesuere:
> Lena and me prefer option 3. We find extra border for 2 too hard. Green/red text on blue background doesn’t blend so well.

My screenshot was a bad example. On "real" pages, the prominent background interferes too much with the diff background (like [here](https://github.com/dodona-edu/dodona/assets/481872/a00c7e80-51f8-49e9-87d6-369e2c07c2eb)). They are both "large areas with the same red/green" color, but have a different function (failing "similar things should look the same, similar looking things should function the same). The group status is not that important that we want such explicit focus (failing the squint test). Correct groups will be collapsed by default which will make the different enough.

pdawyndt:
Not sure about the “collapse correct by default” yet, but may have to see it in practice. My fear is that will not be very supportive for students if the main focus is on what’s wrong, by hiding what is already OK.

jorg-vr:
I approve of the new layout, this is a nice step forward!

I do have some notes on the colors:
- I prefer the light grey outline border. It makes the cards a lot less "heavy" on the page. Otherwise they al seem focused
- The blue color does not go well with the red and green. This is a problem in al three designs, as there is either red/green on blue or the reverse
- I greatly prefer the  third design, that most clearly shows the test state without an ugly hard border. I would even allow the buttons in this design to be green/red. Maybe design 1 could also work if the header background wasn't blue.

app/helpers/renderers/feedback_table_renderer.rb
```diff
@@ -141,98 +141,141 @@ def tab_id(t, i)
...
+    @builder.div(class: 'feedback-table-options sticky') do
+      # summary of tests
+      @builder.div(class: 'tab-summary') do
+        @builder.span(class: 'tab-summary-text') do
+          @builder.text! "#{groups_correct}/#{groups_total} #{I18n.t('submissions.show.correct_group').downcase}:"
+        end
+        @builder.div(class: 'tab-summary-icons') do
+          t[:groups]&.each_with_index do |g, i|
+            @builder.div(class: g[:accepted] ? 'correct' : 'wrong') do
+              @builder.a(href: "#tab-#{tab_i + 1}-group-#{i + 1}", title: "##{i + 1}") do
```
jorg-vr, line 161: Note: Tabnames can be custom (at least for pythia). This could theoretically result in issues here if the custom name has this specific format. But I don't think we should be worried about it.

jorg-vr (changes requested):
Let me start by saying that I approve of the redesign. I think it will greatly improve discoverability of our features and the navigation to each group overcomes the most pressing issues of the extra vertical space.

The left aligned status icon should receive its own column. Right now it looks odd with multiple lines of input (eg submission https://naos.dodona.be/en/submissions/14545820/#tab-1-group-14)
image

I still prefer the non colored borders. It makes the page a lot less intense, but all relevant information is still present. Definitely with the correct default closed, there is already enough difference between both cases.
image

In a compromise I could live with only the wrong borders being colored. This results in my attention being directed to the wrong groups
image
But on an exercise with a lot of mistakes this is still to heavy. In my opinion the colored border tells me where I should focus. If all cards have a border color I need to focus everywhere at once, which makes the page heavy to process. (Only using the border color when focused, eg when redirected there by the navigation circles, would be my expected usage)

For the top bar background color I would suggest  using 'elevation' from [the material 3 style guide](https://m3.material.io/styles/color/roles#22948d54-0450-4cab-8f4f-8853a8c6eccc). Eg elevation + 2 can be achieved by ` background-color: rgba(var(--d-primary-rgb), 8%);`

image
image

Note: elevation using transparency is already deprecated, we should define new color definitions for the new style guide: https://material.io/blog/tone-based-surface-color-m3

app/helpers/renderers/feedback_table_renderer.rb
```diff
@@ -141,98 +141,141 @@ def tab_id(t, i)
...
                      title: 'Start debugger',
                      'data-statements': (g[:data][:statements]).to_s,
                      'data-stdin': (g[:data][:stdin]).to_s) do
-          @builder.div(class: 'tutor-strip-icon') do
-            @builder.i('', class: 'mdi mdi-launch mdi-18')
+            # this is the bug-play-outline icon from https://pictogrammers.com/library/mdi/icon/bug-play-outline/
+            @builder.i(class: 'mdi me-1 custom-material-icons') do
+              @builder << '<svg xmlns="http://www.w3.org/2000/svg" height="24px" viewBox="0 0 24 24" width="24px" fill="currentColor"><path d="M19 7H16.19C15.74 6.2 15.12 5.5 14.37 5L16 3.41L14.59 2L12.42 4.17C11.96 4.06 11.5 4 11 4S10.05 4.06 9.59 4.17L7.41 2L6 3.41L7.62 5C6.87 5.5 6.26 6.21 5.81 7H3V9H5.09C5.03 9.33 5 9.66 5 10V11H3V13H5V14C5 14.34 5.03 14.67 5.09 15H3V17H5.81C7.26 19.5 10.28 20.61 13 19.65V19C13 18.43 13.09 17.86 13.25 17.31C12.59 17.76 11.8 18 11 18C8.79 18 7 16.21 7 14V10C7 7.79 8.79 6 11 6S15 7.79 15 10V14C15 14.19 15 14.39 14.95 14.58C15.54 14.04 16.24 13.62 17 13.35V13H19V11H17V10C17 9.66 16.97 9.33 16.91 9H19V7M13 9V11H9V9H13M13 13V15H9V13H13M17 16V22L22 19L17 16Z" /></svg>'
```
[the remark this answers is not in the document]
bmesuere: No, the helper doesn't work here. Well, the helper does, but the render function doesn't. The explanation is hidden in the commit message: https://github.com/dodona-edu/dodona/pull/5102/commits/a926135ad5eab9d9dc4c542cdc40b7c91ef3203c

app/helpers/renderers/feedback_table_renderer.rb
jorg-vr: if a tab was named "${othertab_name}-group-1", the links could be become non unique. But as I said, this was just a note, not something I expect to happen in practice

app/helpers/renderers/feedback_table_renderer.rb
```diff
@@ -141,98 +141,141 @@ def tab_id(t, i)
...
+        # Add a link to the debugger if there is data
+        if g[:data] && (g[:data][:statements] || g[:data][:stdin])
+          @builder.a(class: 'btn btn-text tutorlink',
                      title: 'Start debugger',
                      'data-statements': (g[:data][:statements]).to_s,
                      'data-stdin': (g[:data][:stdin]).to_s) do
-          @builder.div(class: 'tutor-strip-icon') do
-            @builder.i('', class: 'mdi mdi-launch mdi-18')
+            # this is the bug-play-outline icon from https://pictogrammers.com/library/mdi/icon/bug-play-outline/
+            @builder.i(class: 'mdi me-1 custom-material-icons') do
```
chvp, line 232: Maybe I'm missing something, but why doesn't `mdi mdi-24 mdi-bug-play-outline` not work?

chvp: Ah, because we're using version 5.x of the font. Maybe we should upgrade to 6.x?

chvp (changes requested):
I also agree that the borders are too visually heavy. Correct and incorrect contexts are already visually different enough IMO, so that the border is not needed.

I have less of an opinion on the background color of the card headers (either way).

app/helpers/renderers/feedback_table_renderer.rb
bmesuere: I didn't do this because this meant adding 2k additional icons of which we only would use 1 and the icon font is already huge right now. (It would also mean going over the changelog of 20 version to check if any of our icons wasn't renamed)

niknetniko (commented):
- I would let the box on top expand with correct/wrong statusses as needed, to prevent scrollbars. On platforms were scrollbars are not hidden by default, this is very ugly, as noted already.
- Perhaps another way to save space is to move the text to the back:
image
- I agree with Jorg to use less intense colours as the card headings.
- I agree to only use a coloured border if it is wrong. An alternative would perhaps be to use coloured borders but without background colour in the card title (image the code is properly outlined on the left):
image

- Not completely hiding the correct tests and only collapsing them makes it annoying to find a failed test at the bottom or halfway through the list by scrolling. This is remedied be the links on top, so just something I noticed. Perhaps we also need a new icon, a "collapse all" instead of the eyes.
- I agree with removing the padding from the insides

One other remark is that for exercises where each context has only one testcase and test, the rework loses a lot of information density, as each group takes up almost twice as much space (of course, removing the inside padding will help with this, but still):
image

pdawyndt:
Do "contexts" have an optional description? If yes, where would these descriptions fit? In the title bar? If no, don't bother.

jorg-vr:
Inspired by Niko's no color, no padding varient, which I also like, I tried a more condensed view without cards but only separating lines:
image

But it depends a lot on the grey areas for visual hierarchy, and I am not completely sure how much control we always have in that area. It should probably be tested with some more complex exercises, I just did a small mockup in browser.

I think the cards still look a bit nicer, but this saves a lot of vertical space which might be an asset. And with some minor tweaks this kind of layout could also be perfected

<details> 
  <summary>The HTML of the mockup for reference </summary>
 ```
<div class="groups" style="border: 1px solid var(--d-divider);/*! border-bottom: ; */border-radius: var(--d-border-radius-base);"><div id="tab-1-group-1" class="group correct collapsed" style="border-bottom: 1px solid var(--d-divider);"><div style="display: flex;flex-direction: row;align-items: center;font-size: 16px;font-weight: 400;/*! justify-content: unset; */padding-left: 12px;" class=""><a href="#tab-1-group-1">#1</a><span class="ms-2 me-2">·</span><span class="group-status" style="color: var(--d-green);"><i class="mdi mdi-check mdi-18"></i><span class="ms-1">Correct</span></span><div class="flex-spacer"></div><a class="btn btn-text tutorlink" title="Start debugger" data-statements="" data-stdin="5
1
7
2
0
5
3
6
9
"><i class="mdi me-1 custom-material-icons"><svg xmlns="http://www.w3.org/2000/svg" height="24px" viewBox="0 0 24 24" width="24px" fill="currentColor"><path d="M19 7H16.19C15.74 6.2 15.12 5.5 14.37 5L16 3.41L14.59 2L12.42 4.17C11.96 4.06 11.5 4 11 4S10.05 4.06 9.59 4.17L7.41 2L6 3.41L7.62 5C6.87 5.5 6.26 6.21 5.81 7H3V9H5.09C5.03 9.33 5 9.66 5 10V11H3V13H5V14C5 14.34 5.03 14.67 5.09 15H3V17H5.81C7.26 19.5 10.28 20.61 13 19.65V19C13 18.43 13.09 17.86 13.25 17.31C12.59 17.76 11.8 18 11 18C8.79 18 7 16.21 7 14V10C7 7.79 8.79 6 11 6S15 7.79 15 10V14C15 14.19 15 14.39 14.95 14.58C15.54 14.04 16.24 13.62 17 13.35V13H19V11H17V10C17 9.66 16.97 9.33 16.91 9H19V7M13 9V11H9V9H13M13 13V15H9V13H13M17 16V22L22 19L17 16Z"></path></svg></i>Debuggen</a><a class="btn btn-icon btn-collapse"><i class="mdi mdi-chevron-down"></i></a></div><div class=""><div class="testcase correct"><div class="description"><div class="indicator"><i class="mdi mdi-check mdi-18"></i></div><span class="code">5
1
7
2
0
5
3
6
9
</span></div><div class="test"><div class="description"><span class="badge bg-success">stdout</span></div><div class="test-accepted"><span class="output">7
</span></div></div></div></div></div><div id="tab-1-group-1" class="group correct collapsed" style="border-bottom: 1px solid var(--d-divider);"><div style="display: flex;flex-direction: row;align-items: center;font-size: 16px;font-weight: 400;/*! justify-content: unset; */padding-left: 12px;" class=""><a href="#tab-1-group-1">#2</a><span class="ms-2 me-2">·</span><span class="group-status" style="color: var(--d-green);"><i class="mdi mdi-check mdi-18"></i><span class="ms-1">Correct</span></span><div class="flex-spacer"></div><a class="btn btn-text tutorlink" title="Start debugger" data-statements="" data-stdin="5
1
7
2
0
5
3
6
9
"><i class="mdi me-1 custom-material-icons"><svg xmlns="http://www.w3.org/2000/svg" height="24px" viewBox="0 0 24 24" width="24px" fill="currentColor"><path d="M19 7H16.19C15.74 6.2 15.12 5.5 14.37 5L16 3.41L14.59 2L12.42 4.17C11.96 4.06 11.5 4 11 4S10.05 4.06 9.59 4.17L7.41 2L6 3.41L7.62 5C6.87 5.5 6.26 6.21 5.81 7H3V9H5.09C5.03 9.33 5 9.66 5 10V11H3V13H5V14C5 14.34 5.03 14.67 5.09 15H3V17H5.81C7.26 19.5 10.28 20.61 13 19.65V19C13 18.43 13.09 17.86 13.25 17.31C12.59 17.76 11.8 18 11 18C8.79 18 7 16.21 7 14V10C7 7.79 8.79 6 11 6S15 7.79 15 10V14C15 14.19 15 14.39 14.95 14.58C15.54 14.04 16.24 13.62 17 13.35V13H19V11H17V10C17 9.66 16.97 9.33 16.91 9H19V7M13 9V11H9V9H13M13 13V15H9V13H13M17 16V22L22 19L17 16Z"></path></svg></i>Debuggen</a><a class="btn btn-icon btn-collapse"><i class="mdi mdi-chevron-left"></i></a></div><div class="" style="display: none;"><div class="testcase correct"><div class="description"><div class="indicator"><i class="mdi mdi-check mdi-18"></i></div><span class="code">5
1
7
2
0
5
3
6
9
</span></div><div class="test"><div class="description"><span class="badge bg-success">stdout</span></div><div class="test-accepted"><span class="output">7
</span></div></div></div></div></div><div id="tab-1-group-1" class="group correct collapsed" style="border-bottom: 1px solid var(--d-divider);"><div style="display: flex;flex-direction: row;align-items: center;font-size: 16px;font-weight: 400;/*! justify-content: unset; */padding-left: 12px;" class=""><a href="#tab-1-group-1">#3</a><span class="ms-2 me-2">·</span><span class="group-status" style="color: var(--d-green);"><i class="mdi mdi-check mdi-18"></i><span class="ms-1">Correct</span></span><div class="flex-spacer"></div><a class="btn btn-text tutorlink" title="Start debugger" data-statements="" data-stdin="5
1
7
2
0
5
3
6
9
"><i class="mdi me-1 custom-material-icons"><svg xmlns="http://www.w3.org/2000/svg" height="24px" viewBox="0 0 24 24" width="24px" fill="currentColor"><path d="M19 7H16.19C15.74 6.2 15.12 5.5 14.37 5L16 3.41L14.59 2L12.42 4.17C11.96 4.06 11.5 4 11 4S10.05 4.06 9.59 4.17L7.41 2L6 3.41L7.62 5C6.87 5.5 6.26 6.21 5.81 7H3V9H5.09C5.03 9.33 5 9.66 5 10V11H3V13H5V14C5 14.34 5.03 14.67 5.09 15H3V17H5.81C7.26 19.5 10.28 20.61 13 19.65V19C13 18.43 13.09 17.86 13.25 17.31C12.59 17.76 11.8 18 11 18C8.79 18 7 16.21 7 14V10C7 7.79 8.79 6 11 6S15 7.79 15 10V14C15 14.19 15 14.39 14.95 14.58C15.54 14.04 16.24 13.62 17 13.35V13H19V11H17V10C17 9.66 16.97 9.33 16.91 9H19V7M13 9V11H9V9H13M13 13V15H9V13H13M17 16V22L22 19L17 16Z"></path></svg></i>Debuggen</a><a class="btn btn-icon btn-collapse"><i class="mdi mdi-chevron-down"></i></a></div><div class=""><div class="testcase correct"><div class="description"><div class="indicator"><i class="mdi mdi-check mdi-18"></i></div><span class="code">5
1
7
2
0
5
3
6
9
</span></div><div class="test"><div class="description"><span class="badge bg-success">stdout</span></div><div class="test-accepted"><span class="output">7
</span></div></div></div></div></div><div id="tab-1-group-1" class="group correct collapsed" style="border-bottom: 1px solid var(--d-divider);"><div style="display: flex;flex-direction: row;align-items: center;font-size: 16px;font-weight: 400;/*! justify-content: unset; */padding-left: 12px;" class=""><a href="#tab-1-group-1">#4</a><span class="ms-2 me-2">·</span><span class="group-status" style="color: var(--d-red);"><i class="mdi mdi-close mdi-18"></i><span class="ms-1">Wrong</span></span><div class="flex-spacer"></div><a class="btn btn-text tutorlink" title="Start debugger" data-statements="" data-stdin="5
1
7
2
0
5
3
6
9
"><i class="mdi me-1 custom-material-icons"><svg xmlns="http://www.w3.org/2000/svg" height="24px" viewBox="0 0 24 24" width="24px" fill="currentColor"><path d="M19 7H16.19C15.74 6.2 15.12 5.5 14.37 5L16 3.41L14.59 2L12.42 4.17C11.96 4.06 11.5 4 11 4S10.05 4.06 9.59 4.17L7.41 2L6 3.41L7.62 5C6.87 5.5 6.26 6.21 5.81 7H3V9H5.09C5.03 9.33 5 9.66 5 10V11H3V13H5V14C5 14.34 5.03 14.67 5.09 15H3V17H5.81C7.26 19.5 10.28 20.61 13 19.65V19C13 18.43 13.09 17.86 13.25 17.31C12.59 17.76 11.8 18 11 18C8.79 18 7 16.21 7 14V10C7 7.79 8.79 6 11 6S15 7.79 15 10V14C15 14.19 15 14.39 14.95 14.58C15.54 14.04 16.24 13.62 17 13.35V13H19V11H17V10C17 9.66 16.97 9.33 16.91 9H19V7M13 9V11H9V9H13M13 13V15H9V13H13M17 16V22L22 19L17 16Z"></path></svg></i>Debuggen</a><a class="btn btn-icon btn-collapse"><i class="mdi mdi-chevron-down"></i></a></div><div class=""><div class="testcase wrong"><div class="description"><div class="indicator"><i class="mdi mdi-close mdi-18"></i></div><span class="code">5
1
7
2
0
5
3
6
9
</span></div><div class="test"><div class="description"><span class="badge bg-danger" style="/*! margin-left: 5px; */">stdout</span></div><div class="test-failed"><div class="diffs show-split"><table class="split-diff diff"><colgroup><col class="line-nr"><col class="del-output"><col class="line-nr"><col class="ins-output"></colgroup><thead><tr><th class="line-nr" title="Jouw uitvoer"><i class="mdi mdi-18 mdi-file-account"></i></th><th>Jouw uitvoer</th><th class="line-nr" title="Verwachte uitvoer"><i class="mdi mdi-18 mdi-file-check"></i></th><th>Verwachte uitvoer</th></tr></thead><tbody><tr>
                     <td class="line-nr">1</td>
                     <td class="del"><strong>1</strong></td>
                     <td class="line-nr">1</td>
                     <td class="ins"><strong>4</strong></td>
                   </tr><tr>
                     <td class="line-nr">2</td>
                     <td class="unchanged"></td>
                     <td class="line-nr">2</td>
                     <td class="unchanged"></td>
                   </tr></tbody></table><table class="unified-diff diff"><colgroup><col class="line-nr"><col class="line-nr"><col class="output"></colgroup><thead><tr><th class="line-nr" title="Jouw uitvoer"><i class="mdi mdi-18 mdi-file-account"></i></th><th class="line-nr" title="Verwachte uitvoer"><i class="mdi mdi-18 mdi-file-check"></i></th><th></th></tr></thead><tbody><tr>
                     <td class="line-nr">1</td>
                     <td class="line-nr"></td>
                     <td class="del"><strong>1</strong></td>
                   </tr>
                   <tr>
                     <td class="line-nr"></td>
                     <td class="line-nr">1</td>
                     <td class="ins"><strong>4</strong></td>
                   </tr><tr>
                     <td class="line-nr">2</td>
                     <td class="line-nr">2</td>
                     <td class="unchanged"></td>
                   </tr></tbody></table><to_str></to_str></div></div></div></div></div></div></div>
```
</details>

bmesuere:
Thanks for the feedback. Here are the updates I've implemented:

**Scrollbars**: This is indeed very ugly on windows (I didn't spot this while testing). I now set the `scrollbar-width` to `none` for this specific element, but this only works in Firefox for now. For webkit browsers, I used a pseudo selector to hide the scrollbars in this element. In case these two approaches fail, I've also restricted overflow to the y-axis, which should remove the horizontal scrollbar and alleviate most of the concern. @niknetniko I decided on using a fixed height (and thus scrolling the little area) to be able to support scrolling the cards into view without javascript. A fixed top scroll margin is set in css for this (otherwise the card scrolls to the top of the page and is hidden by the navbar and other sticky elements).

**Card borders**: I reverted the colored borders and once again use the divider color (light gray). For targeted cards, the border is set to the primary color (dark blue). The width stays at 1px (as opposed to the 2px of the previous design).

**Multi-line alignment**: This was a bug. I didn't test with descriptions that spanned multiple lines

image

Regarding the changes I've not implemented:

**Update the icon font**: This would require making the current font even bigger. It seems that [iconify](https://pictogrammers.com/docs/guides/iconify/) might be an alternative. I'll try to add this in a separate PR. We could then gradually migrate all icons.

**Card header color**: I tried all dodona css variable colors as header and almost none "worked". Most were too prominent or didn't fit. Others (like the proposed darker gray) interfered with the other gray areas in the card content. I decided to stay with blue for now. The removal of the colored borders will probably also help make everything easier on the eyes.

**Tweak margins and padding**:  Increasing the content density by reducing white space doesn't seem necessary to me. Part of the goal of this PR was to make the hierarchy and content of the table clearer and reducing the density was an important part in this. In the previous design, you also had to scroll to see the entire table, so the difference is in the amount of scrolling. This is countered by the addition of the top summary which means that in many cases users won't have to scroll at all. However, I'm open to making adjustments to the spacing if we identify any regressions or UI glitches, such as with the multi-line descriptions. Because of the nested nature of the table and optional text and code that can show up almost everywhere, it's easy to miss a few cases.

---

I'll do a few more tests on naos to see if I missed any cases.

niknetniko:
I did a test with a bunch of messages on multiple levels, and they all seem to work:

image

> Tweak margins and padding

It bothers me a bit that the padding on the left and right are not equal (probable with the status icon being there on the left), but that could just be me.

jorg-vr:
There is a bug with the left padding in this example: https://naos.dodona.be/nl/submissions/15701527/#tab-1-group-9
image

I like the new borders, with the primary color for targeted cards.

I find it rather counter intuitive that it is impossible to close the targeted card.
User flow: jump to a card, study it, close when finished to focus on the one below => hmm this does not work? Feels like a bug.

It is also impossible to remove focus from a card, except navigating to another one. I know this it caused by the target implementation but it doens't feel natural. (Not a priority to fix, would have bothered me less if I could close the card)

The navigation help does not give any indication which one is already targeted. Definitely with al icons being the same (either wrong or correct) it can be easy to lose your position. I've noticed the tooltips after writing this => they do help with this problem, although it is still not optimal.

> Card header color: I tried all dodona css variable colors as header and almost none "worked". Most were too prominent or didn't fit. Others (like the proposed darker gray) interfered with the other gray areas in the card content. I decided to stay with blue for now. The removal of the colored borders will probably also help make everything easier on the eyes.

Did you also try `background-color: rgba(var(--d-primary-rgb), 8%);` as I suggested? 
I must say that in light mode the problem of the page being to heavy is already resolved for me in the current iteration. But in dark mode it still bothers me. But this can just be an unresolvable question of taste, so I am willing to accept as is.

jorg-vr (approved):

niknetniko (approved):

bmesuere:
I have tried setting the "currently selected card" by using :focus or :active, but both don't seem to work reliably. I think you would also expect something scroll-spy like for this. If you click an icon, the page scrolls to that card. If you then continue scrolling, it might be confusing if the clicked icon was still highlighted.

If we were to add more JavaScript in a next iteration, we could opt to:
- add a scrollspy (or highlight the clicked icon)
- manually expand the focused card instead of using css (to allow it to be collapsed again)
- dynamically compute the header height and remove the fixed height from the summary icons

chvp (approved):

pdawyndt:
I was in the midst of evaluating a test when the new feedback table was rolled out, and now I feel seriously hampered by the fact that by default all the correct contexts are collapsed (and I also have no way to expand them all). I'm loosing quite some overview, especially because the bash judge (then one whose results I'm currently evaluating) puts every individual test in its own context.

Example: https://dodona.be/nl/feedbacks/752542661/

bmesuere:
You can expand them all by clicking the top eye button:
image

We'll investigate what the best way forward is for evaluations (storing the user preference vs expanding).

PR #5124, opened 2023-11-09
Only return administrating courses in LTI picker

jorg-vr:
This pull request limits the courses returned by the LTI picker to administrating courses.

A user should thus easily find al relevant courses.

- [x] tested on uforatest with naos


bmesuere (approved):
Has this always been a bug?

niknetniko (approved):

jorg-vr:
> Has this always been a bug?
There was a 'policy scope', that would at leas limit it to accessible courses in theory. But I have not thoroughly tested this.
But there was never any code to limit it to administrating courses

PR #5129, opened 2023-11-12
Rephrase evaluation finished toast

### LINKED ISSUE #5125: Rewording of "evaluation finished"
We confused a student who had extra time (compared to other students who already left the evaluation room) by the message "evaluation finished" that is raised when feedback is ready after a solution had been submitted.

When submitting a solution, users get to see a popup "Submission saved" ("Oplossing opgeslagen") followed by a popup "Evaluation finished" (Dutch: "Evaluatie afgerond"). The latter gets another meaning in the context of our "evaluations" (= how we call a mid-term test). We may reconsider the wording in Dodona, and choose something like "Feedback ready" (which follows the wording of the Feedback tab where that feedback is found). Student assumed that the deadline of his evaluation had passed (which was not the case).

bmesuere:
This pull request changes the text on the toast that shows up when testing is done and the feedback is available from "Evaluation finished" to "Testing finished".

Closes #5125


chvp (approved):
I agree with the rephrasing, but I also wonder about the usefulness of the toast itself. Until the related issue was made I mostly forgot about its existence, and I would think there is enough feedback already about the tests being finished.

bmesuere:
I also almost removed it entirely, but there is a use case when testing takes a while. If the users scrolls up and reads the problem description or checks his code again while waiting, it might be useful.

jorg-vr (approved):

PR #5132, opened 2023-11-12
Improve webpack build time

bmesuere:
This pull request tweaks the webpack config to improve the build time. This will speed up the github actions.

There are 2 changes:
- Don't minimize the code (using terser) when building for tests
- Don't typecheck when building the code, this is checked in the linting action

Build times on my macbook:
| Version | Time in s |
|--------|--------:|
| Original version | 34.06 |
| Don't minimize while testing | 16.56 |
| Don't typecheck | 12.29 |

### DIFF
--- a/config/webpack/webpack.config.js
+++ b/config/webpack/webpack.config.js
@@ -22,7 +22,15 @@ const config = {
                     return modulePath.endsWith('.ts') && !modulePath.endsWith('test.ts');
                 },
                 exclude: /node_modules/,
-                use: ["babel-loader", "ts-loader"],
+                use: [
+                    "babel-loader",
+                    {
+                        loader: 'ts-loader',
+                        options: {
+                            transpileOnly: true,
+                        },
+                    }
+                ],
             },
         ],
     },
@@ -62,7 +70,7 @@ if (process.env.NODE_ENV === "development") {
     config.devtool = "inline-source-map";
 }
 
-// disable terser minimization when running tests
+// disable terser minimization when running
 if (process.env.RAILS_ENV === "test") {
     config.optimization.minimize = false;
 }
--- a/config/webpack/webpack.config.js
+++ b/config/webpack/webpack.config.js
@@ -62,6 +62,11 @@ if (process.env.NODE_ENV === "development") {
     config.devtool = "inline-source-map";
 }
 
+// disable terser minimization when running tests
+if (process.env.RAILS_ENV === "test") {
+    config.optimization.minimize = false;
+}
+
 // Test, Staging and Production use default config
 
 module.exports = config;


bmesuere:
We are getting 2 new warnings (which is strange since we're doing less):

```
WARNING in ./app/assets/javascripts/components/saved_annotations/saved_annotation_form.ts 1229:80-95
export 'SavedAnnotation' (imported as 'SavedAnnotation') was not found in 'state/SavedAnnotations' (possible exports: savedAnnotationState)
 @ ./app/assets/javascripts/components/saved_annotations/new_saved_annotation.ts 1597:0-33
 @ ./app/javascript/packs/saved_annotation.js

WARNING in ./app/assets/javascripts/components/saved_annotations/saved_annotation_form.ts 1229:146-1[61](https://github.com/dodona-edu/dodona/actions/runs/6841662871/job/18602269848?pr=5131#step:8:62)
export 'SavedAnnotation' (imported as 'SavedAnnotation') was not found in 'state/SavedAnnotations' (possible exports: savedAnnotationState)
 @ ./app/assets/javascripts/components/saved_annotations/new_saved_annotation.ts 1597:0-33
 @ ./app/javascript/packs/saved_annotation.js
 ```

bmesuere:
The warning seems to be normal behaviour according to the [ts-loader docs](https://github.com/TypeStrong/ts-loader#transpileonly).

> If you enable this option, webpack 4 will give you "export not found" warnings any time you re-export a type:
> 
> ```
> WARNING in ./src/bar.ts
> 1:0-34 "export 'IFoo' was not found in './foo'
>  @ ./src/bar.ts
>  @ ./src/index.ts
> ```
> 
> The reason this happens is that when typescript doesn't do a full type check, it does not have enough information to determine whether an imported name is a type or not, so when the name is then exported, typescript has no choice but to emit the export. Fortunately, the extraneous export should not be harmful, so you can just suppress these warnings:

chvp (approved):
Do we want to disable the warning(s) mentioned?

niknetniko (approved):

bmesuere:
They are just warnings, so I would not disable them since the recommended way to do is is a regex which might also match real errors?

jorg-vr (commented):
does this mean that we can move the types back to devdependencies?

bmesuere:
Possibly indeed, but this needs to be tested.

jorg-vr:
Naos also runs production deploy, sit it can be tested there (or by only installing production deps locally)
Will you tests this in this pr or should I do it in a future pr?

bmesuere:
I would do it in a future PR. There might also be some other optimizations possible. I noticed that we also include terser in the rubygems file. This might mean that terser is also run as part of sprockets.

jorg-vr (approved):

PR #5136, opened 2023-11-13
Stop Sprockets from compressing JS

bmesuere:
This pull request removes terser as a rails dependency and disables js compression using Sprockets.

Since our JS modernization efforts, all JS is processed by webpack first which already runs Terser.

This should speed up deploys.


bmesuere:
This seems to work on naos.

Gemfile
```diff
@@ -21,7 +21,7 @@ gem 'jsbundling-rails', '~> 1.2.1'
 gem 'sprockets-rails', '~> 3.4.2'
 
 # Use Terser as compressor for JavaScript assets
-gem 'terser', '>= 1.1.1'
+# gem 'terser', '>= 1.1.1'
```
jorg-vr, line 24: I would remove instead of comment this

jorg-vr (changes requested):

Gemfile
bmesuere: I left it because it is part of the default rails gemfile (at least, git blame told me you last modified it while updating rails).

chvp (approved):

jorg-vr (approved):

PR #5157, opened 2023-11-15
Recalculate selected code upon copy in submission

### LINKED ISSUE #5155: Allow copy-pasting while adding annotations to evaluation
When adding annotations to submissions in an evaluation (but I guess this is also the case when adding annotations in general), copy-pasting does not show its normal behaviour. The unwanted behaviour does not always occur, and I haven't figured out what's really causing this issue, but it happens way too often to have a smooth UX. The unwanted behaviour is that copy-paste does not seem to work: something is already copied into the paste buffer and the buffer cannot be overwritten by copying something else. Here's a scenario where the seems to occur all the time:

1. select part the submitted code
2. copy the selection
3. open the form to add an annotation
4. copying something from the form does not work (copied section from the submited code is still in the paste buffer)
4. copying something else from the submitted code does not work either (original copied section from the submited code before the form was opened is still in the paste buffer)

image

jorg-vr:
This pull request recalculates the selection when a copy event is triggered. This makes the code copy independent of the selectedRange tracked for the annotation form.

I had to improve the offset calculation a bit, because it did not work correctly if the focusnode itself was a PRE element (instead of a textnode child of a pre element). Now it should be more general and support both cases.

I also disabled range modification when triggered by the copy call. In the other use case the selection is being modified directly after creation which feels intuitive. But when it happens on copy it feels strange, and there is no good reason to do it. So I turned it off in that case.

Closes #5155

### DIFF
--- a/app/assets/javascripts/code_listing.ts
+++ b/app/assets/javascripts/code_listing.ts
@@ -55,6 +55,10 @@ function initAnnotateButtons(): void {
             return; // if there is no code selection, let the browser handle the copy event
         }
 
+        if (userAnnotationState.formShown) {
+            return; // if the annotation form is shown, let the browser handle the copy event
+        }
+
         const selectedCode = submissionState.code.split("\n").slice(selection.row - 1, selection.row + selection.rows - 1);
         // on the first and last line, selection might only cover part of the line
         // only copy the selected columns/characters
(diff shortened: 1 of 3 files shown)


niknetniko (approved):
In my limited testing, copying while the annotation form was open did not result in a log of cases where there were extra newlines or duplicate code, so I'm not sure how big a problem this is, and if it warrants the bigger refactor or not.

bmesuere (approved):

niknetniko (approved):

PR #5158, opened 2023-11-15
Make ctrl-f more usable on submissions

jorg-vr:
This pull request improves browser search (ctrl-f) features on submission pages.

The issue was that search matched each code element three times as it is repeated in each layer.
This issue was reported by Steven, who made the case for the search feature when trying to understand a students code while grading.

A side effect: copying code will only copy spaces while our custom hack isn't working. This will happen more often with https://github.com/dodona-edu/dodona/pull/5157 so accepting this pr will require changes there.

I do not see any other options for fixing this aside of reverting the whole three layered approach for annotations.
But I am open for suggestions.

### DIFF
--- a/app/assets/javascripts/components/annotations/line_of_code.ts
+++ b/app/assets/javascripts/components/annotations/line_of_code.ts
@@ -156,12 +156,14 @@ export class LineOfCode extends ShadowlessLitElement {
 
         for (const range of this.ranges) {
             const substring = this.code.substring(range.start, range.start + range.length);
+            // replace every non-whitespace character with a non-breaking space
+            const nonSearchableSubstring = substring.replace(/\S/g, "\u00a0");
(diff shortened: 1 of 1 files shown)


bmesuere (approved):
Hmm, I tought I suggested this when implementing 3-layer PR, but that there was some reason not to do it.

chvp (approved):

jorg-vr:
I think this was part of the reason
> A side effect: copying code will only copy spaces while our custom hack isn't working. This will happen more often with https://github.com/dodona-edu/dodona/pull/5157 so accepting this pr will require changes there.

I am also not 100% wordbreaks will always be made the same way by browsers. I couldn't find a good definition for what soft-wrap points are considered by the browser. (But couldn't reproduce a breaking case in practice)

PR #5162, opened 2023-11-16
Make team explicit and add publications page

### LINKED ISSUE #4092: Add publication page
Add a page that overviews all publications related to Dodona. We might use the [Unipept](https://unipept.ugent.be/publications) publications page as an inspiration for layout.

We might also cross-reference a similar page on the Dolos homepage.

Relates to #2937

chvp:
Fixes #4092.

Publications are listed in reverse chronological order. Titles and abstracts are kept in English for both the Dutch and English versions of the pages.


niknetniko (changes requested):
Should there be a link somewhere to the actual papers?

niknetniko (approved):

jorg-vr (approved):

bmesuere (changes requested):
Add more links to the publication page:
- signed out nav bar
- signed out home page
- about page

Instead of floating the image, I would use a col class. This will also add some padding between text and image + provide a better small screen experience.

bmesuere (approved):

PR #5167, opened 2023-11-17
Don't show saved annotations from other users to zeus

### LINKED ISSUE #5166: Saved annotations might not always be correctly filtered
Investigate @chvp her issue.

jorg-vr:
This pull request removes saved annotations from the results zeus users receive.
This gives zeus users the intended user experience. In case of detecting bugs in this feature they should use impersonate.

- [x] Tests were added

Closes #5166


chvp (approved):
Thanks!

niknetniko (approved):

PR #5175, opened 2023-11-20
Partially remove saml monkey patch

### LINKED ISSUE #3396: Remove SAML metadata monkeypatches
Belnet requires that specific fields are present in the SAML metadata. The gem we used didn't support this so we monkeypatched the code that generated the metadata to add these fields:
```xml
<md:Organization>
<md:OrganizationName xml:lang="en">UGent - Dodona</md:OrganizationName>
<md:OrganizationName xml:lang="nl">UGent - Dodona</md:OrganizationName>
<md:OrganizationName xml:lang="fr">UGent - Dodona</md:OrganizationName>
<md:OrganizationDisplayName xml:lang="en">UGent - Dodona</md:OrganizationDisplayName>
<md:OrganizationDisplayName xml:lang="nl">UGent - Dodona</md:OrganizationDisplayName>
<md:OrganizationDisplayName xml:lang="fr">UGent - Dodona</md:OrganizationDisplayName>
<md:OrganizationURL xml:lang="en">https://dodona.ugent.be</md:OrganizationURL>
<md:OrganizationURL xml:lang="nl">https://dodona.ugent.be</md:OrganizationURL>
<md:OrganizationURL xml:lang="fr">https://dodona.ugent.be</md:OrganizationURL>
</md:Organization>
<md:ContactPerson contactType="technical">
<md:GivenName>Dodona</md:GivenName>
<md:SurName>Helpdesk</md:SurName>
<md:EmailAddress>dodona@ugent.be</md:EmailAddress>
</md:ContactPerson>
```

In a recent version, ruby saml added support for custom metadata fields. This means that the monkeypatch can probably be removed.

More specifically, I think `lib/SAML/metadata.rb` can be removed (after inspecting the edit history to check if other changes weren't made). The fields that are added in line 128 till 166 should be moved to separate class as is shown in https://github.com/onelogin/ruby-saml/pull/602 

The change should be validated by comparing the generated metadata at `/users/saml/metadata` before and after the fix, and by signing in into ugent from naos.

jorg-vr:
I tested the xml difference on naos and only found a single difference: `use="signing"` was specified in the new version while it wasn't in the old version.

I added a new monkey patch to be able to stick to the old behavior.
With each new monkey patch the value of this pr reduces though...

I also added a new test that just does a string compare on the xml with an expected result.
This isn't the best way to compare xmls, but if KULeuven is this strict, it is best not to allow any changes at all. Even changes in unrelated fields or white space might break the connection for all we know.

Reverts dodona-edu/dodona#5171
closes #3396


chvp (approved):
Can we test this on naos before deploying in production?

chvp:
We have pretty good contacts with Vera Van Noort, I think we can ask her.

bmesuere (changes requested):
I discussed this with @jorg-vr and agree that this "fix" might even be worse than our original monkey patched version.

The proposal is thus to keep our original monkey patch + keep the test that was added here to catch any future changes to the metadata output.

PR #5186, opened 2023-11-20
Update rails cache version

### LINKED ISSUE #5033: Update `config.active_support.cache_format_version`
This variable should be set to `7.1` if we do not revert #5031

jorg-vr:
Closes #5033

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -11,16 +11,6 @@ module Dodona
     # Initialize configuration defaults for originally generated Rails version.
     config.load_defaults 7.1
 
-    # Change the format of the cache entry.
-    # Changing this default means that all new cache entries added to the cache
-    # will have a different format that is not supported by Rails 7.0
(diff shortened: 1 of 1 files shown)


config/application.rb
```diff
@@ -20,7 +19,7 @@ class Application < Rails::Application
     # and you have no plans to rollback.
     # When you're ready to change format, add this to `config/application.rb` (NOT
     # this file):
-    config.active_support.cache_format_version = 7.0
+    config.active_support.cache_format_version = 7.1
```
chvp, line 22: Since `load_defaults 7.1` is specified above, removing this line should result in the same behaviour.

bmesuere: I would then also remove (most of) the comments above

bmesuere (approved):

chvp (approved):

PR #5190, opened 2023-11-22
Add edit submission link when evaluating a submission

### LINKED ISSUE #2054: Add button to edit code and resubmit from evaluation
When evaluating submitted solutions, there's no option to directly edit and resubmit a solution, as is the case from the submission page itself.

jorg-vr:
This pull request adds an edit submission link to to the feedback show page.

image

On smaller screens this results in:
image

The new button now also appears on non owned submissions, replacing the edit button on those pages
image

I also generalized some css about icon buttons in card titles. This also improved the submission index page.
old:
image

new:
image

Closes #2054


config/locales/js/nl.yml
```diff
@@ -324,3 +324,6 @@ nl:
       dark: Donker
       system: Systeem
       theme: Stijl
+    feedbacks:
+      submission:
+        submit: Dien deze oplossing in
```
niknetniko, line 329:
```suggestion
        submit: Deze oplossing indienen
```
This is more consistent with the button next to it

app/assets/javascripts/i18n/translations.json
```diff
@@ -751,6 +756,11 @@
       "favorite-course-do": "Voeg toe aan favorieten",
       "favorite-course-failed": "Cursus aan favorieten toevoegen mislukt",
       "favorite-course-succeeded": "Cursus toegevoegd aan favorieten",
+      "feedbacks": {
+        "submission": {
+          "submit": "Dien deze oplossing in"
```
niknetniko, line 761:
```suggestion
          "submit": "Deze oplossing indienen"
```

niknetniko (approved):

app/assets/javascripts/components/annotations/annotation_options.ts
```diff
@@ -23,6 +26,14 @@ export class AnnotationOptions extends i18nMixin(ShadowlessLitElement) {
...
     }
 
+    get canResubmitSubmission(): boolean {
+        return userState.hasPermission("submission.submit_as_own");
+    }
+
+    get resubmitPath(): string {
+        return `/courses/${courseState.id}/exercises/${exerciseState.id}/?edit_submission=${submissionState.id}`;
+    }
+
```
bmesuere, line 36: I understand that the button visually fits in with the rest of these options, but functionality wise, it has nothing to with the rest of this class. Can't these 2 getters move to the submission?

bmesuere (changes requested):
The functionality looks good, but the buttons can use some margin.
In the normal view, the buttons are too close to each other horizontally (if you hover over the new button). On small screens, the same is true vertically.

I would also move the breakpoint down 1 tier (to large). This then overflows in Dutch for an evaluation, but I would then hide the word "annotations" (the label next to the toggle buttons) instead.

app/assets/javascripts/code_listing.ts
```diff
@@ -17,14 +17,18 @@ import {
 
 const MARKING_CLASS = "marked";
 
-function initAnnotations(submissionId: number, courseId: number, exerciseId: number, userId: number, code: string, questionMode = false): void {
+function initAnnotations(submissionId: number, courseId: number, exerciseId: number, userId: number, code: string, userIsStudent: boolean, canSubmitAsOwn: boolean): void {
```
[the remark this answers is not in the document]
jorg-vr: This function is only used in a single place, so making it an object would be very artificial, definitely with the rather specific options.

Although a case could be made to group the first options into a `submission` object. This would be more reusable in the frontend. But I see no reason to refactor that right now.

bmesuere (approved):

PR #5204, opened 2023-11-28
Update papyros to version 2.0.0

jorg-vr:
This pull request is related to the papyros build update https://github.com/dodona-edu/papyros/pull/558

Papyros moved from a prebundled file to separate files. This allows us to avoid loading a lot of dependencies twice (eg codemirror).

Required changes:
- [x] load papyros css from css instead of from js
- [x] Avoid requiring commons chunk in serviceworker
- [x] Allow service worker to run on root with header `headers['Service-Worker-Allowed'] = '/'`

Future work: Avoid local duplicate of serviceworker
This would greatly simplify the setup, removing extra parameters from papyros and removing extra local configuration.
I haven't done this yet as I haven't yet found how to set the header `headers['Service-Worker-Allowed'] = '/'` to a specific asset.
Also, the number of changes in this papyros version was already quite large. So keeping this change for the future seemed sensible.

### Bundle
Before 13.74MB
image

After 8.06MB
image
The tar with python code of 1.35MB should be added to this total
Bringing it to 9.41MB


jorg-vr:
finally figured out the issue in the current setup.
First of all everything works correctly in chrome.

In firefox the read requests do not get intercepted by the service worker. (while write request are still intercepted as expected)

The probable cause:
- with the new build changes, the workers are separate packages instead of inline files.
- these are served from /assets/... (as is every other asset
- For some reason firefox decides they should trigger the serviceworker from the assets scope

I confirmed this issue by using a service worker on the /assets scope, which does intercept the read requests
This is not a solution as write requests are still scoped under the activity page

Prefered solution would probably be to run the serviceworker on the global scope, catching all.
This might be feasible by simple headers, but I haven't been able to force rails to provide those headers on a specific asset

This might also have side effects when running multiple activities simultaneously, but I'll have to prove this theory in practice

jorg-vr:
I'll publish the non beta version of 2.0.0 once this pr is approved

bmesuere (changes requested):
I tested this in Safari, but received an error when entering a value for `input`:
```
RuntimeError: The service worker for reading input isn't working. Try closing all this site's tabs, then reopening. If that doesn't work, try using a different browser.
```

There also is a slight UI problem with the header: 
image

bmesuere:
**Update**: it did work after a hard refresh of the page

jorg-vr:
The header layout is a separate bug already present in production, will fix in separate pr

I am unsure if I can avoid the need for a hard refresh...

bmesuere (approved):

pdawyndt:
I also needed a hard refresh on my Windows 10 machine with Google Chrome. Initially running a program got stuck after the first line of input was entered (interactive mode).

PR #5205, opened 2023-11-28
Fix line of code not shown on invalid judge output

jorg-vr:
This pull request fixes a line of code not being shown when the judge returns out of bounds column indexes.

Eg of submission with the issue: https://dodona.be/nl/submissions/16080236/#code
Here the second line is hidden.
image

The issue is not reproducible as the judge has already been fixed, but from the error message I discovered the issue.

Although this was caused by the judge returning to many columns, we should deal with such case more gracefully. With the new fix, a column number out of bounds will be treated as marking the whole line.


niknetniko (approved):

pdawyndt:
This solution is literally: *the whole shebang*

bmesuere (approved):

PR #5214, opened 2023-12-04
Always use svg series icon

### LINKED ISSUE #5213: Series icon layout is missing in some cases
* A course when you are not signed in, e.g. https://dodona.be/nl/courses/293/
image
* A course when you are signed in and a series without activities e.g. https://dodona.be/en/courses/65
image

In the last screenshot you can see the difference between the icon with and without layout.

Chrome 119.0.6045.200 on Windows.

jorg-vr:
This pull request fixes #5213

It were actually to separate bugs:
1. the fallback for non signed in users was the simplified icon
2. empty series were not postloaded

The fix removes the simplified icon, always using the svg version.
The amount of html remains limited for non loaded series because the svg code is moved to a web component.

Closes #5213


chvp (commented):
The tooltip location seems off: screenshot_2023-12-05-132544

chvp (approved):

app/assets/javascripts/components/series_icon.ts
```diff
@@ -0,0 +1,76 @@
...
+                <defs>
+                    <linearGradient id="rainbow" gradientTransform="rotate(135, 0.5, 0.5)" >
+                        <stop stop-color="var(--red)" offset="1%" />
+                        <stop stop-color="var(--orange)" offset="25%" />
+                        <stop stop-color="var(--yellow)" offset="40%" />
+                        <stop stop-color="var(--green)" offset="60%" />
+                        <stop stop-color="var(--blue)" offset="75%" />
+                        <stop stop-color="var(--purple)" offset="99%" />
+                    </linearGradient>
+                </defs>
```
bmesuere, line 58: only add this if needed?

app/assets/javascripts/components/series_icon.ts
```diff
@@ -0,0 +1,76 @@
...
+
+/**
+ * @element d-series-icon
+ *
+ * this icon visualizes the status of a series for a give user
+ *
+ * @cssprop --icon-color - the color of the icon
+ * @cssprop --icon-background-color - the background color of the icon
+ * @cssprop --deadline-icon-color - the color of the deadline icon
+ * @cssprop --deadline-icon-background-color - the background color of the deadline icon
```
bmesuere, line 13: Are these specified in a general css file? Why not add them here to make this component self contained?

app/assets/javascripts/components/series_icon.ts
```diff
@@ -0,0 +1,76 @@
...
+ *
+ * @attr {string} deadline-class - a css class to apply related to the deadline if any
+ * @attr {string} deadline-icon - the icon to show for the deadline if any
+ * @attr {string} overlay-class - a css class to apply to the overlay if any
+ * @attr {string} progress-class - a css class to apply that is related to the progress
+ * @attr {string} progress-icon - the icon to show that indicates the progress, defaults to mdi-school
+ * @attr {number} size - the size of the icon in pixels
+ */
+@customElement("d-series-icon")
+export class SeriesIcon extends ShadowlessLitElement {
```
bmesuere, line 24: I would refactor the properties of this component a bit since they now expose the inner working. The attributes should define the wanted state e.g. deadline = missed where the component then applies the appropriate icon and color. This would also allow to make the types stricter (enum like).

bmesuere (changes requested):

app/assets/javascripts/components/series_icon.ts
jorg-vr: Right now it is a shadowless lit component, which forces all style definitions to be in the global scope.
Changing it to use the shadow dom would not allow any use of globally defined classes. In this this would mean we had to redefine the material design icons used within this class. I do not think that is desired.

Sadly there is no way to 'partially' use the shadow dom. Which is why we also aren't using the shadow dom in most other components for now.

bmesuere (approved):

PR #5269, opened 2024-01-04
Always show navigation on submission pages

### LINKED ISSUE #5268: Navigation not available when submission is evaluated
Submissions that have been evaluated lose the option to navigate to other submissions for the same student and same exercise (in the course).

Example: https://dodona.be/nl/submissions/16468774

jorg-vr:
This pull request introduces the submission navigation on pages with published scores.

The score is now moved to the next row. It is now expanded by default and no longer toggleable as a student visiting this page will almost always want to see this information.

I made it into a card to help it stand out more. This style is similar to the feedback cards below:
image
image

In the card some extra info and a link to the series overview is provided.
On this overview page I also simplified the help texts. I also removed the "feedback" header from the central column, as that might give the impression that that column contains personalized feedback, instead of score item names.
image

Closes #5268 

**Side effect**: Due to the outlined card style change the login methods on a profile page are also styled differently. But I think its an improvement.
Old:
image
New:
image


chvp (approved):

bmesuere (changes requested):
- Always showing the previous submissions is ok for me
- Showing the score full width is ok for me
- Removing the collapse feature for the score is also ok for me

I agree that the score could used some additional styling, but I'm not sure about reusing the exact style of the feedback table since they look very similar but show something completely different. This might however be a non-issue because the score is only shown on a fraction of the pages. 

I do think we can improve of the UI. There is no context right now, just a dump of the score. The user might be confused where this comes from. Can we use a proper title instead of the score? Is there an already used icon we can add? Can we link to a page that shows the rest of this evaluation for that user? Since we have space, maybe also add a text label to the edit button (similar to the debug button)...

app/views/feedbacks/_score_card.html.erb
```diff
@@ -0,0 +1,49 @@
+<% total = policy(feedback.evaluation_exercise).show_total? %>
+<% scores = policy_scope(feedback.scores) %>
+
+
+<% if total || feedback.score_items.present? %>
+  <div class="card-outlined card scores">
+    <div class="card-title card-title-colored-container">
+      <span><%= t(".score")%></span>
```
bmesuere, line 8: I would use Evaluatie/Evaluation as title instead of punten.

config/locales/views/feedbacks/en.yml
```diff
@@ -59,12 +59,8 @@ en:
...
       total_score: Total grade
-      score_breakdown: Grade breakdown
-    score_link:
-      score: "Grade: %{score} / %{max}"
-      score_short: "%{score} / %{max}"
-      present: "There are %{count} score items"
-      view_breakdown: View grade breakdown
+      score: "Grade"
       view_feedback: Go to the feedback
+      info_html: "This solution has been evaluated by a teacher. They provided the following feedback. <a>See all evaluated solutions</a>."
```
bmesuere, line 66:
```suggestion
      info_html: "This solution has been manually evaluated by a teacher as part of <a> an evaluation</a>. Below, you can find your feedback."
```
I think the href/url is missing here.

config/locales/views/feedbacks/nl.yml
```diff
@@ -59,12 +59,8 @@ nl:
...
       total_score: Totaalscore
-      score_breakdown: Puntenverdeling
-    score_link:
-      score: "Score: %{score} / %{max}"
-      score_short: "%{score} / %{max}"
-      present: "Er zijn %{count} scoreonderdelen"
-      view_breakdown: Puntenverdeling bekijken
+      score: "Punten"
       view_feedback: Ga naar de feedback
+      info_html: "Deze oplossing werd nagekeken door een lesgever. Deze gaf onderstaande beoordeling. <a href=\"%{url}\">Bekijk de andere beoordelingen</a>."
```
bmesuere, line 66:
```suggestion
      info_html: "Deze oplossing werd manueel nagekeken door een lesgever als deel van <a href=\"%{url}\">een evaluatie</a>. Hieronder vind je jouw feedback."
```

bmesuere (changes requested):

bmesuere (approved):

PR #5272, opened 2024-01-05
Fix user label disappearing on global feedback update

### LINKED ISSUE #5252: User label in grading module sometimes disappears
It happens at least when setting all the scores to 0 at the same time, and when removing a score.
[bug.webm](https://github.com/dodona-edu/dodona/assets/42220376/935797b5-c539-4172-9240-2f6a0a9ac324)

jorg-vr:
Closes #5252


chvp (approved):

bmesuere (commented):
Can you explain a bit more? This feels like a hack to me.

jorg-vr:
When the all scores are updated (by clicking the top level thumbs up or thumbs down) the whole score card is rerendered. (the html is replaced using javascript)

This card also contains the users labels. But the view file expects these to be present as a variable.

In the show function this variable is defined in the controller. But in the js response on update the same view is reused, but the variable is never defined. Thus the view simply did not draw any labels.

Another option would be to move the function to fetch the user labels from the controller to the view file. This way the definition would only be written once. 
But it is unclear to me which best practice takes precedence here: keeping the logic in the controller or avoiding code duplication.

bmesuere:
In this case, I would:
- remove `@user_labels` as a separate variable in both actions and just fetch it from the view since it is part of the feedback object
- add a shorthand to the model to directly fetch the labels

bmesuere (approved):

PR #5273, opened 2024-01-05
Fix display of escaped characters in 'restore boilerplate'

### LINKED ISSUE #5270: "Restore the initial code" not working correctly for escape sequences
In the "Hand in" tab of an exercise, choosing the "Restore the initial code" option does not work correctly when said initial code contains escape sequences. In particular:
- `\0` is replaced by the empty string
- `\n` is replaced by a literal newline character

Reproduced locally on commit 7b1b1bc

// cc: @wkverstr

jorg-vr:
- [x] Tests were added

Closes #5270

### DIFF
--- a/app/views/activities/show.html.erb
+++ b/app/views/activities/show.html.erb
@@ -180,7 +180,7 @@ end %>
             <%= @course&.id || "null" %>,
             <%= raw "\"#{@series&.deadline&.httpdate}\"" || "null" %>,
             "<%= submissions_url %>",
-            `<%= (@activity.exercise? && @activity.boilerplate&.gsub('`', '\\\\`'))|| "" %>`,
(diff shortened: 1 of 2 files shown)


app/views/activities/show.html.erb
```diff
@@ -180,7 +180,7 @@ end %>
             <%= @course&.id || "null" %>,
             <%= raw "\"#{@series&.deadline&.httpdate}\"" || "null" %>,
             "<%= submissions_url %>",
-            `<%= (@activity.exercise? && @activity.boilerplate&.gsub('`', '\\\\`'))|| "" %>`,
```
niknetniko, line 183: Would it be cleaner (or possible) to use [`escape_javascript`](https://api.rubyonrails.org/classes/ActionView/Helpers/JavaScriptHelper.html#method-i-escape_javascript) here, instead of doing this manually?

bmesuere (approved):
👍

niknetniko (approved):

PR #5274, opened 2024-01-05
Translate exercise names in search results

### LINKED ISSUE #5254: Translate name of exercise in overview of saved annotations
When language setting is Dutch, the names of the exercises still appear in English in the overview of saved annotations:

image

jorg-vr:
This pull request fixes exercise names appearing in the wrong language when searching through saved annotations.

- [x] Tests were added

Closes #5254

### DIFF
--- a/app/assets/javascripts/state/SavedAnnotations.ts
+++ b/app/assets/javascripts/state/SavedAnnotations.ts
@@ -34,6 +34,10 @@ class SavedAnnotationState extends State {
     @stateProperty private paginationByURL = new StateMap<string, Pagination>();
     @stateProperty private byId = new StateMap<number, SavedAnnotation>();
 
+    private get url(): string {
+        return `/${I18n.locale}${URL}`;
+    }
+
     private async fetchList(url: string): Promise<Array<SavedAnnotation>> {
         const response = await fetch(url);
         this.listByURL.set(url, await response.json());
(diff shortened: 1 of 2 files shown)


chvp (approved):

app/assets/javascripts/state/SavedAnnotations.ts
```diff
@@ -34,6 +34,10 @@ class SavedAnnotationState extends State {
     @stateProperty private paginationByURL = new StateMap<string, Pagination>();
     @stateProperty private byId = new StateMap<number, SavedAnnotation>();
 
+    private get url(): string {
+        return `/${I18n.locale}${URL}`;
```
bmesuere, line 38: Can't we use I18n in a cleaner way? This is a hidden dependency which relies on a global variable being set without any fallback.

jorg-vr: The way I18n is handled is indeed legacy.
If we change how it  is used, we should probably do a refactor of all usages.

Do you want me to update all in this pr or make a new pr?

bmesuere: Since is has nothing to do with the rest of this fix, I think it's best done in a separate PR.

Feel free to check if there is similar technical debt we could fix.

bmesuere (approved):

PR #5275, opened 2024-01-05
Fix redirect to series show from overview page of hidden series

### LINKED ISSUE #5255: Returning to hidden series from "all descriptions" fails
For exercise-series, there is an option to "Show all learning activity descriptions":
image

Clicking this you get to a page with the link: "https://dodona.be/en/series/<SERIES_ID>/overview/?token=<TOKEN>&user_id=<USER_ID>"
At the top of this page, in the light-blue "menu"-bar, there is an indication on what you are currently viewing, and each thing is clickable to return to that specific thing:
image
 
The issue appears when you want to go back to the hidden exercise-series. Clicking on it in the blue menu-bar does not take you back to that series, but instead to the course-page (with all its non-hidden exercise-series). 
I would think that it doesn't use the access-token to go back to that series, so it doesn't have permission to view it, and thus returns to the course page. 
This could be a bug introduced by [ pr "Require valid token to view series activities" #5249  ](https://github.com/dodona-edu/dodona/pull/5249).

jorg-vr:
This pull request provides a redirect to the hidden series with a valid token from the overview page of a hidden series.

Closes #5255

### DIFF
--- a/app/controllers/series_controller.rb
+++ b/app/controllers/series_controller.rb
@@ -71,7 +71,9 @@ class SeriesController < ApplicationController
   def overview
     @title = "#{@series.course.name} #{@series.name}"
     @course = @series.course
-    @crumbs = [[@course.name, course_path(@course)], [@series.name, course_path(@series.course, anchor: @series.anchor)], [I18n.t('crumbs.overview'), '#']]
+    # redirect to series detail page if a token is provided, as the course page wont contain the hidden series even with a valid token
+    series_link = params[:token].present? ? series_path(@series, token: params[:token]) : course_path(@series.course, anchor: @series.anchor)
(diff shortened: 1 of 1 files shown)


app/controllers/series_controller.rb
```diff
@@ -71,7 +71,9 @@ def show
   def overview
     @title = "#{@series.course.name} #{@series.name}"
     @course = @series.course
-    @crumbs = [[@course.name, course_path(@course)], [@series.name, course_path(@series.course, anchor: @series.anchor)], [I18n.t('crumbs.overview'), '#']]
+    # redirect to series detail page if a token is provided, as the course page wont contain the hidden series even with a valid token
```
bmesuere, line 74:
```suggestion
    # If a token is provided, this is a hidden series that isn't shown on the course page. 
    # Therefore, link to the series page instead of the course page from the breadcrumbs.
```

bmesuere (approved):

PR #5276, opened 2024-01-05
Fix ctrl click for notifications

### LINKED ISSUE #5247: Ctrl + click'ing a notification opens the link in both current and new tab
I had a new notification today, and wanted to open it in a new tab. On many websites this can be done by pressing ctrl + clicking the link, click it with the third mouse-button (often the mouse-wheel), or right-click and select "Open link in new tab".

The last 2 options work as expected, but ctrl + click 'ing the link opens the link in the current open tab and in a new tab.

jorg-vr:
This pull request fixes #5247 if the actual link part of the notification is clicked. (quite common)

It is not solved in al cases as that would increase complexity to much and be prone to errors due to different browser behavior. See https://github.com/dodona-edu/dodona/pull/5276/commits/9f592aa7a834ad4362ef05e3f8f14991326abbfb and https://robertcooper.me/post/table-row-links

Closes #5247

### DIFF
--- a/app/assets/javascripts/notification.ts
+++ b/app/assets/javascripts/notification.ts
@@ -42,7 +42,7 @@ export class Notification {
         // We only want to install the click handler for the full element on the small notification view.
         if (installClickHandler) {
             this.element.addEventListener("click", event => {
-                // Use default browse behavior if the user clicked on the link itself.
+                // Use default browser behavior if the user clicked on the link itself.
                 if (event.target instanceof HTMLAnchorElement) return;
 
                 this.visit();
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/notification.ts
```diff
@@ -42,6 +42,9 @@ export class Notification {
         // We only want to install the click handler for the full element on the small notification view.
         if (installClickHandler) {
             this.element.addEventListener("click", event => {
+                // Use default browse behavior if the user clicked on the link itself.
```
chvp, line 45:
```suggestion
                // Use default browser behavior if the user clicked on the link itself.
```

chvp (approved):

bmesuere (approved):

PR #5277, opened 2024-01-06
Update API - submissions.json

### LINKED ISSUE #5248: Discrepancy between json and html view of /submissions
When using the provided API, I stumbled upon an issue with submissions. Some activities are set to private within a course, but get a link pointing to it without a reference to it's associated course. Trying to view this private activity (as a student) returns error 403 (forbidden). 
This issue does not occur when viewing the submissions through the browser, as their links correctly refer to the course.

This occurs both on the "global" submissions for my user-account, and for the submissions for a certain activity. Note however that on the html-version of activity-submissions their is no link included to the activity. This is only on the "global" submissions.

The json-version also does not include the name of the corresponding activity, while the html-one does. It would be a nice improvement if the API included the name as well. It would also reduce the amount of "GET" requests needed to get the info for submissions.

In the html-version:
image

In the json-version:
```
... 
{
    "created_at":"2023-12-15T11:14:20.000+01:00",
    "status":"correct","summary":"Correct",
    "accepted":true,"id":16399349,
    "url":"https://dodona.be/nl/submissions/16399349.json",
    "user":"https://dodona.be/nl/courses/2876/members/37325.json",
    "has_annotations":false,

    "exercise":"https://dodona.be/nl/activities/273501711.json",

    "course":"https://dodona.be/nl/courses/2876.json"
}, ....
```

BWindey:
This pull request aims to fix the discrepancy between the html and json results for user submissions (not tied to single exercise), as mentioned in #5248 .

This updates dodona/app/views/_submission_basic.json.jsonbuilder to include the same logic found in _submission.html.erb (same directory). 

First tests seem to indicate that this is indeed working, but I don't have access to the live production server to test the same case that caused #5248 .

- [ ] Tests were added

Closes #5248  .

### DIFF
--- a/app/views/submissions/_submission_basic.json.jbuilder
+++ b/app/views/submissions/_submission_basic.json.jbuilder
@@ -6,5 +6,11 @@ else
   json.user user_url(submission.user.id, format: :json)
 end
 json.has_annotations submission.annotated?
-json.exercise activity_url(submission.exercise, format: :json)
+
+if submission.course.nil?
+  json.exercise activity_url(submission.exercise, format: :json)
+else
+  json.exercise course_activity_path(submission.course, submission.exercise, format: :json)
+end
+
 json.course course_url(submission.course, format: :json) if submission.course
--- a/app/views/submissions/_submission_basic.json.jbuilder
+++ b/app/views/submissions/_submission_basic.json.jbuilder
@@ -10,7 +10,7 @@ json.has_annotations submission.annotated?
 if submission.course.nil?
   json.exercise activity_url(submission.exercise, format: :json)
 else
-  json.exercise course_activity_path(submission.course, submission.exercise, format: :json)
+  json.exercise course_activity_url(submission.course, submission.exercise, format: :json)
 end
 
 json.course course_url(submission.course, format: :json) if submission.course


BWindey:
I took the freedom to include the name of exercises in the result of a GET request to "/submissions.json". In the html version, this name is already displayed, but the json version did not have this. 

This might be something you don't want to add, or maybe just not together with the changes above mentioned. If you would like this to be removed, or have its own pull request, I'd gladly (re)move this change. 

My reasoning for adding this:
If a user of the api wants to get all their submissions and display each submission with the exercise-name, they currently have to make a GET request per submission, resulting in up to 31 requests needed to display this. With the exercise-name included, the needed info can be displayed in 1 request.

Adding extra info the api might not always be a good idea though. It can introduce unwanted overhead, but I still believe this is a good addition as it:
- can reduce the number of api calls needed
- doesn't need an extra query on the database, as the needed info already resides in the submission-object used

On top of that I did some speed-comparisons, which you can read here:
```
For the manual tests I left out the first request as it took about 170ms
instead of around 50ms. 
The request always returned 4 submissions for my test-account.

The first two tests test the server-side performance (as they take their timings
from the server-log-messages) while the two last tests test client-side speed.

- Manual test by reloading the submissions.json page in the brower
57 times for both configurations. Averages were calculated by taking
the output of the log-messages in the terminal
      "Completed 200 OK in ...ms ..."
adding the X milliseconds and dividing by 57.
Without exercise-name:
      40,543859649ms per GET
With exercise-name:
      39,140350877ms per GET

- Bash script tests with log-files and computing average like first test
Test ran for 300 seconds, and used curl in a random interval between 0.1s and 5.09s
Without exercise-name:
      40,985294118
With exercise-name:
      38,744525547

- Tests with hyperfine, 500 runs each
Tests were executed by calling curl ".../submissions.json" (and headers)
Without exercise-name:
      Time (mean ± σ):      51.0 ms ±  10.8 ms    [User: 5.4 ms, System: 3.6 ms]
      Range (min … max):    38.4 ms … 186.3 ms    500 runs
With exercise-name:
      Time (mean ± σ):      51.6 ms ±   8.9 ms    [User: 5.4 ms, System: 3.3 ms]
      Range (min … max):    39.6 ms … 138.7 ms    500 runs

- Tests with hyperfine, 5000 runs each
Without exercise-name:
      Time (mean ± σ):      49.6 ms ±   9.1 ms    [User: 6.1 ms, System: 3.5 ms]
      Range (min … max):    35.8 ms … 175.6 ms    5000 runs
With exercise-name:
      Time (mean ± σ):      49.9 ms ±   9.6 ms    [User: 5.9 ms, System: 3.5 ms]
      Range (min … max):    36.9 ms … 180.5 ms    5000 runs
``` 

These tests seem to indicate that adding an extra field to the api does not impact the performance of the server in a noticeable way.

Altough this testing might be a bit too much, as I don't think this specific request will be made often.

BWindey:
I also think that the API can use some updates (courses still have colors tied to them), is there somewhere that can be discussed? Or is that something you want to work on/decided internally?

jorg-vr (approved):
Looks good. Thanks for the contribution and for including the timing test results.

app/views/submissions/_submission_basic.json.jbuilder
```diff
@@ -6,5 +6,14 @@ else
   json.user user_url(submission.user.id, format: :json)
 end
 json.has_annotations submission.annotated?
-json.exercise activity_url(submission.exercise, format: :json)
+
+if submission.course.nil?
+  json.exercise activity_url(submission.exercise, format: :json)
+else
+  json.exercise course_activity_url(submission.course, submission.exercise, format: :json)
+end
```
niknetniko, line 14:
```suggestion
json.exercise activity_scoped_url(activity: submission.exercise, course: submission.course, options: { format: :json })
```

I think we can achieve the same by re-using this existing method? That way we don't need to dirty our views with more conditionals than necessary.

app/views/submissions/_submission_basic.json.jbuilder
```diff
@@ -6,5 +6,14 @@ else
...
+
+json.exercise_name submission.exercise.name
+
+
```
niknetniko, line 18: Why are there two newlines here?

niknetniko (changes requested):

app/views/submissions/_submission_basic.json.jbuilder
BWindey: That's messy indeed. Do I remove all new-lines except the last one (EOF)?

jorg-vr: yes

PR #5288, opened 2024-01-13
Update API - notifications.json

### LINKED ISSUE #5286: API /notifications.json has field ..._url but shows ..._path
When looking at the output of [dodona.be/notifications.json](https://dodona.be/notifications.json), I notice that the field "notifiable_url" does not contain a url, but a path (doesn't contain "https://dodona.be"). 
This is inconsistent with other json pages.

image
The 2 files that create the result:
https://github.com/dodona-edu/dodona/blob/main/app/views/notifications/_notification.json.jbuilder#3
https://github.com/dodona-edu/dodona/blob/main/app/helpers/notifications_helper.rb#9

BWindey:
This pull request fixes the API returning a path for a variable called notifiable_url, in notifications.json. 

Before:
```
[
  {
    "id": 17,
    ...
    "notifiable_url": "/nl/submissions/2441/#code"
  }
]

```

After:
```
[
  {
    "id": 17,
    ...
    "notifiable_path": "/nl/submissions/2441/#code",
    "notifiable_url": "http://dodona.localhost:3000/nl/submissions/2441.json#code"
  }
]
```

I'm not sure which one is preferred, the path or the url. I've noticed that paths are rarely used in the current state of the API, while url's appear often, if not always. 

- [ ] Tests were added

Closes #5286 .


jorg-vr:
This api variable might be used. These usecases might be broken by this update. Did you check for usages in the Dodona codebase?

We should probably also check for usecases in the VSCode and jetbrains plugins. I know at least one of them checks for notifications

BWindey:
I didn't think of the codebase using those functions. That is now fixed (it was used in 4 places, according to RubyMine).

I will see if I can find the usage of notifiable_url in the VSCode and Jetbrains plugins.

BWindey:
I thought checks didn't run automatically when someone outside the Dodona team pushed a commit to their PR, and it's even on a draft PR?

PR #5289, opened 2024-01-13
Remove "actions" for labels for users who can't do actions on labels

BWindey:
This pull request removes the actions-column entirely for users who don't have the right to perform actions on labels.

### Pc normal window
Before:
image
After:
image

### Small window 
(simulated by making window as small as possible and zooming in to 130%)
| Before: | After: |
| ----------- | --------- |
image |image

### Pc normal window
Before:
image
After:
image

### Small window
(simulated by making window as small as possible and zooming in to 130%)
| Before: | After: |
| ----------- | -------- |
image | image

### DIFF
--- a/app/views/labels/_labels_table.html.erb
+++ b/app/views/labels/_labels_table.html.erb
@@ -1,15 +1,22 @@
 <div class="table-scroll-wrapper">
   <table class="table table-index table-resource">
     <colgroup>
-      <col class="col-8"/>
-      <col class="col-3"/>
-      <col class="col-1"/>
+      <% if policy(Label).edit? %>
+        <col class="col-8"/>
+        <col class="col-3"/>
+        <col class="col-1"/>
+      <% else %>
+        <col class="col-10"/>
+        <col class="col-3"/>
+      <% end %>
     </colgroup>
     <thead>
     <tr>
       <th><%= t "labels.index.name" %></th>
       <th><%= t "labels.index.exercises" %></th>
-      <th><%= t "labels.index.actions" %></th>
+      <% if policy(Label).edit? %>
+        <th><%= t "labels.index.actions" %></th>
+      <% end %>
     </tr>
     </thead>
     <tbody>
--- a/app/views/labels/_labels_table.html.erb
+++ b/app/views/labels/_labels_table.html.erb
@@ -7,7 +7,7 @@
         <col class="col-1"/>
       <% else %>
         <col class="col-10"/>
-        <col class="col-3"/>
+        <col class="col-2"/>
       <% end %>
     </colgroup>
     <thead>
--- a/app/views/labels/_labels_table.html.erb
+++ b/app/views/labels/_labels_table.html.erb
@@ -1,7 +1,7 @@
 <div class="table-scroll-wrapper">
   <table class="table table-index table-resource">
     <colgroup>
-      <% if policy(Label).edit? %>
+      <% if policy(Label).edit? or policy(Label).destroy? %>
         <col class="col-6"/>
         <col class="col-3"/>
         <col class="col-3"/>
@@ -14,8 +14,8 @@
     <tr>
       <th><%= t "labels.index.name" %></th>
       <th class="text-end"><%= t "labels.index.exercises" %></th>
-      <% if policy(Label).edit? %>
-        <th class="text-end"><%= t "labels.index.actions" %></th>
+      <% if policy(Label).edit? or policy(Label).destroy? %>
+        <th class="text-end"></th>
       <% end %>
     </tr>
     </thead>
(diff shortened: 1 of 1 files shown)


app/views/labels/_labels_table.html.erb
```diff
@@ -1,15 +1,22 @@
...
     <colgroup>
-      <col class="col-8"/>
-      <col class="col-3"/>
-      <col class="col-1"/>
+      <% if policy(Label).edit? %>
+        <col class="col-8"/>
+        <col class="col-3"/>
+        <col class="col-1"/>
+      <% else %>
+        <col class="col-10"/>
```
chvp, line 9:
```suggestion
        <col class="col-9"/>
```
The column widths should add up to 12 in bootstrap. See https://getbootstrap.com/docs/5.0/layout/columns/ and https://getbootstrap.com/docs/5.0/layout/grid/.

chvp (changes requested):
Thanks for your contribution! Just a small remark on your code.

app/views/labels/_labels_table.html.erb
BWindey: Those are really useful links! I now understand more of what's what. I tested changing it to 9, and also changing the other column to 2, to see which one looks better. Personally I prefer using "10 - 2". Do you agree?  
10 - 2
image
9-3
image

chvp: I have no particular preference. Maybe try aligning the second column to the right, given its numeric nature that would probably look good.

BWindey: I changed the exercise-count righ-aligned, and the actions too. The latter had to be right-aligned to keep some distance between the count and the buttons.
image
image
image

jorg-vr (approved):
Thanks for the contribution.

Could you put the latest images in the pr description?

Also did you test this on smaller screen widths? The 9-3 vs 10-2 discussion might be more relevant in those cases

BWindey:
Updated the original PR to include the images

app/views/labels/_labels_table.html.erb
```diff
@@ -1,23 +1,30 @@
...
+      <% end %>
     </colgroup>
     <thead>
     <tr>
       <th><%= t "labels.index.name" %></th>
-      <th><%= t "labels.index.exercises" %></th>
-      <th><%= t "labels.index.actions" %></th>
+      <th class="text-end"><%= t "labels.index.exercises" %></th>
+      <% if policy(Label).edit? %>
+        <th class="text-end"><%= t "labels.index.actions" %></th>
```
bmesuere, line 18:
```suggestion
        <th class="text-end"></th>
```

The label is redundant, so I would remove it.

bmesuere (approved):

app/views/labels/_labels_table.html.erb
BWindey: That would remove the "Actions" title, correct?
image

BWindey: If so, then we could remove the conditional entirely for the tabel header.

bmesuere: Indeed, it would remove the Actions label. Removing the conditional for the table header is not a good idea. You should have the same number of header columns as body columns.

BWindey: That seems logical, ok.

I just noticed something too. There are always 3 body columns, as the conditional is seperated into permission to edit and permission to destroy. Both checks happen inside the <td> column tag. 

How does that work with there only being 2 columns defined (for users without permission), but 3 actual columns, the last one empty?

```html
...
  <colgroup>
      <% if policy(Label).edit? %>
        <col class="col-6"/>
        <col class="col-3"/>
        <col class="col-3"/>
      <% else %>
        <col class="col-10"/>
        <col class="col-2"/>
      <% end %>
    </colgroup>
  
  ...
  <tbody>
    <% labels.each do |label| %>
      <tr>
        <td><%= link_to label.name, label_path(label) %></td>
        <td class="text-end"><%= link_to label.activities.count, activities_path({"labels[]" => label.name}) %></td>
        <td class="text-end">
          <% if policy(Label).edit? %>
            <%= link_to edit_label_path(label), class: "btn btn-icon" do %>
              <i class="mdi mdi-pencil"></i>
            <% end %>
          <% end %>
          <% if policy(Label).destroy? %>
            <%= link_to label, method: :delete, data: {confirm: t("general.are_you_sure")}, class: "btn btn-icon btn-icon-filled d-btn-danger" do %>
              <i class="mdi mdi-delete"></i>
            <% end %>
          <% end %>
        </td>
      </tr>
    <% end %>
  </tbody>

...
```

chvp (approved):

PR #5301, opened 2024-01-16
Extend allowed tags and attributes (svg)

tibvdm:
This pull request adds additional SVG tags:
`<defs>` `<use>`

And attributes:
`color`

- [x] Tests were added

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -146,10 +146,10 @@ module ApplicationHelper
   def sanitize(html)
     @tags ||= Rails::Html::SafeListSanitizer.allowed_tags.to_a +
               %w[table thead tbody tr td th colgroup col style summary details] +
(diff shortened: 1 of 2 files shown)


chvp (approved):

bmesuere:
Is this really needed? There is a reason we filter svg's and `use` is one of the tricky elements which can be exploited.

pdawyndt:
@bmesuere We have an assignment in today's exam, where the goal is to generate an SVG that makes use of the `use`-tag. The SVG contains repeated patterns, and the `use`-tag is used to keep the generated output as small as possible. When the output grows too large, Dodona no longer diffs the output and it becomes really hard to see for the students where some (possible small) differences are located.

bmesuere:
I that case is propose to allow it for now, but disable it again after grading is done.

PR #5305, opened 2024-01-18
Refactor i18n to not use global scope

jorg-vr:
This pull request removes the `I18n` object from the global scope.
Instead the `i18n` object can be imported from `i18n/i18n`.

It also introduces the `DodonaElement` as a new meta webcomponent used to inherit from.
This component is a combination of the `ShadowlessLitElement`, the `StateController` and the `i18nMixin`.
A lot of custom components used at leased two of the three parts, so making a generic combined class made it easier to include all. I particularly introduced it in this pr because multiple web components used `i18n` without the `i18nMixin` which might result in untranslated elements on the page.

I also made some minor changes in the `I18n` class to fix typing issues:
 - I renamed the accessor `locale` to `loc` as `locale` is defined as a function in `Polyglot`, which we extend. 
 - I removed the usage of `super.warn`, as this is a private function in `Polyglot`. I instead directly called the `warning` method they used directly.


bmesuere (approved):

app/assets/javascripts/components/meta/dodona_element.ts
```diff
@@ -0,0 +1,10 @@
+import { i18nMixin } from "components/meta/i18n_mixin";
+import { ShadowlessLitElement } from "components/meta/shadowless_lit_element";
+import { StateController } from "state/state_system/StateController";
+
+export class DodonaElement extends i18nMixin(ShadowlessLitElement) {
```
chvp, line 5: Both this and `ShadowlessLitElement` wrap `i18nMixin` when extending their superclass. Is that really necessary? I would think that it's redundant?

app/assets/javascripts/components/copy_button.ts
```diff
@@ -53,7 +54,7 @@ export class CopyButton extends ShadowlessLitElement {
     constructor() {
         super();
 
-        // Reload when I18n is loaded
+        // Reload when i18n is loaded
         ready.then(() => this.requestUpdate());
     }
```
chvp, line 59: In other cases (e.g. `CourseLabelsSearchBar`) this is removed. Is there a reason for this inconsistency?

app/assets/javascripts/i18n/i18n.ts
```diff
@@ -23,41 +24,47 @@ export class I18n extends Polyglot {
...
     }
 
-    get locale(): string {
-        return super.locale();
+    get loc(): string {
+        return this.locale();
+    }
+
+    set loc(locale: string) {
+        this.locale(locale);
```
chvp, line 43: This wrapper seems a bit unnecessary to me? Or am I wrong in thinking that every use of `i18n.loc` could be replace with `i18n.locale()`?

app/assets/javascripts/components/meta/dodona_element.ts
jorg-vr: I forgot to remove it in shadowless lit element, will fix

app/assets/javascripts/components/copy_button.ts
jorg-vr: No, will also remove it here

app/assets/javascripts/i18n/i18n.ts
jorg-vr: I have a personal preference for the getter, setter approach, but you're not wrong.

I'll remove the wrapper and update the usages

chvp: I don't necessarily have a strong opinion about this, but the `loc` name is less clear than `locale`, IMO.

jorg-vr: I removed this getter,setter because I couldn't think of a better name and it might not be bad to remain closer to the polyglot api

chvp (approved):

PR #5306, opened 2024-01-20
Swap order of status icon in series-activity-table

### LINKED ISSUE #5238: Swap order of status-icon and draft-icon in left margin of series
The icons would be aligned better if the order of the status-icon and the draft-icon were swapped in the left margin of a series.

Current order:

image

Preferred order:

image

BWindey:
This pull request reverses the order of the activity-status (solved/wrong) and the draft-icon, as mentioned in #5238 

image

I have no idea if there are tests related to this, I tried looking through the tests/ folder, but didn't see anything that I could relate to this change.
- [ ] Tests were added

Closes #5238  .


BWindey:
Oh, could it be that I'm marked as "contributor", so tests run automatically for my pull requests now too? It also didn't automatically call for reviewers.

BWindey:
Issue doesn't have a priority assigned to it.

PR #5307, opened 2024-01-20
Revert ChatGPT "motivational" messages

### LINKED ISSUE #4598: Revert to original motivational messages
More info at [Revert to original motivational messages #4595](https://github.com/orgs/dodona-edu/discussions/4595)

BWindey:
This pull request reverts the addition of addition of messages written by ChatGPT, as mentioned in #4598 .
This is a simple removal, but maybe you would prefer to have a setting that toggles those messages? If so, I can look into that, but it'll require a bit more work.

I've looked through the ChatGPT messages to see if they had parts that would fit as additional messages that don't seem weird, but I've noticed that they're either not fit (in my eyes), or (for the Dutch version) are the same as the old messages, but with an additional "you solved this exercise", which is already indicated by all the rewarding green you see.

- [ ] Tests were added

Closes #4598 .


bmesuere:
We haven't yet decided in which direction we want to go with this, so I'll close this PR.

To prevent this from happening in the future, first ask if it's ok to go ahead and implement an issue and/or ignore issues without a priority assigned to them.

PR #5308, opened 2024-01-22
Add basic styling for details html element

### LINKED ISSUE #5237: Provide default formatting of <details> HTML element in descriptions
**Problem**: When using the HTML element `<details>` in descriptions of assignments as a replacement of the Bootstrap accordeon element, its representation does not look very decorated by default.

**Solution**: Add CSS that makes the HTML element `<details>` in the description of assignments look good by default, without the need to add specific CSS in the description itself. We could then also advertise the use of this element in the [exercise descriptions](https://docs.dodona.be/en/references/exercise-description/) reference tutorial (HTML/MarkDown).

jorg-vr:
This pull request adds basic styling for the details html element

image
image

- [x] Documentation update can be found at dodona-edu/dodona-edu.github.io#383

Closes #5237

### DIFF
--- a/app/assets/stylesheets/components.css.scss
+++ b/app/assets/stylesheets/components.css.scss
@@ -27,6 +27,7 @@
 @import "components/courselabel-edit.css.scss";
 @import "components/action-card.css.scss";
 @import "components/progress_bar.css.scss";
+@import "components/details.css.scss";
 
 .flex-spacer {
   flex-grow: 1;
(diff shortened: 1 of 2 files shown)


bmesuere (changes requested):
I would add a bit of bottom margin to the details element (maybe the same value as for paragraphs?). Right now, it's a bit close to the bottom paragraph:
image

bmesuere (approved):

PR #5320, opened 2024-01-29
Don't show zeus messages on submissions when in demo mode

jorg-vr:
This pull request removes zeus messages on submission when in demo mode.

Before
image

After
image


niknetniko (approved):
> Another more generic and impactful way to fix this: Remove all zeus rights when in demo mode by overwriting user.zeus?. But this might be to risky if there is no actual need for this feature.

I have no opinion on this.

bmesuere (approved):

niknetniko (approved):

PR #5330, opened 2024-01-29
Require at least one correct submission and a valid config before publishing

### LINKED ISSUE #5084: Check valid configuration when publishing draft exercise
When publishing a draft exercise, we could check if the configuration of the exercise is valid. This would prevent publishing of invalid draft exercises.

jorg-vr:
This pull request adds a validation to activities that checks if an activity has at least one correct submission and a valid config before publishing.

This info is displayed to the user in the notice.

image

- [x] Tests were added

Closes  #5084

### DIFF
--- a/app/models/activity.rb
+++ b/app/models/activity.rb
@@ -57,7 +57,7 @@ class Activity < ApplicationRecord
   has_many :labels, through: :activity_labels
 
   validates :path, uniqueness: { scope: :repository_id, case_sensitive: false }, allow_nil: true
-  validate :require_valid_submission_before_publish, on: :update
+  validate :require_correct_submission_before_publish, on: :update
 
   token_generator :repository_token, length: 64
   token_generator :access_token
@@ -512,18 +512,18 @@ class Activity < ApplicationRecord
     hash
   end
 
-  def valid_submission?
+  def correct_submission?
     return true if content_page?
 
     submissions.any?(&:accepted?)
   end
 
-  def require_valid_submission_before_publish
+  def require_correct_submission_before_publish
     return if draft?
     return unless draft_was
-    return if valid_submission?
+    return if correct_submission?
 
-    errors.add(:base, I18n.t('activerecord.errors.models.activity.no_valid_submission'))
+    errors.add(:base, I18n.t('activerecord.errors.models.activity.no_correct_submission'))
     self.draft = true
   end
 end
--- a/app/models/activity.rb
+++ b/app/models/activity.rb
@@ -523,7 +523,7 @@ class Activity < ApplicationRecord
     return unless draft_was
     return if valid_submission?
 
-    errors.add(:draft, I18n.t('activerecord.errors.models.activity.no_valid_submission'))
+    errors.add(:base, I18n.t('activerecord.errors.models.activity.no_valid_submission'))
     self.draft = true
   end
 end
--- a/app/models/activity.rb
+++ b/app/models/activity.rb
@@ -57,7 +57,7 @@ class Activity < ApplicationRecord
   has_many :labels, through: :activity_labels
 
   validates :path, uniqueness: { scope: :repository_id, case_sensitive: false }, allow_nil: true
-  validate :require_valid_submission_before_publish
+  validate :require_valid_submission_before_publish, on: :update
 
   token_generator :repository_token, length: 64
   token_generator :access_token
--- a/app/models/activity.rb
+++ b/app/models/activity.rb
@@ -57,6 +57,7 @@ class Activity < ApplicationRecord
   has_many :labels, through: :activity_labels
 
   validates :path, uniqueness: { scope: :repository_id, case_sensitive: false }, allow_nil: true
+  validate :require_valid_submission_before_publish
 
   token_generator :repository_token, length: 64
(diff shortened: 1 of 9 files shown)


app/models/activity.rb
```diff
@@ -510,4 +511,19 @@ def unique_labels(hash)
...
+    submissions.any?(&:accepted?)
+  end
+
+  def require_valid_submission_before_publish
+    return if draft?
+    return unless draft_was
+    return if valid_submission?
+
+    errors.add(:base, I18n.t('activerecord.errors.models.activity.no_valid_submission'))
+    self.draft = true
```
chvp, line 527: This shouldn't be necessary. If errors are added during validation, the record isn't saved to the database.

config/locales/models/nl.yml
```diff
@@ -2,6 +2,8 @@ nl:
   activerecord:
     errors:
       models:
+        activity:
+          no_valid_submission: "Er moet minstens één geldige oplossing zijn, vooraleer de oefening kan worden gepubliceerd."
```
chvp, line 6:
```suggestion
          no_valid_submission: "Er moet minstens één correcte oplossing zijn, vooraleer de oefening kan worden gepubliceerd."
```

chvp (changes requested):
- [x] In general I would not use the `valid` terminology, but use `correct` everywhere.

app/models/activity.rb
jorg-vr: It isn't saved, but the unsaved activity is reused to re render the edit screen with the same invalid state and the corresponding errors.
For the case of draft this is odd as this hides the draft notice on top...

chvp (approved):

bmesuere (changes requested):
In the current implementation, the check is only done when trying to publish. I would do this every time when we render the "this exercise is a draft" callout and add some kind of checklist there so it is clear which steps need to be taken, like
- have a problem description in Dutch and/or English
- have a valid configuration
- have at least one correct submission

In the future, we could expand that list if we would think of other requirements.

config/locales/views/activities/en.yml
```diff
@@ -101,8 +101,13 @@ en:
...
       preloaded_restore: Restore the initial code.
-      alert_draft_html: This activity is a <a href="https://docs.dodona.be/en/faq/activities/#what-is-a-draft-activity" target="_blank">draft</a> and thus not visible for students.
-      alert_draft_edit: Publish activity.
+      alert_draft:
+        text_html: This activity is a <a href="https://docs.dodona.be/en/faq/activities/#what-is-a-draft-activity" target="_blank">draft</a> and thus not visible for students.
+        edit: Publish activity.
+        before_publish: "Before you can publish this activity, the following conditions must be met:"
+        valid_config: "The JSON configuration must be valid."
+        is_valid: "The activity must have a name and a description."
+        correct_submission: "There must be at least one correct submission."
```
bmesuere, line 110:
```suggestion
        text_html: This activity is currently a <a href="https://docs.dodona.be/en/faq/activities/#what-is-a-draft-activity" target="_blank">draft</a> and is not yet visible to students.
        edit: Publish activity.
        before_publish: "Before you can publish this activity, please ensure the following criteria are met:"
        valid_config: "The exercise must have a valid configuration file."
        is_valid: "The activity must have a name and a description."
        correct_submission: "You must submit at least one correct solution."
```

config/locales/views/activities/nl.yml
```diff
@@ -101,8 +101,13 @@ nl:
...
       preloaded_restore: Zet de voorbeeldcode terug.
-      alert_draft_html: Deze oefening is nog een <a href="https://docs.dodona.be/nl/faq/activities/#wat-is-een-conceptactiviteit" target="_blank">concept</a>. Ze is niet zichtbaar voor studenten.
-      alert_draft_edit: Deze oefening publiceren.
+      alert_draft:
+        text_html: Deze oefening is nog een <a href="https://docs.dodona.be/nl/faq/activities/#wat-is-een-conceptactiviteit" target="_blank">concept</a>. Ze is niet zichtbaar voor studenten.
+        edit: Deze oefening publiceren.
+        before_publish: "Vooraleer je deze oefening kan publiceren, moet aan de volgende voorwaarden voldaan zijn:"
+        valid_config: "De JSON-configuratie moet geldig zijn."
+        is_valid: "De oefening moet een naam en een beschrijving hebben."
+        correct_submission: "Er moet minstens één correcte oplossing zijn."
```
bmesuere, line 110:
```suggestion
        text_html: Deze oefening is momenteel een <a href="https://docs.dodona.be/nl/faq/activities/#wat-is-een-conceptactiviteit" target="_blank">concept</a> en nog niet zichtbaar voor studenten.
        edit: Deze oefening publiceren.
        before_publish: "Vooraleer je deze oefening kan publiceren, moet aan de volgende voorwaarden voldaan zijn:"
        valid_config: "De oefening moet een geldig configuratiebestand hebben."
        is_valid: "De oefening moet een naam en een beschrijving hebben."
        correct_submission: "Je moet minstens één correcte oplossing indienen."
```

bmesuere (approved):

chvp (approved):

Issue #5014, opened 2023-09-27
Integrate pyodide tutor


jorg-vr:
The current python tutor implementation only works for the old python judge and requires backend support.

In https://github.com/dodona-edu/pyodide-tutor we have tested the usage of the python tutor with pyodide. This works well.
We should integrate this project in Dodona to support python debugging for all python judges.

### Questions of note:
The developer of the python tutor ([Philip Guo](https://pg.ucsd.edu/)) has made his [repo](https://github.com/pgbovine/OnlinePythonTutor/) private. What does this mean for the future development of the project?

A note from the developper on this from [here](https://docs.google.com/document/d/1quk8gdvgzaYrZaOSiVzCia8tCXBFmuTk0awMw_-OTuE/edit):
> Q: Can I run a local copy of Python Tutor on my own server? I want better reliability guarantees. Or I want to customize a version to integrate with my own course’s software or materials.
>
>A: Sorry, the code is not openly available. We do not have the resources to maintain a robust open-source codebase with the intricate customizations required to run a modern scalable web service. Also, we do not have the time to help install, configure, and customize the service on your end. Simply giving you the source code is not enough to get it up-and-running, since so much of the code is tied to our own custom infrastructure. If you want to integrate Python Tutor into your own software or materials, the [iframe embed technique](https://docs.google.com/document/d/1quk8gdvgzaYrZaOSiVzCia8tCXBFmuTk0awMw_-OTuE/edit#bookmark=id.pbz67veiykx5) is by far the best way to do so.

Both this project and https://github.com/dodona-edu/pyodide-tutor contain a local copy of the python tutor javascript, but in a different version. I assume we'll want to use https://github.com/dodona-edu/pyodide-tutor/blob/main/src/pytutor.ts as it seems most recent?

As the developer repo is private, we'll either have to contact him to keep up to date with changes, or claim our own responsibility to keep the project up to date.

Eg: Right now there is still a heavy dependency on jquery and everything is in a single file, if we claim ownership, we should probably clean up this code to make it more maintainable?

For future reference: [notes on the state and vision of the python tutor by the developper](https://docs.google.com/document/d/13_Bc-l2FKMgwPx4dZb0sv7eMfYMHhRVgBRShha8kgbU/edit).

- [x] Allow other judges to use the tutor
- [x] Use pyodide to run backend https://github.com/dodona-edu/pyodide-trace-library
- [x] cleanup backend code https://github.com/dodona-edu/json-tracer
[...]

PR #5336, opened 2024-02-02
Use papyros to debug testcases

jorg-vr:
This pull request introduces the new papyros debugger as replacement for the old tutor based debugger.

This contains a new version of papyros (https://github.com/dodona-edu/papyros/pull/615), which adds:
- support for test code lines added at the end of the file
- support for files provided by a js function

This allows us to fully setup a testcase by prefilling the input, adding extra testcode lines and providing all files used by the testcase.

image

The testcode is clearly separated from the student code using a background color and an explanatory comment.
image

There are two icons that allow the student to remove this code or edit it. When the edit option is chosen it just becomes part of the student code:
image

To be able to use the debugger, papyros is now also available on the submission detail and feedback show pages.

Adding this allows removing our dependency on the old python tutor frontend code, the pyodide-trace-library and the fscreen dependency.

Related to: #5014

Note: if approved, I will release a related non-beta papyros version to use


app/assets/javascripts/coding_scratchpad.ts
```diff
@@ -21,103 +21,80 @@ const CODE_TRACE_PARENT_ID = "scratchpad-trace-wrapper";
...
-                        },
-                        () => {
-                            setCode(editor, papyros.getCode());
-                            closeButton.click();
-                            // Open submit panel if possible
-                            document.getElementById(SUBMIT_TAB_ID)?.click();
-                        }
-                    );
+let papyros: Papyros | undefined = undefined;
+let editor: EditorView | undefined = undefined;
```
bmesuere, line 25: I don't think the `= undefined` is needed here.

app/assets/javascripts/coding_scratchpad.ts
```diff
@@ -21,103 +21,80 @@ const CODE_TRACE_PARENT_ID = "scratchpad-trace-wrapper";
...
+export async function initPapyros(programmingLanguage: ProgrammingLanguage): Promise<Papyros> {
+    if (!papyros) { // Only create Papyros once per session, but only when required
+        // Papyros registers a service worker on a specific path
+        // We used to do this on a different path
+        // So we need to unregister old serviceworkers manually as these won't get overwritten
+        navigator.serviceWorker.getRegistrations().then(function (registrations) {
+            for (const registration of registrations) {
+                if (registration.scope !== document.location.origin + "/") {
+                    registration.unregister();
```
bmesuere, line 34: I think this might cause a bug because the unregister is async and the register below doesn't wait for the unregisters (but the chance of that happening is quite low).

```
await Promise.all(navigator.serviceWorker.getRegistrations().then(registrations =>
    registrations.filter(registration => registration.scope !== document.location.origin + "/")
                 .map(registration => registration.unregister())
));
```

app/assets/javascripts/coding_scratchpad.ts
```diff
@@ -142,6 +119,32 @@ function initCodingScratchpad(programmingLanguage: ProgrammingLanguage): void {
...
+        // until after the page is loaded
+        const showButton = document.getElementById(SHOW_OFFCANVAS_BUTTON_ID);
+        showButton.classList.add("offcanvas-show-btn");
+        showButton.classList.remove("hidden");
+        showButton.addEventListener("click", () => {
+            initPapyros(programmingLanguage);
+            editor ||= window.dodona.editor;
+            if (editor) { // Start with code from the editor, if there is any
+                const editorCode = editor.state.doc.toString();
+                const currentCode = papyros.getCode();
```
bmesuere, line 137: can't papyros be undefined?

app/assets/javascripts/tutor.ts
```diff
@@ -55,105 +46,24 @@ export function initTutor(submissionCode: string): void {
...
+        new Offcanvas(document.getElementById(OFFCANVAS_ID)).show();
+        await papyros.codeRunner.provideFiles(inlineFiles, hrefFilesFull);
 
-        modal.addEventListener("hidden.bs.modal", () => {
-            if (fscreen.fullscreenElement) {
-                document.querySelector("#tutor .modal-dialog").classList.remove("modal-fullscreen");
-                fscreen.exitFullscreen();
-            }
-        });
+        document.getElementById(DEBUG_BUTTON_ID).click();
```
bmesuere, line 66: Can't we trigger the debug from code instead of fake clicking the button?

bmesuere (changes requested):
This is amazing work, @jorg-vr! This gets rid of a lot of legacy code, improves the current experience, allows us to add an even better integration in the future and is incredibly fast!

I encountered a small issue when running a simple input/output exercise (sum of 2 numbers): the input of the test was correctly loaded into the input area, but the lines were marked as checked from the start. 

I don't know if possible, but can you add a little more whitespace between the real code and the test code?

Some other suggestions, probably for a later PR:
- If you click debug, wait for the traces to generate and then edit the code you are still debugging the old code. In addition, there is no good way to make the code arrow go away. I would enter a "debug mode" when clicking debug which restricts some features to make this more clear (show stop button, read only editor).
- Remove the "run your code" button and add a new, more prominent button as discussed with the previous PR
- When debugging a test from the feedback table, it would make sense to also show that test in the modal. This might help figuring out what the expected value/output was. 
- The "code arrow" doesn't scroll into view when it's not visible. Maybe we can reuse the active line feature of the editor?

app/assets/javascripts/coding_scratchpad.ts
jorg-vr: The registration happens on scope "/", which equals to `document.location.origin + "/"`
This code will thus never unregister the newly registered serviceworker.

It is here because we used to register the service worker on a different, more specific path. Which caused the old serviceworkers to get priority.

I was actually thinking about removing this code in this pr, as for 99% of cases it will no longer be relevant. But I decided to wait another semester, to avoid issues

app/assets/javascripts/coding_scratchpad.ts
```diff
@@ -12,7 +12,7 @@ const CODE_EDITOR_PARENT_ID = "scratchpad-editor-wrapper";
 const PANEL_PARENT_ID = "scratchpad-panel-wrapper";
 const CODE_OUTPUT_PARENT_ID = "scratchpad-output-wrapper";
 const CODE_INPUT_PARENT_ID = "scratchpad-input-wrapper";
-const OFFCANVAS_ID = "scratchpad-offcanvas";
+export const OFFCANVAS_ID = "scratchpad-offcanvas";
```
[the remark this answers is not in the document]
jorg-vr: I have a personal preference for this style. It is sometimes useful to know whether something is exported when looking at the definition. I have never had a use case where I needed a quick overview of all exported definitions from a file.

I have almost always used this style over the past years, except when explicitly modifying older files.
This has caused for a divergence in styles across files. I am open to fixing this in another pr.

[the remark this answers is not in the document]
bmesuere: I liked the idea of having a clear overview in the "old" style, but have no strong opinion about this. I just noticed a mix of two styles. According to "The Internet" multiple exports from the same file is a code smell. We can discuss this in person over lunch or coffee ;)

bmesuere (approved):

chvp (approved):

PR #5383, opened 2024-02-22
Add mermaid support

### LINKED ISSUE #4168: Add Mermaid 🧜‍♀️ to Markdown parser
When writing a description for an exercise, it would be great if we could include flowcharts generated by a Markdown'ish syntax.

Currently, I'm working with Visio and exporting it to images. By including the flowcharts in the markdown I can keep all information about an excersies centraly at one location in the repo.

Mermaid is a Javascript library lets you create diagrams and visualizations using text and code[^1].

[^1]: [Mermaid](https://mermaid-js.github.io/mermaid/#/)

image

image

jorg-vr:
This pull request adds support for mermaid in dodona.

- [ ] check package size  increase
- [ ] only load mermaid on relevant pages?
- [ ] avoid gitlab links being inserted

- [ ] Tests were added

Closes #4168


niknetniko (changes requested):
Seems to work. One niche issue:

- When changing dark/light mode, the mermaid.js stuff is not update to dark/light mode: it needs a hard refresh.

Would it be an idea to include a basic test for the markdown stuff so it doesn't break unnoticed? Something like:

<details>

<summary>kramdown_test.rb</summary>

```ruby
require 'test_helper'

class KramdownTest < Minitest::Test
  include ApplicationHelper

  def test_mermaid_works
    markdown = <<~EOS
      This is an example of codeblocks using mermaid:

      ```mermaid
      graph TD;
        MoreUsers-->MorePipeline;
        MorePipeline-->MoreRevenue;
        MoreRevenue-->MoreFeatures;
        MoreFeatures-->MoreUsers;
      ```
    EOS

    expected_html = <<~EOS
      <p>This is an example of codeblocks using mermaid:</p>

      <div class="mermaid">graph TD;
        MoreUsers-->MorePipeline;
        MorePipeline-->MoreRevenue;
        MoreRevenue-->MoreFeatures;
        MoreFeatures-->MoreUsers;
      </div>
    EOS

    actual_html = markdown_unsafe(markdown)

    assert_equal expected_html, actual_html
  end
end
```

</details>

bmesuere:
This has a way bigger impact than I expected. We should discuss if we really want to add this. Unfortunately, there doesn't seem to be a server-side option.

jorg-vr:
@niknetniko 
I noticed this as well, but there is no mermaid support to change the active theme. https://github.com/mermaid-js/mermaid/issues/1945
One of the comments does describe a workaround, but I do not think it relevant for such a niche issue. We are probably the only ones who regularly switch themes, this graphs will only be present on a small subset of pages, and it is just a minor visual issue.

I will add the test, thanks :)

@bmesuere I'll drop by to discuss

PR #5386, opened 2024-02-26
Add tips that nudge users to better use of Dodona

### LINKED ISSUE #5335: Nudge users in the right direction with info messages
To help users get the most out of Dodona, we can guide them away from common pitfalls if we identify suboptimal practices.

---

We have seen users create courses with multiple series that only contain 1 or 2 exercises. They sometimes end up with 50+ series. On the series edit page we could add a blue or green callout if a series contains fewer than 3 exercises.

> [!TIP]
> Dodona works best if series contain between 3 and 10 exercises. Consider adding more exercises below.

---

We have also seen series with 40+ exercises in the wild. This is really bad for the page load performance and features like our graphs. On the series edit page we could add a blue or green callout if a series contains more than 10 exercises.

> [!TIP]
> Dodona works best if series contain between 3 and 10 exercises. Adding more can affect loading times.

---

We want to steer users towards TESTed instead of a legacy judge for a number of programming languages. On draft exercises using such legacy judge, we can display an info message to inform them about tested and link to the docs.

> [!TIP]
> You are using the legacy xxxxx judge. For new exercises, we recommend switching to the TESTed for a better experience. See our documentation for how to make the switch.

---

Some users struggle with the syntax of the tests and config file. If we encounter an internal error on a draft exercise, we can point them to the documentation (and maybe offer to contact us).

> [!TIP]
> Facing an internal error? This usually indicates a problem with your exercise or test configuration. Check our documentation for guidance or reach out to us for assistance.

---

Feel free to add other suggestions.

jorg-vr:
This pull request introduces user tips in Dodona. These tips are meant to nudge users in the right direction, so they can make use of Dodona in the most optimal way.

image
For the above notice I had to introduce a database column that tracks if a judge is deprecated
image

image
image
These notices are dynamically added or removed when changing the number of exercises.

image

Closes #5335

### DIFF
--- a/config/locales/views/activities/en.yml
+++ b/config/locales/views/activities/en.yml
@@ -110,6 +110,9 @@ en:
         valid_config: "The exercise must have a valid configuration file."
         is_valid: "The exercise must have a name and a description."
         correct_submission: "You must submit at least one correct solution."
+      tips:
+        title: Tip
+        deprecated_judge_html: "You are using the legacy %{judge} judge. For new exercises, we recommend switching to the TESTed for a better experience. See <a href='https://docs.dodona.be/nl/guides/exercises/testsuites/' target='_blank'>our documentation</a> for how to make the switch."
     series_activities_add_table:
       course_added_to_usable: "Adding this exercise will allow this course to use all of the private exercises in this exercise's repository. Are you sure?"
     edit:
--- a/config/locales/views/activities/en.yml
+++ b/config/locales/views/activities/en.yml
@@ -112,7 +112,7 @@ en:
         correct_submission: "You must submit at least one correct solution."
       tips:
         title: Tip
-        deprecated_judge_html: "You are using the legacy %{judge} judge. For new exercises, we recommend switching to the TESTed for a better experience. See <a href='https://docs.dodona.be/nl/guides/exercises/testsuites/' target='_blank'>our documentation</a> for how to make the switch."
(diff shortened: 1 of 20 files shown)


chvp (approved):

config/locales/views/activities/en.yml
```diff
@@ -110,6 +110,9 @@ en:
         valid_config: "The exercise must have a valid configuration file."
         is_valid: "The exercise must have a name and a description."
         correct_submission: "You must submit at least one correct solution."
+      tips:
+        title: Tip
+        deprecated_judge_html: "You are using the legacy %{judge} judge. For new exercises, we recommend switching to the TESTed for a better experience. See <a href='https://docs.dodona.be/nl/guides/exercises/testsuites/' target='_blank'>our documentation</a> for how to make the switch."
```
bmesuere, line 115:
```suggestion
        deprecated_judge_html: "You are using the legacy %{judge} judge. For new exercises, we recommend switching to the TESTed judge for a better experience. See <a href='https://docs.dodona.be/nl/guides/exercises/testsuites/' target='_blank'>our documentation</a> for how to make the switch."
```

bmesuere (changes requested):
I do think it's important to update the series tip dynamically. We want to inform people when they are adding a lot of exercises and make them think twice at that moment. If we only show it on page load, there a big chance they won't even see the message and if they do, it might already be too late.

jorg-vr:
I have updated them to be dynamic

bmesuere (approved):

PR #5406, opened 2024-02-29
Allow adding scores in submission export

### LINKED ISSUE #2503: Allow export of submissions used by an evaluation
#2462 adds a basic export. However, it might be nice to provide a unified evaluation export that (can) include both the grades and the submissions itself.

jorg-vr:
This pull request adds the scores CSV to the submission export when the relevant option is checked

image

I chose the simplest implementation. This means the scores wont respond to filtering in users and activities etc...
We might implement these more advanced features should there ever be any demand.

Students now also get access to the scores export, as they had access to the submission export. For this reason I limited their CSVs to a single row with their own scores.
I see this as added value as this improves the options for the student to extract their information from dodona.

- [x] Tests were added

Closes #2503

### DIFF
--- a/app/controllers/exports_controller.rb
+++ b/app/controllers/exports_controller.rb
@@ -15,6 +15,7 @@ class ExportsController < ApplicationController
     @series = Series.find(params[:id])
     authorize @series, :export?
     authorize @user, :export? if @user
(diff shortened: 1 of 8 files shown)


jorg-vr:
see related closed issue: https://github.com/dodona-edu/dodona/issues/2503

PR #5411, opened 2024-03-04
Add export of course users

### LINKED ISSUE #3617: Improve discoverability of student list export
Course admins are currently able to download a CSV with user information (dodona id, user email, user labels) by using the `Edit all labels` wizard:
image

This CSV can be useful in some cases (e.g. programmatically adding users to an evaluation). So I think it may be a good idea to add a separate action to the context menu of the members table. For example:
image

This export can be identical to the one downloaded with the `Edit all labels`-wizard.

jorg-vr:
This pull request adds an export of all users in a course. 
image

The csv is a different content type for the index api endpoint.
This means that all filters and sort features of the index page can also be applied.

The export now also includes the student progress in the course.
image

This export also replaces the old specific export labels csv, in the edit all labels flow.

Closes #3617

### DIFF
--- a/app/controllers/course_members_controller.rb
+++ b/app/controllers/course_members_controller.rb
@@ -33,11 +33,22 @@ class CourseMembersController < ApplicationController
     @course_memberships = apply_scopes(@course.course_memberships.order_by_status_in_course_and_name('ASC'))
                           .includes(:course_labels, user: [:institution])
                           .where(status: statuses)
-                          .paginate(page: parse_pagination_param(params[:page]))
 
     @title = I18n.t('courses.index.users')
     @crumbs = [[@course.name, course_path(@course)], [I18n.t('courses.index.users'), '#']]
     @course_labels = CourseLabel.where(course: @course)
+
+    respond_to do |format|
+      format.html do
+        @course_memberships = @course_memberships.paginate(page: parse_pagination_param(params[:page]))
+      end
+      format.js do
+        @course_memberships = @course_memberships.paginate(page: parse_pagination_param(params[:page]))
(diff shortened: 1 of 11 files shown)


niknetniko (approved):

bmesuere (changes requested):
I agree with the export-related code changes and the fact that we make this more prominent, but I'm not convinced that the overflow menu of the search bar is the right place to put those options. They are still hidden and I would never check that menu.

Can we add figure out another way to place these actions? Maybe right-aligned buttons (but that might not be possible on smaller screens)?

app/controllers/course_members_controller.rb
```diff
@@ -33,11 +33,22 @@ def index
...
     @course_labels = CourseLabel.where(course: @course)
+
+    respond_to do |format|
+      format.html do
+        @course_memberships = @course_memberships.paginate(page: parse_pagination_param(params[:page]))
+      end
+      format.js do
+        @course_memberships = @course_memberships.paginate(page: parse_pagination_param(params[:page]))
+      end
+      format.csv do
```
bmesuere, line 48: The x_sendfile setting isn't present here (and was in the old code). Is this something that isn't needed anymore?

bmesuere (approved):

app/controllers/course_members_controller.rb
jorg-vr: "X-Sendfile is a special header option that tells the web server to ignore the content of the response and replace it by the file that is specified in the X-Sendfile header."

This is not relevant in this case, as the response is now the actual csv

Issue #3457, opened 2022-03-03
Tracking: export feature update


bmesuere:
```[tasklist]
### To triage
- [x] https://github.com/dodona-edu/dodona/issues/2919
- [ ] https://github.com/dodona-edu/dodona/issues/2503
- [x] https://github.com/dodona-edu/dodona/issues/4334
- [ ] https://github.com/dodona-edu/dodona/issues/4304
- [ ] https://github.com/dodona-edu/dodona/issues/4244
- [ ] https://github.com/dodona-edu/dodona/issues/2561
- [ ] https://github.com/dodona-edu/dodona/issues/3617
- [ ] https://github.com/dodona-edu/dodona/issues/5338
- [x] auto start download when ready
```

jorg-vr:
@bmesuere this has medium priority, but all medium priority subtasks are done.
Should I lower the priority of this issue?

jorg-vr:
finished in https://github.com/dodona-edu/dodona/pull/5534

PR #5418, opened 2024-03-05
Automatically trigger download of exported files

jorg-vr:
This pull request simplifies the download process of submission exports.

When a users clicks download after submitting the form, a loading bar will be shown while the backend is preparing the zip. Once this is done, the download is automatically triggered.

A retry button is shown, to use should the automatic download fail. ALso a go back button is added to easily return to the page where you came from.

Peek 2024-03-11 13-56

This pr removes the old page with the index of downloads. This is not seen as a relevant feature for dodona. A user should keep track of their downloads themselves.

I also removed the notification for when a download is ready.
I did some benchmarks and the expected zip speed > 100 submissions per second.
This means that for every normal usecase it should only take a few seconds and never more then a minute.
Downloading all 381 417submissions from [Peters programming course](https://dodona.be/en/courses/2802/) does take an hour. If we want to keep async support for these cases, I can revert this.

- [x] Documentation update can be found at dodona-edu/dodona-edu.github.io#403

Part of #3457


config/locales/js/nl.yml
```diff
@@ -327,3 +327,7 @@ nl:
     feedbacks:
       submission:
         submit: Deze oplossing indienen
+    download_button:
+      download: Downloaden
+      done: Gedownload, ga terug
+      downloading: Bezig met het voorbereiden van de download, dit kan enkele minuten duren. Sluit dit deze pagina niet.
```
niknetniko, line 333:
```suggestion
      downloading: Bezig met het voorbereiden van de download, dit kan enkele minuten duren. Sluit deze pagina niet.
```

niknetniko (changes requested):
Do we also want to handle the case where the download doesn't happen or isn't successful? For example, if you have your browser set to ask where a download should be saved and you press cancel, the button still says downloaded, and there is no way I see to download again (except going back through everything).

jorg-vr:
Hmm, maybe the text should not say 'downloaded', but ready to download

We could add a separate 'retry' and 'go back' button

Maybe:
"Export is ready. The download should start automatically. <btn-outline>Retry</btn-outline> <btn-filled>Go Back</btn-filled>"

jorg-vr:
@niknetniko 
image

niknetniko (approved):

bmesuere (changes requested):
Functionality-wise, it would make sense to make the waiting/downloading step 3 in the stepper. This, also because:
- the text above the buttons in the action area of the card is a bit strange
- making "go back" the primary action is strange
- next to the automatic download, you also need an explicit button to trigger the download. Some browsers ask for permission to download first, or block such automatic downloads. In such cases, the download would fail.

jorg-vr:
> Functionality-wise, it would make sense to make the waiting/downloading step 3 in the stepper.

Great idea. Indeed a big improvement

> making "go back" the primary action is strange

Is it? I think this is the action everyone would preform after download?

> next to the automatic download, you also need an explicit button to trigger the download.

This is what the "retry download" button is for. Is this not clear enough? or do you expect something totally different?

bmesuere (approved):

app/assets/stylesheets/components/btn.css.scss
```diff
@@ -134,7 +134,8 @@
     border-color: var(--d-btn-color);
   }
 
-  &:disabled {
+  &.disabled,
+  &:disabled{
```
chvp, line 138:
```suggestion
  &:disabled {
```

PR #5421, opened 2024-03-07
Improve discoverability of evaluations and scoresheets

### LINKED ISSUE #5408: Make evaluations more discoverable
The evaluations feature should be more discoverable.

After discussion the suggested solution is adding a footer row to series card with action buttons.

Aside from evaluations we can also move the link to the seriesscoresheet here.

Depending on the implementation of #4340 we might also include a link to live series stats here.

jorg-vr:
This pull request improves the discoverability of the evaluations feature.

Important series actions, such as evaluations and the scoresheet are now displayed in the new footer.

image

For students only feedback is considered important. A button is inserted in the header for them. If no feedback is present, no button is shown.
image

Other series actions are still hidden behind the dropdown. For teachers this dropdown is move to the footer.
image
image

- [x] Documentation update can be found at dodona-edu/dodona-edu.github.io#420

Closes #5408


chvp (approved):

bmesuere (commented):
Some things to discuss:
- Does it make sense to add buttons to the bottom, but keep the overflow button at the top? I would expect the overflow to be in the same area. I do acknowledge that for students, without bottom bar, the top right seems the most logical.
- I feel like we should prioritize the buttons depending on the state. If a series has no evaluation, the scoresheet button (and/or the future "live view" button) should be more prominent than the button to start an evaluation which will be hardly used (we might even move it to the overflow button in case the series has a deadline which hasn't passed yet?). In case the series has an evaluation, the opposite is possibly true.

jorg-vr:
> Does it make sense to add buttons to the bottom, but keep the overflow button at the top? I would expect the overflow to be in the same area. I do acknowledge that for students, without bottom bar, the top right seems the most logical.

I tried out moving it to the bottom, but I was hesitant about what we would do for students, as they can have a combination of series both with and without bottom bar.

We can also just always leave it on top for students, and always in the bottom row for teachers.

But I don't have a strong opinion.

> I feel like we should prioritize the buttons depending on the state. If a series has no evaluation, the scoresheet button (and/or the future "live view" button) should be more prominent than the button to start an evaluation which will be hardly used (we might even move it to the overflow button in case the series has a deadline which hasn't passed yet?). In case the series has an evaluation, the opposite is possibly true.

I experimented with the other button styles, but I don't think they look visually appealing:
image
image

It will start to look very crowded if every series on a page has a primary action button.
From a UX perspective, I would also never consider either of these actions to be the primary action a teachers wants to perform on this series. In every case they remain optional links.

We could hide the evaluation button when the deadline hasn't passed yet. But I am wondering why we would do that?
It might just confuse the user why the button is present on some but not all series.
If we want to make evaluating impossible before the deadline passed, we should disable the button and add a tooltip explanation.

bmesuere (changes requested):
For students:
- no bottom bar
- "show evaluation"-button at the same position as graph toggle buttons

For teachers:
- no border buttons
- if series has evaluation, make it primary
- move 3 dots to bottom bar

bmesuere (approved):

Issue #5410, opened 2024-03-01
Remove the hassle of emailing secret links to lists of students at the start time of the evaluation


jorg-vr:

BTWS2:
Relevant closed issue https://github.com/dodona-edu/dodona/issues/2675 : Show/Hide series based on time.

Issue #5409, opened 2024-03-01
Block students from submitting after they leave the room


jorg-vr:

BTWS2:
Great feature request! For the time being I have to check this manually after each evaluation. 

There should be a default (e.g. last one before the deadline). Some students will forget to mark a submission as final at the end of the evaluation.

Issue #2675, opened 2021-04-30
Show/Hide series based on time


BTWS2:
Series would (dis)appear based on a predefined time entered by the teacher.

Advantages:
* Series of a test/exam would (dis)appear at a predefined time, no more action required by the teacher.
* New series could appear at the same time the lesson begins or at the beginning of the week.
* Teachers can plan ahead and don't have to remember to make it available at the right time.

![image](https://user-images.githubusercontent.com/56451049/116728001-9eb1f600-a9e5-11eb-93be-fcb98abc9154.png)

chvp:
This is too complex for a minimal usecase and therefore out-of-scope. If you really want to do this, you can use the API and trigger the changes yourself programmatically.

BTWS2:
I hadn't thought of that. Do you have any comments about this code? @chvp 

````python
import asyncio
import http.client
import json
from datetime import datetime

# Change the visibility of a series with the API
# Docs: https://thepieterdc.github.io/dodona-api-java/

TOKEN = "YOUR_TOKEN"  # UPDATE
SERIES_NUMBER = 123456  # UPDATE
START = datetime(year=2021, month=6, day=7, hour=15, minute=55, second=0)  # UPDATE
END = datetime(year=2021, month=6, day=7, hour=15, minute=56, second=0)  # UPDATE

conn = http.client.HTTPSConnection("dodona.ugent.be")
headers = {
    "Content-type": "application/json",
    "Accept": "application/json",
    "Authorization": TOKEN
}


async def change_series_visibility(series_number, status):
    """status -- open, hidden or closed"""
    info = {'visibility': status}
    json_info = json.dumps(info)
    conn.request("PUT",
                 f"/nl/series/{series_number}/",
                 headers=headers,
                 body=json_info)
    res = conn.getresponse()
    data = res.read()
    parsed = json.loads(data)
    print(json.dumps(parsed, indent=4))
    conn.close()


async def wait_until(dt):
    # sleep until the specified datetime
    now = datetime.now()
    await asyncio.sleep((dt - now).total_seconds())


async def run_at(dt, coro):
    await wait_until(dt)
    return await coro


loop = asyncio.get_event_loop()
task_start = loop.create_task(run_at(START, change_series_visibility(SERIES_NUMBER, "open")))
task_end = loop.create_task(run_at(END, change_series_visibility(SERIES_NUMBER, "closed")))
loop.run_until_complete(asyncio.gather(task_start, task_end))
````

[...]

PR #5422, opened 2024-03-07
Add visibility options to show series between start and end time

jorg-vr:
This pull request

- [ ] Tests were added

Closes # .


bmesuere (changes requested):
This is a good idea, but I would change the UI a bit:
- [ ] swap the position of the description and visibility
- [ ] make the visibility select radio buttons. This will allow adding a bit more info for the options and have space for the to and from field and hidden link.
- [ ] depending on how it looks, make the visibility a separate "section" like we do for courses.

- [ ] The info messages inside the series cards now depend on the presence of start and/or end date. I would instead make them depend on the current time and current visibility.

pdawyndt:
This solution still requires the need for making separate series when we have evaluations or exams with students that have a special status which gives them extra time. I would have hoped that the option to creation evaluations beforehand would have allowed to set a scheduled visibility per student (or default that can be overruled for individual students), with the option to also change visibility manually when students leave the exam room early.

bmesuere:
> This solution still requires the need for making separate series when we have evaluations or exams with students that have a special status which gives them extra time. I would have hoped that the option to creation evaluations beforehand would have allowed to set a scheduled visibility per student (or default that can be overruled for individual students), with the option to also change visibility manually when students leave the exam room early.

This is out of scope for this PR.

pdawyndt:
Agree that this is for a future PR, but having two series is a hassle. I would definitly prefer the option to have a deadline per student: of course we need the option to set a common deadline for all students at once and then overrule that for a selection of students by hand. But these deadlines have dual use: they determine when students can get access (see + submit) to the exercises in a series, but they are also used to pick "the last submission before the deadline" when evaluating the students. The latter is something we now have to do manually anyway. The advantage could be that we could do this once (and beforehand), as we also made mistakes against it in the future since this is a very implicit process.

Now we only make a separate series for students that have a deviating deadline: so it only pupose is showing the custom deadline to the students that have extra time. After the evaluation we delete the series again,  create a single evaluation for all students in the same series, automatically have Dodona pick the last submission for the deadline that holds for most students and then manually change the selected submission for students with an alternative deadline. A solution where we would need to attach evaluations to two separate series would mean that we never get rid anymore of the duplicated series, and that we have a hassle working in two different evaluations.

config/locales/views/series/nl.yml
```diff
@@ -11,6 +10,17 @@ nl:
...
+        title: Zichtbaarheid
+        subtitle: "Voor wie is deze reeks zichtbaar?"
+        open: "Zichtbaar voor iedereen"
+        open_help: "Deze reeks wordt altijd getoond op de cursuspagina."
+        closed: "Verborgen voor studenten"
+        closed_help: "Deze reeks is niet toegankelijk voor studenten. Ze is enkel zichtbaar voor lesgevers."
+        hidden: "Enkel zichtbaar via geheime link"
+        hidden_help: "Deze reeks is enkel zichtbaar voor lesgevers. Studenten kunnen toegang krijgen via de geheime link."
+        timed: "Zichtbaar voor studenten tussen start en eindtijd"
+        timed_help: "Deze reeks wordt zichtbaar voor studenten tussen de gegeven start en eindtijd. Het is ook mogelijk om enkel de starttijd of eindtijd in te stellen."
```
niknetniko, line 23:
```suggestion
        timed_help: "Deze reeks wordt zichtbaar voor studenten tussen de gegeven start- en eindtijd. Het is ook mogelijk om enkel de starttijd of eindtijd in te stellen."
```

config/initializers/field_with_errors.rb
```diff
@@ -0,0 +1,9 @@
+ActionView::Base.field_error_proc = Proc.new do |html_tag, instance|
+  class_attr_index = html_tag.index 'class='
+
+  if class_attr_index
+    html_tag.insert class_attr_index+7, 'is-invalid '
```
niknetniko, line 5: Should this contain a comment that explains where the +7 comes from? (It wasn't immediately obvious for me)

niknetniko (commented):
I'm reasonably sure this is out of scope for this PR, but this "timed" status seems to mean open in the given period, or closed otherwise. However, there is no way to have the status be "hidden" in the given period, right?

Pure code-wise, I was wondering if it would be cleaner to not have a new status, but just attach the start and/of end date to the existing status "open"? The code already seems to support missing start and end date, so a status of "timed" without start or end date is equivalent to "open". This would potentially supporting "hidden" with a time periode easier.

However, I don't have a strong opinion either way, so it's fine as-is.

jorg-vr:
@niknetniko 
It is true that `timed`, with both start and end time nil equals `open`

We could thus indeed completely remove it as extra 'option'.
This could even result in a UX closer to the one suggested in  #2675

Making sure that it also works well with 'hidden' will probably require some extra implementation work. Using the times together with 'closed' just seems odd. But anyway, I don't think their is a relevant usecase for either of these options?

I feel like the current UX work is rather clear. And although this is not required, I like that how it works behind the scenes matches with what the user sees.

For the full implementation of #5410, the chosen route does mean two more visibility options: `for_users` and `for_users_timed`. Which might not be ideal, and would be simplified with the checkbox approach.

I also don't have a strong opinion on this, and was postponing fully resolving this issue to the next PR.

bmesuere (commented):
Before merging this, we should discuss the suggested feature in person. I've been thinking about this for the past few days and we might create false expectations. I'm ok with the "open" part, but not sure about the "close" part. Depending on implementation details and situation, students might still have access to exercises after the close date.

jorg-vr:
Results of discussion:
This pr will be limited to a visibility start time.

We do want to support automatic closing in the future, but it might be related to how we implement #5409 and before this change it would be useful to reevaluate linking submissions directly to a series. This way we can make sure that a closed series always means that submitting is impossible.

niknetniko (approved):

config/locales/views/series/nl.yml
```diff
@@ -11,6 +10,17 @@ nl:
...
+        title: Zichtbaarheid
+        subtitle: "Voor wie is deze reeks zichtbaar?"
+        open: "Zichtbaar voor iedereen"
+        open_help: "Deze reeks wordt altijd getoond op de cursuspagina."
+        closed: "Verborgen voor studenten"
+        closed_help: "Deze reeks is niet toegankelijk voor studenten. Ze is enkel zichtbaar voor lesgevers."
+        hidden: "Enkel zichtbaar via geheime link"
+        hidden_help: "Deze reeks is enkel zichtbaar voor lesgevers. Studenten kunnen toegang krijgen via de geheime link."
+        timed: "Zichtbaar voor studenten na starttijd"
+        timed_help: "Deze reeks is niet zichtbaar voor studenten. Ze zal zichtbaar worden op de gegeven starttijd."
```
bmesuere, line 23:
```suggestion
        timed_help: "Deze reeks is enkel zichtbaar voor studenten na de opgegeven starttijd."
```

config/initializers/field_with_errors.rb
```diff
@@ -0,0 +1,12 @@
+ActionView::Base.field_error_proc = Proc.new do |html_tag, instance|
```
bmesuere, line 1: I don't know if I really like this. It feels like a hack (e.g. I think it would break if spaces were used after the class attribute).

What exactly is the problem you're trying to solve?

bmesuere (changes requested):
Would it make sense to set the current date as default for the start time?

config/initializers/field_with_errors.rb
jorg-vr: In the default behavior, rails adds an extra div with an error class around the input which has caused errors
image

This does not work well with our more complex inputs, such as this one, which was inlined and part of a bootstrap input group with buttons

I have encountered this issue before and those times I was able to resolve it by adding extra css style rules for the specific case.

This would be a lot harder to do in this specific case. And thinking about it, adding the extra div just isn't a good solution.

Changing the style of the input element should be done by adding a class to it.

> I think it would break if spaces were used after the class attribute

From my understanding, a class definition would be either `class="` or `class='`.
Are you saying `class ="` or `class= "` would also be valid?
Because in that case I would update the code to use a regex, or have a look at our linter to make sure this never happens.

jorg-vr:
> Would it make sense to set the current date as default for the start time?

In my opinion it would not. If you would want to use the current date, there would be no reason to select this visibility option.
Leaving it empty, with a validation is in my opinion the best way to make sure the user fills out the date relevant for them

bmesuere (approved):
👍

PR #5430, opened 2024-03-12
Make search actions and options more discoverable

jorg-vr:
This pull request moves the search actions to actual buttons with text.
This makes them more discoverable.

To achieve this I split up 'options' and 'actions', which also improves the code.
Options have also been simplified, as they were always used to set a single variable to true or false.

I also updated course tabs to make use of the existing search_tab component. This allowed me to get rid of some custom js in course.ts that interacted with actions which should only be shown on a specific tab.

This pr also contains a minor bugfix for the loading bar, which I noticed while working on this.

To make the actions sufficiently different from the existing dropdowns, the dropdowns have been redesigned:
image

Actions for course members
image
image

Actions for solutions
image

- [x] Tests were added

Fixes https://github.com/dodona-edu/dodona/pull/5411#pullrequestreview-1924977675


jorg-vr:
Maybe an idea for the scoresheet to make the series option a dropdown filter and the download button a search action.

This would win some vertical space

bmesuere (changes requested):
This is a step in the right direction, but I'm not entirely convinced yet.
- If we group the filter settings, we need a better word for "options"
- To me, auto refresh is not a filter so it doesn't belong there
- The actions now look very similar to filters. The only difference is the left vs. right alignment. I feel like this is not enough.

Sorry for not coming up with alternatives myself right now.

jorg-vr:
Ok I've tried some different styles for the action buttons

<details>
<summary>Outline</summary>

image
image
image
image
image
image



</details>

<details>
<summary>Tonal</summary>

image
image
image
image
image
image


</details>

<details>
<summary>Filled</summary>

image
image
image
image
image
image


</details>

<details>
<summary>Secondary outline</summary>

image
image
image
image
image
image


</details>

<details>
<summary>No outline for dropdowns</summary>

image
image
image

</details>

Of these options I still prefer the outline buttons.
Tonal buttons look okay in light mode, but way to prominent in dark mode. The opposite can be said for filled buttons.

A case can be made to give the accept all and decline all a success and danger color. But that doesn't fix the general problem of styling the action buttons.

bmesuere (approved):
This looks good!

They can use a bit of vertical padding when they use multiple lines:
image

chvp (approved):

PR #5446, opened 2024-03-20
Refactor ACM-IDM provider to support LeerID

### LINKED ISSUE #4164: LeerID
Op te volgen: https://onderwijs.vlaanderen.be/nl/leerid

> LeerID is de single sign-on oplossing van de Vlaamse overheid, die leerlingen van het leerplichtonderwijs toelaat om via 1 gebruikersnaam en 1 wachtwoord aan te melden op allerhande digitale leermiddelen en toepassingen.

jorg-vr:
This pull request adds support for LeerID.

In order to add this support I refactored the existing OIDC omniauth strategy.

First of all I renamed this strategy to ACM-IDM. 
I did this rename as the strategy included a lot of ACM-IDM specific fixes. And other strategies that need OpenIDConnect, such as surf and elixir, extend from the OpenIDConnect strategy directly instead of from our specific OIDC version.

This rename allowed me to move the client_id and issuer from our database to our credentials.

It also removed the need for all the code that kept track of the provider id. Instead the provider is now based on the institution information like most other strategies. For LeerID I used the `ov_orgcode` for this. If this is not present it defaults to the Flemish government.

- [ ] It would preferable to also rename the callback endpoint to acmidm now that we are in contact with the flemish government.

I also simplified the code in the lib files to more closely match the structure used for Surf and Elixir. This makes them easier to compare and comprehend.

- [ ] Tested on naos
- [ ] Edit provider 904 on naos and in production

Closes #4164


jorg-vr:
I'm not sure about the new name ACM-IDM, as the dash is not posible in variable names, and the abbreviation might not be well known. I am op to other suggestions. 'VlaamsOverheid' nor 'LeerID' seem broad enough as name.

PR #5456, opened 2024-03-28
Add regex to make sure git remote is an ssh url

### LINKED ISSUE #5445: Block use of HTTP urls for new repositories

jorg-vr:
This pull request adds a validation on repository and judge remotes to make sure they aren't http(s) urls.

We do this because we can't interact with https remotes.

I added a more complex regex in [8806ce1](https://github.com/dodona-edu/dodona/pull/5456/commits/8806ce18cc75f2197ca90fb1077513bf1c6ac274) but that caused a lot of issues with our tests because we use local remotes there.

The validation now only focuses on the url starting with https. Other mistakes, such as typo's should be noticed by our existing validation that checks is the host is readable.

image
image

- [x] Tests were added

Closes #5445

### DIFF
--- a/app/models/judge.rb
+++ b/app/models/judge.rb
@@ -22,7 +22,7 @@ class Judge < ApplicationRecord
 
   validates :name, presence: true, uniqueness: { case_sensitive: false }
   validates :image, presence: true
-  validates :remote, presence: true, uniqueness: { case_sensitive: false }, format: { without: /\Ahttp/ }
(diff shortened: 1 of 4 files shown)


jorg-vr:
@niknetniko that regexs works to check git URLS, but it doesn't exclude http(s) urls, which is the whole point of this pr
Or am I missing something?

niknetniko (approved):

jorg-vr:
replaced by #5458

PR #5460, opened 2024-03-29
Fix sticky next button when evaluating

jorg-vr:
Peek 2024-03-29 11-27

### DIFF
--- a/app/assets/javascripts/evaluation.ts
+++ b/app/assets/javascripts/evaluation.ts
@@ -117,24 +117,3 @@ export function initEvaluationStepper(): void {
 
     init();
 }
-
-/**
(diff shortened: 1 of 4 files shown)


chvp (approved):

bmesuere (commented):
Would this be a css-only fix (adding negative margin and padding)?

```
@media (min-width: 992px) {
    .feedback-metadata {
        position: sticky;
        top: calc(var(--d-navbar-height) + 20px) !important;
        min-height: 100px;
        max-height: calc(100vh -(var(--d-navbar-height) + 20px));
        padding-bottom: 80px;
        margin-bottom: -80px;
        overflow-y: auto;
    }
}
```

bmesuere (approved):

PR #5471, opened 2024-04-05
Add dynamic counts to search filters

jorg-vr:
This pull request contains a refactor of search filters. It adds dynamic counts of the expected results to all search filters. When already partially filtered, it also limits the other filters to only show relevant options.

image
image
image

The initial goal was to get rid of some old code that used js functions as part of 'filterCollection' objects to extract the color and paramVal from a filter option. This made most related code quite ugly and there was no longer any need for it.

So with cc610d658950ff2679f3a362dca044e81194b20a and 810fa47aae22c7cc2bd3ece8d745be697e3497c4 I made filterCollection a pure map of data objects.

With these changes it became a lot simpler to dynamically update filters.
I tried to write generic concerns that automatically define a dynamic option count when defining a scope. This way the end result should be less specific code.

In the most basic case writing a filter for a given field of a model is as simple as writing `filterable_by :field` in the model and writing `has_filter :field` in the controller. In the controller, the `@filters` variable can then be set by calling `filters(elements)` on the unfiltered list of elements. `@filters` is then used to update the shown filter options and counts. 
Common cases, such as enums and foreign keys can be filtered using `filterable_by :field, is_enum: true` and `filterable_by :field, model: Model` respectively.

Naturally there are always some more complex cases, in which case a custom name_hash function can be written or even a full custom count function. See `activity.rb` for some examples of this.

The most complex case were `course_labels` as these required the course as extra argument. For now these are generalized in their own specific reusable function.


app/views/layouts/_searchbar.html.erb
```diff
@@ -29,198 +29,135 @@
...
-          color: function () {
-            return "indigo";
-          },
+          color: "indigo",
         };
       <% end %>
 
+        <% if @filters.present? %>
+          filterCollections = <%= raw @filters.map { |v| [v[:param], v] }.to_h.to_json %>;
+        <% end %>
```
chvp, line 159: The indent seems weird here?

chvp (approved):

chvp:
(There are still some test and linting issues here.)

app/controllers/concerns/has_filter.rb
```diff
@@ -0,0 +1,38 @@
+module HasFilter
+  extend ActiveSupport::Concern
+  included do
+    delegate :filter_options, to: :class
+    self.filter_options = []
```
bmesuere, line 5: I'm not sure about this. filter_options is delegated to the class, but then set? In addition, vscode shows a warning here because "set a class-level attribute within an instance context"

app/models/activity.rb
```diff
@@ -73,35 +72,47 @@ class Activity < ApplicationRecord
...
     by_language = all # allow chaining of scopes
     by_language = by_language.where(description_en_present: true) if languages.include? 'en'
     by_language = by_language.where(description_nl_present: true) if languages.include? 'nl'
     by_language
   }
+  define_singleton_method('description_languages_filter_options') do
+    count_by_description_languages = unscoped.where(id: select(:id)).select(
+      'COUNT(CASE WHEN description_nl_present THEN 1 ELSE NULL END) AS nl',
+      'COUNT(CASE WHEN description_en_present THEN 1 ELSE NULL END) AS en'
+    )
```
bmesuere, line 95: I'm not too fond of adding explicit sql to ruby code

app/models/concerns/filterable.rb
```diff
@@ -1,8 +1,83 @@
...
-    before_save :set_search
-    scope :by_filter, ->(filter) { filter.split.map(&:strip).select(&:present?).inject(self) { |query, part| query.where("#{table_name}.search LIKE ?", "%#{part}%") } }
+  class_methods do
+    def search_by(*columns)
+      define_method(:set_search) do
+        self.search = columns.map { |column| send(column) || '' }.join(' ')
+      end
+
+      before_save :set_search
+      scope :by_filter, ->(filter) { filter.split.map(&:strip).select(&:present?).inject(self) { |query, part| query.where("#{table_name}.search LIKE ?", "%#{part}%") } }
```
bmesuere, line 11: Where does `part` come from? Can a user inject something here?

app/models/concerns/filterable.rb
```diff
@@ -1,8 +1,83 @@
 module Filterable
   extend ActiveSupport::Concern
 
-  included do
-    before_save :set_search
-    scope :by_filter, ->(filter) { filter.split.map(&:strip).select(&:present?).inject(self) { |query, part| query.where("#{table_name}.search LIKE ?", "%#{part}%") } }
+  class_methods do
```
bmesuere, line 4: This class is very complex and needs more documentation.

app/models/concerns/filterable.rb
```diff
@@ -1,8 +1,83 @@
...
+        count = unscoped.where(id: select(:id)).joins(associations).group(column).count
+
+        names = name_hash.call(count.keys)
+
+        count.map { |key, value| { id: key.to_s, name: names[key].to_s, count: value } }
+             .filter { |option| option[:name].present? }
+      end
+    end
+
+    def filterable_by_course_labels(through_user: false)
```
bmesuere, line 59: Why is there a specific course label filter in a generic concern?

bmesuere (changes requested):
This PR seems to include a bunch of changes from other PRs as well (dependencies, fix for graph time, ...) Maybe double check why this is and if we won't merge things we don't want.

Did you test the performance of these changes? I don't know if adding the counts needs a lot of additional queries? e.g. for the submissions table this might be quite impactful.

Can you explain the user of the Filters header?

app/models/concerns/filterable.rb
jorg-vr: this is the query string by the user (so user generated text)
But it is give as the second argument for a where query that uses a `?`. When replacing a `?` rails will handle the proper escaping to avoid sql injection

app/models/concerns/filterable.rb
jorg-vr: The course_labels filter differs slightly from all the other filters as it requires an extra argument: `course_id`

Adding the flexibility to support an extra argument to `filterable_by` would make that method even more complex. Course_labels is also the only usecase of this feature.

Though this usecase is reused by three models. So it still made sense to not write the functionality three times.
If you like I could write this in a separate concern.
But I would prefer to keep it here as `filterable_by_course_labels` is just a specific implementation of `filterable_by` with slight variations. If changes should be applied to one method they might also be relevant for the other.

app/controllers/concerns/has_filter.rb
jorg-vr: This variable has to be delegated to the class.
If I don't do that, the list of filters is stored in the concern, which means that every class that includes the concern is using the same list.

I tried a lot of different options and the only other method that worked was storing a dictionary where lists were indexed by controller name.
I found delegating the variable to the class to be a cleaner solution.

jorg-vr:
> This PR seems to include a bunch of changes from other PRs as well (dependencies, fix for graph time, ...) Maybe double check why this is and if we won't merge things we don't want.

These were all changes already on main. A new merge with main has fixed the issue of them being listed here.

> Can you explain the user of the Filters header?

This is only used in a single place. This usecase is saved annotations. The saved annotations list is the only page that doesn't fetch javascript upon search, but instead fetches the json data and feeds that to the web components.
This request now also needs the filter information to update the filters.
I treated the filter metadata the same way as the pagination metadata by including them in the headers of the request.

jorg-vr:
> Did you test the performance of these changes? I don't know if adding the counts needs a lot of additional queries? e.g. for the submissions table this might be quite impactful.

I minimized the amount of queries. But testing it on naos, the queries are unusable slow.
I'll have a look if they can be easily improved

jorg-vr:
We might have to reconsider this whole pr :(
I underestimated how hard it would be to optimize 'COUNT' queries...
There is no simple fix such as adding indexes as it seems innoDB does not use these indexes for counting.

The only suggested solution is keeping track of summary tables. But implementing this in a flexible way for all our search pages would add a whole new layer of complexity...

jorg-vr:
Maybe I overreacted.

The newly added count queries have a similar speed as the existing count query that already ran for pagination.
For most pages it doesn't have a big impact. Most pages only have one or two filters and this results into at most 100ms extra response time on naos, but often a lot less.

I have already removed the counts for the submission pages, as it was unbearably slow there (but these pages where already way to slow on naos anyway). 
For reference the pagination count takes 15.8 seconds out of the total 16.7 seconds on the submissions page. Adding up to two extra count queries to this page tripled response time. 
Side note: removing the page count on this page would greatly speed up the page.

The activities page is an edge case. With 7 filters the increase in query time can be up to 300ms, which is about 20% of the total time required to render the page. This makes interactivity on naos feel slightly suboptimal. 
But our production server is always a bit faster, and the UX improvement is greatest for activity selection, so it might be worth a 20% increase in loading time. 
This page also has room for speed ups in other places, such as removing the need for n+1 queries for popularity and activity status.

app/controllers/concerns/has_filter.rb
chvp: This also seems weird to me. It should be possible to delegate to the controller instance instead of the class, which would be better, IMO.

app/controllers/feedbacks_controller.rb
```diff
@@ -43,9 +44,11 @@ def show
   end
 
   def edit
-    @submissions = apply_scopes(Submission)
+    @submissions = Submission
```
chvp, line 47:
```suggestion
    @submissions = Submission.all
```
This should at least by a activerecord query, IMO. Right now it's just referring to the class, which seems counter-intuitive.

app/models/application_record.rb
```diff
@@ -1,4 +1,5 @@
 class ApplicationRecord < ActiveRecord::Base
+  include Filterable
```
chvp, line 2: Do we want this in `ApplicationRecord`? Not all models will use this, I guess, so I would do it in the relevant models.

app/models/concerns/filterable.rb
chvp: I would move this to a separate concern.

app/models/activity.rb
```diff
@@ -73,35 +72,45 @@ class Activity < ApplicationRecord
...
     by_language = by_language.where(description_nl_present: true) if languages.include? 'nl'
     by_language
   }
+  define_singleton_method('description_languages_filter_options') do
+    scope = unscoped.where(id: select(:id))
+
+    [{ id: 'nl', name: I18n.t('js.nl'), count: scope.where(description_nl_present: true).count },
+     { id: 'en', name: I18n.t('js.en'), count: scope.where(description_en_present: true).count }]
+      .filter { |lang| lang[:count].positive? }
+  end
```
chvp, line 97: Can we add an extra method to the concern that handles cases like this (and popularity below). It also occurs in submission, IIUC.

config/locales/models/en.yml
```diff
@@ -175,6 +175,7 @@ en:
           permission_change: Change in permission level
           exercise_repository: Creation of exercise repository
           error: Error
+          no_auth_id_sign_in: User logged in without auth_id
```
chvp, line 178: Is this relevant to this PR? What does this mean?

test/javascript/components/search/search_field.test.ts
```diff
@@ -3,7 +3,7 @@ import { aTimeout, fixture, nextFrame } from "@open-wc/testing-helpers";
 import userEvent from "@testing-library/user-event";
 import { screen } from "@testing-library/dom";
 import { SearchField, SearchFieldSuggestion } from "components/search/search_field";
-import { Label, FilterCollection } from "components/search/filter_collection_element";
+import { Label, FilterOptions } from "../../../../app/assets/javascripts/components/search/filter_element";
```
chvp, line 6: Is this full path necessary? It didn't seem to be before?

test/javascript/components/search/search_token.test.ts
```diff
@@ -4,7 +4,7 @@ import { fixture, nextFrame } from "@open-wc/testing-helpers";
 import userEvent from "@testing-library/user-event";
 import { screen } from "@testing-library/dom";
 import { html } from "lit/development";
-import { Label } from "components/search/filter_collection_element";
+import { Label } from "../../../../app/assets/javascripts/components/search/filter_element";
```
chvp, line 7: Same here.

chvp (changes requested):

app/models/application_record.rb
jorg-vr: I changed this.

The reason I did put it here was because already 9 different models used it, so that is pretty general. Also the concern does not have any effect upon include. It only makes extra class functions available to the models.

But in the end it doesn't matter much.

app/models/activity.rb
jorg-vr: I have now added this extra method:
```ruby
def filter_options_for(name, &block)
  define_singleton_method("#{name}_filter_options", &block)
end
```

Which does make it a bit more readable. But except for hiding the `define_singleton_method` it doesn't do much, as these functions don't have much in common

config/locales/models/en.yml
jorg-vr: This is an event type which did not have a translation yet

In production this causes an ugly missing translation text:
image

With the current change, the missing translation became a hard error (breaking my page), which made me notice and fix it.

This event is created in the authentication flow as suggested here: https://github.com/dodona-edu/dodona/pull/3663#discussion_r886917099

It has actually never been logged so far. I'll create an issue to clean up the authentication code and remove the event type

chvp (approved):

bmesuere:
https://github.com/coderabbitai review

jorg-vr:
@coderabbitai review

bmesuere:
@coderabbitai review

jorg-vr:
rebase and request review

bmesuere (approved):

PR #5509, opened 2024-04-29
Update major Node and Eslint versions

jorg-vr:
This pull request contains an update of eslint and a required update of node as well.

It replaces #5474 which gets frequently closed by dependabot. See https://github.com/dependabot/dependabot-core/issues/7647

As this is a major version change, this also required some changes on our side:
 - I updated all github runners to node 22 as older versions are no longer supported
 - I moved our .eslintrc and .eslintignore to modern eslint.config.js using this [migration guide](https://eslint.org/docs/latest/use/configure/migration-guide) (eslintrc is no deprecated)
 - I simplified the config by always using the typescript eslint config (Almost everything is typescript now anyway and the typescript specific rules dont seem to break on javascript files)
 - I fixed some minor issues the linter started complaining about
 - yarn lint now also explicitly lints the test directory
 - Added the rule to error aon explicit any
 - Removed the rule that warns on line length
 
 Before release we should
 - [x] update our production servers to node 22 (tests should run in the same context as production)
 - [ ] set "no-unused-vars" rule back to error once this issue is fixed: https://github.com/43081j/eslint-plugin-lit/issues/195
 - [x] Discuss if we want to keep a line length warning (We now have hundreds of violations, which wouldn't be improved by shorthening them in my opninon)
 - [x] Discuss if we want to enforce the 'no-explicit-any' rule. Because everything is typescript now, there is less need to use `any`.
 - [x] Fix my editor to use the new config


niknetniko:
To give 2cents:

> * [ ]  Discuss if we want to keep a line length warning (We now have hundreds of violations, which wouldn't be improved by shorthening them in my opninon)

I am strongly in favour of dropping the warning (I already do this locally).

> * [ ]  Discuss if we want to enforce the 'no-explicit-any' rule. Because everything is typescript now, there is less need to use `any`.

I think this would be good, but it might be a lot of work.

bmesuere (approved):

PR #5510, opened 2024-04-29
Fix dates in tooltips for series graph

jorg-vr:
This pull request fixes the dates in the series graph.
For some reason the step size was substracted from the startdate to find the end-date, which was simply wrong.


chvp (approved):
Thanks!

bmesuere (approved):

PR #5512, opened 2024-05-03
Add button to undo submission selection in evaluations

### LINKED ISSUE #5501: How to undo selecting submission after evaluation deadline
I came across an edge case today. If the teacher accidentally selects a submission after the defined deadline, there is (as far as I know) no way to undo that action.

Summary
1. No submission before deadline.
2. Teacher selects submission after deadline accidentally.
3. No way to go back to "no submission before deadline" state.

E.g. the deadline of the evaluation is 24/04/2024 at 15:30. If I accidentally select a submission after this deadline. How do I undo this? 

image

jorg-vr:
This pull request introduces a 'deselect' button when selecting submissions for a given feedback.
image

I also fixed the confirm messages on these buttons, which used a deprecated style (see https://stackoverflow.com/questions/16668949/how-to-add-confirm-message-with-link-to-ruby-on-rails)

Closes #5501

### DIFF
--- a/app/assets/stylesheets/components/table.css.scss
+++ b/app/assets/stylesheets/components/table.css.scss
@@ -58,8 +58,6 @@
   text-align: right;
 
   &.submissions-table {
-    text-align: center;
-
     // by default the size of the element in the col is used as width if width is smaller than that element.
     // This width is just here to force everything to the right
     width: 1px;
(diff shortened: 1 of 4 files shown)


bmesuere (changes requested):
I would remove the trash icon since we don't remove the submission itself which might confuse the user. 

I also think our buttons are right aligned (centered here) but I might be wrong.

bmesuere (approved):

niknetniko (approved):

PR #5526, opened 2024-05-13
Update publications page

rien:
This pull request
- Adds the new Dolos article on SoftwareX (replacing the conference abstract from 2023)
- Adds the new ECHO preprint (arXiv)
- Removes the Dodona conference abstract (to avoid confusion with the SoftwareX article with the same name)


bmesuere (changes requested):
I would remove all conference abstracts.

bmesuere (approved):

PR #5534, opened 2024-05-16
Add plagiarism detection for evaluations

### LINKED ISSUE #5528: Add a `Detect plagiarism` button in the evaluation view
Teachers don't go very often to pages similar to https://dodona.be/en/courses/???/series/???/activities/???/submissions/.

In my opinion it makes sense to add `Detect plagiarism` button(s) in the evaluation view on pages similar to https://dodona.be/en/evaluations/???/ to make it more discoverable. Plagiarism checks makes most sense if the submissions are graded and students can earn points.

This is a quick mock-up what it could look like:
image

Detecting for plagiarism for a series in it's entirety would be even better, but Dolos doesn't support that.

jorg-vr:
This pull request adds detect plagiarism buttons to activities in an evaluation. These generate a plagiarism report using the submissions selected for the evaluation.

I also updated the design a bit. 
- 'Evaluate' has become an outline button instead of an icon caret. This makes it much more discoverable for new users and it deserves to be seen as it is probably the primary action.
- As there are two buttons now, I gave the exercise table full page width. The help text has been moved to an expandable details block at the top
 
image
image

Closes #5528, #2503 and #3457


chvp (approved):
Maybe this was discussed in the initial PR, but I don't particularly like the wording "Plagiaat controleren". As far as I understand it, Dolos can never say with certainty plagiarism occurred, only whether certain submissions are similar to others.

jorg-vr:
This has not been discussed before. But I am not sure about the alternative.

I think the button should clearly state the intent of the action, which is detecting plagiarism.
More detailed information about what to do with the info, and witch caveats to take  into account should be listed in Dolos.

But I am open to changing the button text if you have a good alternative suggestion

app/views/evaluations/_exercises_progress_table.html.erb
```diff
@@ -1,11 +1,13 @@
+<%= javascript_include_tag 'dolos' %>
```
bmesuere, line 1: I didn't do a search, but is this a common pattern in our code base? I thought we included all js in the head through a yield/content_for combination.

bmesuere (changes requested):

bmesuere (approved):

PR #5573, opened 2024-06-03
Don't try to show timezone or iframe warning if not present

### LINKED ISSUE #5572: Hide timezone warning sometime errors
The code to hide the timezone warning is called on every page, but the warning itself isn't included on every page (e.g. the sign in page). This causes an error on these pages.

The error was introduced when we migrated away from jquery where such calls failed silently when the query selector yielded zero results.

Sentry Issue: [DODONA-FRONTEND-2](https://dodona.sentry.io/issues/643702/?referrer=github_integration)

```
TypeError: Cannot read property 'classList' of null
  at checkTimeZone (./app/assets/javascripts/utilities.ts:80:32)
  at None (/en/series/2897/overview/:583:10)
```

jorg-vr:
This pull request fixes #5572

I also applied the same fix for check Iframe, but I am open for discussion on that one, as we might want to get notified should that warning not be present when desired.

Closes #5572.

### DIFF
--- a/app/assets/javascripts/utilities.ts
+++ b/app/assets/javascripts/utilities.ts
@@ -77,13 +77,13 @@ function getArrayURLParameter(name: string, _url?: string): string[] {
 
 function checkTimeZone(offset: number): void {
     if (offset !== new Date().getTimezoneOffset()) {
-        document.querySelector("#time-zone-warning").classList.remove("hidden");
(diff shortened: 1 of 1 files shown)


jorg-vr:
We haven't yet encountered the iFrameIssue on sentry. It was just right below in the code, so my I fell upon it

And yes the second ?. is required

niknetniko (approved):

PR #5580, opened 2024-06-04
Don't try to enable submit button when not logged in

jorg-vr:
For some reason the error does not reproduce on my devise, which made it take a lot longer to debug.
But simply logging the the button showed it as null when not logged in.

When logged in, the button is always present.

Fixes [DODONA-FRONTEND-9](https://dodona.sentry.io/issues/773861/)

### DIFF
--- a/app/assets/javascripts/exercise.ts
+++ b/app/assets/javascripts/exercise.ts
@@ -318,7 +318,9 @@ async function initExerciseShow(exerciseId: number, programmingLanguage: string,
     }
 
     function enableSubmitButton(): void {
-        if (!loggedIn) return;
+        if (!loggedIn) {
+            return;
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/exercise.ts
```diff
@@ -318,13 +318,17 @@ async function initExerciseShow(exerciseId: number, programmingLanguage: string,
     }
 
     function enableSubmitButton(): void {
+        if (!loggedIn) return;
```
niknetniko, line 321: I think we generally prefer if's to have braces (or at least I do)

niknetniko (changes requested):

niknetniko (approved):

PR #5581, opened 2024-06-04
Make iframe methods available on submissions page

jorg-vr:
Fixes [DODONA-FRONTEND-A](https://dodona.sentry.io/issues/773912/)


bmesuere (approved):
The same needs to happen on the series overview page

PR #5582, opened 2024-06-04
Fix sentry release information

jorg-vr:
This pull request adds release information to sentry. It also takes the environment information from rails instead of recalculating it based on hostname.

It also removes the sending of release info to sentry, as the release name is now contained in the tag. And sourcemaps are already pushed upon build.

I have also loaded sentry within frames (where it wasn't present yet).

### DIFF
--- a/app/assets/javascripts/sentry.ts
+++ b/app/assets/javascripts/sentry.ts
@@ -2,7 +2,7 @@ import * as Sentry from "@sentry/browser";
 
 export function initSentry(): void {
     const environmentTag = document.querySelector("meta[name='environment']");
-    const environment = environmentTag ? environmentTag.getAttribute("content") : "unknown";
+    const environment = environmentTag?.getAttribute("content") ?? "unknown";
     const releaseTag = document.querySelector("meta[name='version']");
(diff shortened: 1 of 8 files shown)


app/assets/javascripts/sentry.ts
```diff
@@ -1,18 +1,15 @@
...
-    "dodona.be": "production",
-};
-
 export function initSentry(): void {
-    const environment = window.location.hostname in HOST_TO_ENVIRONMENT ?
-        HOST_TO_ENVIRONMENT[window.location.hostname] : "unknown";
+    const environmentTag = document.querySelector("meta[name='environment']");
+    const environment = environmentTag ? environmentTag.getAttribute("content") : "unknown";
+    const releaseTag = document.querySelector("meta[name='version']");
+    const release = releaseTag ? releaseTag.getAttribute("content") : "unknown";
```
niknetniko, line 7:
```suggestion
    const release = releaseTag?.getAttribute("content") ?? "unknown";
```

niknetniko (approved):

bmesuere (approved):

PR #5607, opened 2024-06-13
Disable plagiarism detection when language is not supported by dolos

### LINKED ISSUE #5591: Disable plagiarism detection when programming language is not supported by Dolos
Issue https://github.com/dodona-edu/dolos/issues/1558 is caused by Dodona launching a plagiarism detection analysis for the HTML programming language.

Dolos will probably never support plagiarism detection on HTML exercises (https://github.com/dodona-edu/dolos/issues/1029#issuecomment-1404733726). It would make sense to hide or disable this button for those exercises.

jorg-vr:
This pull request disables the detect plagiarism button for languages that aren't supported by dolos:
image

A tooltip with this explanation is shown on the disabled button.
I had to introduce some new disabled button css to be able to support this tooltip.

The dolos languages are now also hardcoded. Making an API call every time for the supported languages seems a bit overkill. But we'll have to maintain this list. @rien I don't know how often these are expected to change? 

Closes #5591

### DIFF
--- a/app/assets/javascripts/dolos.ts
+++ b/app/assets/javascripts/dolos.ts
@@ -6,9 +6,19 @@ import { i18n } from "i18n/i18n";
 
 const LOADER_ID = "dolos-loader";
 const DOLOS_URL = "/dolos_reports";
+// List from https://github.com/dodona-edu/dolos/issues/1029 on 2024-06-13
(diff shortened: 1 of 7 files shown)


bmesuere (approved):

rien:
The list in the issue you are refering to is not complete (for example: R is missing) For a better up-to-date list you can reffer to https://github.com/dodona-edu/dolos/tree/main/parsers

Issue #5548, opened 2024-05-23
Link submissions to series


jorg-vr:
We should keep track of in which series a submission to an exercise happened.

Doing this will allow better access management control, eg. knowing and restricting which students participate in a certain test series.

This will have consequences when a teacher removes a series, or moves an exercise to a different series after students have submitted.
We'll have to treat these cases properly.

We will also face the difficulty of assigning past submissions to a specific series.

pdawyndt:
About adding/removing series: for each test or exam we now make two series ahead of the exam that each have their own deadlines (one for students that follow the regular deadline and one for students that get extra time). Immediately after the test/exam we remove the second series, create an evaluation that combines both groups of students, and manually adjust the selected submissions for those students that got extra time. In the new scenario, this is also something we would have to rethink. 

If personal deadlines could be linked to individual students, that would no longer require us to temporarily create multiple series, and the above issue would be resolved.

PR #5609, opened 2024-06-14
Save series_id for submissions

jorg-vr:
This pull request saves the series_id when creating a submission or an activity read state.

It also limits visibility of hidden series to course_members.

This series ID is used for:
- better links to the activity in series in course from a submission (defaults to previous behavior if no series present)
- determining the next suggested activities on the homepage (will only suggest course if no series_id present)
- validating if activity is accessible within that specific series when submitting (defaults to previous behavior if no series present)

This ID currently isn't used for the activity status. This is planned for a next step, but we'll also need a decent change for existing submissions.

Next steps:
- Use series id for activity status
- Handle removal of series/removal of activities from series 
- Update existing submissions or correctly handle submissions without series_id

- [X] Tests were added

Part of  #5548


pdawyndt:
While working on this issue, we might also consider storing some other submission properties that are now missed. It concerns those properties that are passed to the judge upon submission. One such property is the language setting of the user upon submission. As this is not stored, we no longer have that information upon re-evaluation. As a result, the language setting of the person re-evaluating is taken as the language passed to the judge upon re-evaluation (which changes the feedback in that respect).

app/assets/javascripts/exercise.ts
```diff
@@ -139,7 +139,7 @@ function initExerciseDescription(): void {
     initCodeFragments();
 }
 
-async function initExerciseShow(exerciseId: number, programmingLanguage: string, loggedIn: boolean, editorShown: boolean, courseId: number, _deadline: string, baseSubmissionsUrl: string, boilerplate: string): Promise<void> {
+async function initExerciseShow(exerciseId: number, programmingLanguage: string, loggedIn: boolean, editorShown: boolean, courseId: number, _deadline: string, baseSubmissionsUrl: string, boilerplate: string, seriesId: number): Promise<void> {
```
bmesuere, line 142: This is becoming a very long parameter list. Do we already use option objects in other places.

app/models/activity.rb
```diff
@@ -313,15 +313,24 @@ def usable_by?(course)
...
   end
 
-  def accessible?(user, course)
-    if course.present?
+  def accessible?(user, course: nil, series: nil)
+    if series.present?
+      return false if course.present? && course != series.course
+
+      course = series.course
+      return false unless series.accessible_to?(user) && series.activities.pluck(:id).include?(id)
```
bmesuere, line 321: Is there a reason why this return is here and not 2 lines higher with the other early termination return statement?

app/models/activity.rb
```diff
@@ -313,15 +313,24 @@ def usable_by?(course)
...
       if user&.course_admin? course
         return false unless course.activities.pluck(:id).include? id
       elsif user&.member_of? course
         return false unless course.accessible_activities.pluck(:id).include? id
       else
         return false unless course.visible_activities.pluck(:id).include? id
       end
+    end
+
+    if course.present?
```
bmesuere, line 333: I'm a bit confused about this `course.present?` here. There is already a similar check in the elsif above and all paths end in a return there. So we only end up here as a continuation of the first if above if none of the 2 return statements return? It seems like this could be restructured to be more clear. At least, comments should be added explaining all these cases.

app/models/submission.rb
```diff
@@ -57,6 +59,7 @@ class Submission < ApplicationRecord
   scope :before_deadline, ->(deadline) { where(submissions: { created_at: ...deadline }) }
   scope :in_time_range, ->(start_date, end_date) { where(created_at: start_date.to_datetime..end_date.to_datetime) }
   scope :in_course, ->(course) { where course_id: course&.id }
+  scope :in_series, ->(series) { where series_id: series&.id }
   scope :in_series, ->(series) { where(course_id: series.course.id).where(exercise: series.exercises) }
```
bmesuere, line 63: I don't think we can have 2 `in_series` scopes?

db/migrate/20240614111643_add_series_id_to_submissions.rb
```diff
@@ -0,0 +1,6 @@
+class AddSeriesIdToSubmissions < ActiveRecord::Migration[7.1]
+  def change
+    add_column :submissions, :series_id, :integer
+    add_index :submissions, :series_id
```
bmesuere, line 4: Adding an index on this table has a big impact. The current indexes were added only after checking if they were used (and what the best combination of potentially combining multiple columns in the index was).

bmesuere (changes requested):

app/models/submission.rb
jorg-vr: no, and this new one should not have been in this update.
Before we can change this scope, we'll have to determine what to do with submissions without series_id

app/models/activity.rb
jorg-vr: I've added comments

The reason this if is separate is that the course can be updated in the 'if series.present?' case

Also the `elsif course.present?` can be removed once we disallow submitting in a course without series, which will greatly simplify this code

app/assets/javascripts/exercise.ts
jorg-vr: not that I know, but we can start doing it here if you desire

bmesuere (approved):

app/models/series.rb
```diff
@@ -206,4 +206,24 @@ def invalidate_caches(user)
     invalidate_started?(user: user)
     invalidate_wrong?(user: user)
   end
+
+  def visible_to?(user)
```
niknetniko, line 210: Would it be possible to use these methods in the series policy as well? Because now this logic is essentially duplicated in the `show?` method of the series policy, while I think we should strive to only have one source of truth for these kinds of things. Alternatively, if there are differences, we should probably put them in a comment here or in the policy to indicate why.

niknetniko (commented):
Besides the single comment, the code looks good.

Do we also want to include this information in the JSON API, or is it better to do this later?

app/models/series.rb
jorg-vr: Even though there was a difference in implementation I opted for reusing the code.

This results in a change: hidden series are now only accessible to users who are member of a course (Instead of any signed in user as was the case previously)

I think this is an improvement

niknetniko (approved):

PR #5642, opened 2024-06-25
Allow expand/collapse of targeted feedback groups

### LINKED ISSUE #5135: Expand/collapse feedback card that has the focus
The feedback card that has the focus is automatically expanded. The collapse status cannot be toggled, which seems intentional, but leads to weird UX. I would suggest that all feedback cards can be expanded/collapsed, even the one that has the focus.

jorg-vr:
This pull request allows targeted feedback groups to be toggled. They are still expanded upon clicking the link by javascript instead of css.

Closes #5135

### DIFF
--- a/app/assets/javascripts/submission.ts
+++ b/app/assets/javascripts/submission.ts
@@ -7,6 +7,7 @@ function initSubmissionShow(parentClass: string, mediaPath: string, token: strin
         initTabLinks();
         initCollapseButtons();
         initHideCorrect();
+        initTabSummaryLinks();
         contextualizeMediaPaths(parentClass, mediaPath, token);
     }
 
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/submission.ts
```diff
@@ -38,6 +39,16 @@ function initSubmissionShow(parentClass: string, mediaPath: string, token: strin
         });
     }
 
+    function initTabSummaryLinks(): void {
+        document.querySelectorAll(".tab-summary-icons a").forEach(l => {
+            l.addEventListener("click", () => {
+                const groupId = l.attributes["href"].value;
+                const group = document.getElementById(groupId.substring(1));
```
bmesuere, line 46: add a comment explaining this

bmesuere (approved):
The css solution was so much cleaner 😢

PR #5644, opened 2024-06-25
Revert to original motivational messages

### LINKED ISSUE #4598: Revert to original motivational messages

jorg-vr:
Closes #4598


bmesuere (approved):
The reason we hadn't reverted this before is that we also wanted to improve the experience for the user after correctly solving an exercise. I'm ok with reverting this now, but also open a new issue for improving the experience as well.

PR #5646, opened 2024-06-25
Add ctrl+enter/cmd+enter keyboard shortcut to submit code from editor

### LINKED ISSUE #3564: Run code keyboard shortcut
It would (in my opinion) be quite helpful if it were possible to run code with a simple shortcut from the keyboard. After developing locally (using an IDE, ...), we could then just copy-paste our code into the Dodona input field, then hit the shortcut (as we have our fingers already on the Ctrl-V keys).

I suggest the combination Ctrl-Enter (which is the same as Github uses for submission and I believe quite a few others).

jorg-vr:
This pull request adds the keyboard shortcut Ctrl-Enter (and cmd+enter on mac) for submitting from the editor.

Closes #3564

### DIFF
--- a/app/assets/javascripts/editor.ts
+++ b/app/assets/javascripts/editor.ts
@@ -18,6 +18,7 @@ import {
     highlightActiveLine,
     highlightActiveLineGutter,
     highlightSpecialChars,
+    KeyBinding,
     keymap,
     lineNumbers
 } from "@codemirror/view";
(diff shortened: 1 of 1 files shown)


bmesuere (commented):
Is it always ctrl, or will it use cmd on mac (which is the behaviour we want)?

bmesuere (approved):
👍 I updated the title

PR #5549, opened 2024-05-23
Redesign the activity page

jorg-vr:
This pull request is currently exploratory

- [ ] Tests were added
- [ ] Documentation update can be found at dodona-edu/dodona-edu.github.io#

Closes # .

PR #5650, opened 2024-06-27
Allow collapsing series on a course page

### LINKED ISSUE #5457: Hiding/collapse exercise series
Dodona used in a secondary school here, all students combined in one course.

For evaluations we make seperate series with a selection of assignments and share these with the students using the secret link + SafeExamBrowser. However, since there are many classes and many evaluations things are getting **messy**. We are losing the overview ourselves.

Here's an view of all the series we've added so far...
Schermafdruk van 2024-03-28 22 52 11

We were thinking about a seperate Dodona course where only the evaluations would be handled, but perhaps that's not a good solution. For example, the overview of the results for each student would be visible in the main course then.

Being able to structure the different series more would be a better solution. Creating a folder-like-structure  with 'graded' series would then perhaps add the possibility to 'collapse' this entire folder and hide all these series from view.

jorg-vr:
This pull request adds buttons to expand/collapse a series on a course page.

This option is stored in local storage, so remembered while on the same machine, without creating unwanted database tables.

Can be useful for teachers with too many series because of tests/exams.
But it can also be useful for students who want to hide past/finished series.

Right now this is quite hidden as I consider it an advanced feature. Should we notice that it is highly appreciated we should consider moving it to a chevron in the top right corner. 

image
image
image

Peek 2024-06-27 14-47

Closes #5457

### DIFF
--- a/app/assets/javascripts/course.ts
+++ b/app/assets/javascripts/course.ts
@@ -69,6 +69,11 @@ class Series {
         return document.getElementById(this.cardId);
     }
 
+    private get scrollspy(): HTMLElement {
+        const anchor = this.card.querySelector(".anchor");
+        return document.querySelector(`#scrollspy-nav a[href="#${anchor.id}"]`);
+    }
+
     static findAll(cardsSelector = ".series.card"): Array<Series> {
         const cards = document.querySelectorAll(cardsSelector);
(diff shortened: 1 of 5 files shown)


jorg-vr:
we'll try a different redesign after #5549

PR #5791, opened 2024-09-12
Remove legacy sign in method migration for office 365 and smartschool

jorg-vr:
This pull request removes the migration for the legacy sing in methods for smartschool and office365 which were introduced two years ago.

Smartschool and office 365 accounts that haven't signed in for two years will no longer be able to sign in to their old account.

The database columns `identifier_based_on_username` and `identifier_based_on_email` still exist, should we ever need to fix an existing user and confirm that it was actually an affected user.


bmesuere (commented):
Do we log if this code path is taken? Maybe we can first log it using events (https://github.com/dodona-edu/dodona/blob/main/app/models/event.rb) or something similar and remove it after a month if it wasn't used.

jorg-vr:
I will close this pr for now. We have logged 28 usages of these legacy sign in methods in the last month (https://dodona.be/en/events/?event_type=other)

So the time does not yet seem right to block of this method

PR #5838, opened 2024-10-01
Fix filtering on new saved annotation page

jorg-vr:
This pull request fixes the filtering of annotations on the 'new saved annotations' page.

This issue was a quite complex rails pluralization issue.

Problem was `joins(:submission)` creates `INNER JOIN submissions ON submissions.id = annotations.submission_id` while filters like `where(submission: { exercise_id: exercise_id })` create `INNER JOIN submissions AS submission ON submission.id = annotations.submission_id`.

If both the `joins` and the `where` clause are present, the table name in the query is `submission` singular. If only the join clause is present the table name is `submissions` plural. 

Starting to use 'submission' singular in most places and avoiding explicitly naming the table in pluck clauses resolves the issues.

- [x] Tests were added


jorg-vr:
try to avoid pluck and no explicit sql

bmesuere (approved):

PR #5854, opened 2024-10-10
Fix legacy 'exercises' path support

jorg-vr:
This pull request fixes the support for the path `/courses/?/series/?/exercises/?/submissions`. 

The `/exercises` paths have all been renamed to `/activities`, yet our routes still offer legacy support for `/exercises`. I didn't find any internal references that use the `/exercises` name, so I assume the bug was discovered by an external link.

I am also open to dropping the support for `/exercises`

- [x] test were added

Fixes an issue come in by mail, where a user tried to navigate to https://dodona.be/en/courses/156/series/2328/exercises/1963184607/submissions
For reference, the internally linked path https://dodona.be/en/courses/156/series/2328/activities/1963184607/submissions does work

### DIFF
--- a/app/controllers/submissions_controller.rb
+++ b/app/controllers/submissions_controller.rb
@@ -175,7 +175,9 @@ class SubmissionsController < ApplicationController
     end
 
     @series = Series.find(params[:series_id]) if params[:series_id]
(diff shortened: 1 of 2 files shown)


bmesuere (approved):
Since `exercises` is supported in our routes, we should handle it properly in the controllers. 👍

PR #5864, opened 2024-10-15
Add csv differ support for empty rows

jorg-vr:
This pull request fixes a bug in the csv differ that occurred when empty lines where present.

In the simplified view the diff is displayed by column. For this reason, each row needs the same amount of columns.
To solve this I padded too short rows with empty string

- [x] Tests were added


bmesuere (approved):
👍 
Strange that github says that the new lines aren't covered by tests while you did add one.

jorg-vr:
Seeing the large reduction in codecov percentage, something probably went wrong when uploading the coverage results. Happens sometimes

PR #5876, opened 2024-10-18
Fallback to textdiffer when csv parsing fails

jorg-vr:
This pull request lets the differ fallback to the Textdiffer in case the csv is malformed.

This issue was detected by submissions from the postgres judge, which returned to many escaping characters before a newline for a while. This issue is already resolved in the judge as well.

Example of a problematic submission: https://dodona.be/en/submissions/19072577.json

- [x] Tests were added


app/helpers/renderers/feedback_table_renderer.rb
```diff
@@ -319,17 +319,31 @@ def messages(msgs)
...
-    else
-      TextDiffer
+    if t[:format] == 'csv'
+      begin
+        return CsvDiffer.new(t[:generated], t[:expected]) if CsvDiffer.limited_columns?(t[:generated]) && CsvDiffer.limited_columns?(t[:expected])
+      rescue CSV::MalformedCSVError
+        # default to text if parsing failed
+      end
     end
+    TextDiffer.new(t[:generated], t[:expected])
```
bmesuere, line 341: I'm not too fond of the flow of this code. The fallback to the textdiffer is implicit and relies on error handling. I would stay closer to the original code and duplicate the textDiffer line in the rescue block.

bmesuere (commented):
I'm not too fond of these changes code-wise.

It duplicates some logic + in the new logic, fallback to the textdiffer is implicit and relies on error handling (i.e. the rescue block is empty)

jorg-vr:
I removed the code duplication and also made the choice for the TextDiffer explicit

I think the deduplication is an improvement but I personally prefer the the implicit fallback as it simplifies the amount of cases, because most cases have the same result. With implicit fallback the method would look like this:

```ruby
  def with_differ_class(t)
    if t[:format] == 'csv'
      begin
        return yield(CsvDiffer) if CsvDiffer.limited_columns?(t[:generated]) && CsvDiffer.limited_columns?(t[:expected])
      rescue CSV::MalformedCSVError
        # use the default differ
      end
    end

    yield(TextDiffer)
  end
```

bmesuere (approved):

PR #5891, opened 2024-10-22
Added json-output for the questions endpoint

### LINKED ISSUE #5890: JSON output generated for questions of a course is just an empty list
`https://dodona.be/nl/courses/<course ID>/questions.json` always returns an empty array.

thvmulle:
This pull request fixes the empty list output generated by the JSON response for `https://dodona.be/nl/courses/<course ID>/questions.json`.

The JSON now returns an object that has three properties:
```json
{
"unanswered": [...],
"in_progress": [...],
"answered": [...]
}
```

Each of these properties contains a list with questions that are either unanswered, being answered or have been answered.

- [x] Tests were added

Closes #5890 .

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,20 +1,24 @@
 module.exports = {
     preset: "ts-jest/presets/js-with-ts",
     reporters: ["default", "jest-junit"],
-    globals: {
-        "ts-jest": {
-            tsconfig: {
-                "allowJs": true,
-                "target": "es6",
-            },
-        },
-    },
     setupFiles: ["<rootDir>/test/javascript/setup-jest.ts"],
     roots: ["test/javascript/", "app/assets/"],
     moduleDirectories: [
         "node_modules",
         "app/assets/javascripts"
     ],
+    transform: {
+        '\\.tsx?$': [
+            'ts-jest',
+            {
+                tsconfig: {
+                    "allowJs": true,
+                    "target": "es6",
+                },
+            }
+        ],
+        '\\.jsx?$': ['babel-jest', {plugins: ['@babel/plugin-transform-modules-commonjs']}]
+    },
     transformIgnorePatterns: ["node_modules/?!(d3)"],
     collectCoverageFrom: ["app/assets/javascripts/**/*.{js,ts}"],
 };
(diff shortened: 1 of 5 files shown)


jest.config.js
```diff
@@ -1,20 +1,24 @@
 module.exports = {
```
jorg-vr, line 1: you've done something wrong when merging the changes from main into this branch.

Not a big issue in this case but it makes it harder to review your changes. So something to look at in the future

test/controllers/courses_controller_test.rb
```diff
@@ -905,6 +905,72 @@ def with_users_signed_in(users)
     end
   end
 
+  test 'super admins are able to view questions in JSON format' do
+    add_admins
+    super_admins = @admins.reject(&:student?)
+    with_users_signed_in super_admins do |who|
```
jorg-vr, line 911: you should also add `with_delayed_jobs` here. It doesn't change the outcome f the test, but without it, you might as well remove the `:in_progess` case

jorg-vr (approved):

jest.config.js
thvmulle: You're right, something went wrong during the rebase from the main branch. Future PR's will not be this messy.

test/controllers/courses_controller_test.rb
```diff
@@ -905,6 +905,79 @@ def with_users_signed_in(users)
...
+
+      json_response = response.parsed_body
+
+      assert_empty json_response['unanswered'], 'There should be 0 unanswered questions in the JSON response'
+      assert_empty json_response['in_progress'], 'There should be 0 in_progress questions in the JSON response'
+      assert_empty json_response['answered'], 'There should be 0 answered questions in the JSON response'
+    end
+  end
+
+  test 'non admins cannot view questions in JSON format' do
```
jorg-vr, line 968: I prefer the method used here:
image

test/controllers/courses_controller_test.rb
```diff
@@ -905,6 +905,79 @@ def with_users_signed_in(users)
...
+    add_not_admins
+    with_users_signed_in @not_admins do |who|
+      get questions_course_path(@course), as: :json
+
+      if who == 'not signed in'
+        assert_response :unauthorized, "#{who} should not be able to view questions in JSON format"
```
jorg-vr, line 974: I am still slightly worried about the fact that this test also succeed when this line was `assert_response :ok`

Do you have any idea why? (if no, I'll do some further research as this might cause false positives in other tests)

jorg-vr (approved):

jorg-vr (approved):

PR #5984, opened 2024-12-03
Allow using `use` tag in sanitized html

jorg-vr:
This pull request adds the possibility to use `use` tags in html that will pass through our sanitizer.
Looking at the code in the underlying library I deem it safe to use and I have added testcases that should warn us if support in that library changes.

- [x] Tests were added


pdawyndt:
@jorg-vr my use case for using the `use` tag in SVG is to reference the id of a `defs` tag in the same attribute (`#id`) so if that could be supported than we don't even need to base64 detour and would be OK for me

jorg-vr:
@pdawyndt your usecase would be supported by this pr

bmesuere (approved):

pdawyndt:
Confirmation that everything works on my side. Thanks for taking this up.

PR #6049, opened 2025-01-08
Correctly copy all series info when copying courses

### LINKED ISSUE #5743: Copy advanced options from a series when make a new course
When make a new course based on a previous course the option of the exercise numbers is not being copied over.

Steps to reproduce, copy a course using these options:
copy_course

Then in an exercise-series none of the advanced options are ticked of.
advanced options

jorg-vr:
This pull request fixes multiple bugs related to copying courses.

Firstly a bug in the test for students not being able to copy hidden series. This test also didn't copy visible series as the student wasn't correctly added to the course in the test.

Secondly a bug caused the order of series which have not been manually reordered to be reversed in a copied course.

Thirdly I made sure that advanced series settings are also copied when a course is copied: https://github.com/dodona-edu/dodona/issues/5743

- [x] Tests were added

Closes #5743

### DIFF
--- a/test/controllers/courses_controller_test.rb
+++ b/test/controllers/courses_controller_test.rb
@@ -747,7 +747,7 @@ class CoursesPermissionControllerTest < ActionDispatch::IntegrationTest
     assert_includes course.subscribed_members, user
   end
 
-  test 'a course copied by staff who isn't course admin should not include hidden/closed series' do
+  test "a course copied by staff who isn't course admin should not include hidden/closed series" do
     user = create :user, permission: :staff
     course = create :course
     course.enrolled_members << user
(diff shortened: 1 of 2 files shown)


test/controllers/courses_controller_test.rb
```diff
@@ -748,11 +748,16 @@ def with_users_signed_in(users)
   end
 
   test 'a course copied by a regular student should not include hidden/closed series' do
```
bmesuere, line 750:
```suggestion
  test 'a course copied by staff who isn't course admin should not include hidden/closed series' do
```

bmesuere (approved):

PR #6062, opened 2025-01-15
Upgrade to rails 8.0.1

jorg-vr:
This pull request upgrades rails from 7.2.2.1 to [8.0.1](https://rubyonrails.org/2024/9/27/rails-8-beta1-no-paas-required)

Not all our dependencies already support rails 8.
For `exception_notification` I had to temporarily track a branched updated version, until an official release is made.
For `annotate` I had to switch to `annotaterb`, a fork of the original which does receive regular updates. These changes where applied in #6068 

I also had to edit the `i18njs` settings. This gem loads the rails translations in memory and then writes a subset of them to a json file. There was a bug in how they extracted that subset. For this they made use of an internal rails api, which has probably changed slightly.
After multiple hours of investigation I didn't find the exact cause, but I found that fully specifying the keys (with wildcards) instead of just the start, does fix the issue.
With the new patterns, `datetime` is no longer matched as key. But I didn't find any usages in the fronted, so I didn't add a specific pattern for it.

## Changed settings
In development changes follow the new defaults. Although I only changed some settings we didn't really use anyway.

In production I removed `config.action_controller.perform_caching = true` as this is also done in the new default. This doesn't have any actual impact as true is the default value.

- [x] Test deploy on naos

I moved some potential future work to a separate issue to not overdo this pr: https://github.com/dodona-edu/dodona/issues/6064


bmesuere (changes requested):
Can you split the annotate upgrade to a separate PR? Since this is a major rails upgrade, it would be good to only have those change here.

bmesuere (changes requested):
Can you add a bit more info to the PR start post about:
- why were the js time translations removed? We use an external library for js i18n so it's strange that this has an impact.
- the changes to the development environment config
  - is there a reason why the merge was removed from the headers? We now seem to overwrite previous headers? Or was this a default config change?
  - we now use a memory store cache for development instead of the cache being disabled? Why?
- what about the cache setting in production?

jorg-vr:
I've added some info to the pr header.

> is there a reason why the merge was removed from the headers? We now seem to overwrite previous headers? Or was this a default config change?

There was a default config change here, but I did overlook that we set other headers before this line which didn't happen in the default config. Thus I've updated the pr to keep the 'merge'

bmesuere (approved):
👍

PR #6067, opened 2025-01-20
Update all annotions to new gem order

jorg-vr:
This reverts commit 5c66322b930e4be27aba57330bf6dfe9cd59465f.

This pull request updates or db annotaions in our ruby files to the format genrated by `annotate_rb`, which we upgraded to in #6062 

This is only a change in order, with the columns now being sorted alphabetically, except for default rails columns and foreign keys, which are always at the end.

I think this new order is an improvement, so didn't try to maintain the old order with settings updates.

### DIFF
--- a/.annotaterb.yml
+++ b/.annotaterb.yml
@@ -26,8 +26,8 @@
 :include_version: false
 :show_check_constraints: false
 :show_complete_foreign_keys: false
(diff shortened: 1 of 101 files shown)


jorg-vr:
closed in favor of #6068

